PR #23, opened 2026-08-10
docs(benchmark): Terminal-Bench 2.0 four-way study + improvement plan

OsherElhadad:
## What

The second benchmark of the study, after SWE-bench Verified: **Terminal-Bench 2.0, 89 open-ended terminal tasks**, `claude-code` on `aws/claude-sonnet-5`, run live through the harness. Same four arms as SWE — baseline (`off` passthrough), context-guru (`codesmart`), headroom (`hd-cache`), rtk.

The claude-code trajectory parser, the cache-aware cost model and the summarizer are *agent*-specific rather than benchmark-specific, so every number is computed identically to the SWE arms and the two benchmarks are directly comparable in methodology. Only the Harbor dataset and jobs-root differ.

## Headline (89-task raw)

| dimension | baseline | context-guru | headroom | rtk |
|---|--:|--:|--:|--:|
| solved / 89 | 56 (62.9%) | 58 (65.2%) | **64 (71.9%)** | 55 (61.8%) |
| total billed cost | **$100.81** | $102.55 | $114.75 | $118.83 |
| cache-read tokens | 216.0M | 204.6M | **198.0M** | 254.0M |
| cache-write tokens | **4.01M** | 6.53M | 12.37M | 7.05M |
| mean steps (completed) | **31.5** | 34.7 | 32.0 | 40.0 |

## The correction, stated up front

Six baseline trials are **degenerate** — the baseline aborted in 2–6 steps (16–800 s) where the compaction arms ran 50–160 steps, so the cost delta on those six measures the baseline not doing the work. `extract-moves-from-video` alone is **$24.10 of headroom's $24.65 "regression"**.

Recomputed over the **83 clean tasks: context-guru −9.7%, headroom −16.0%, rtk +6.4%.** Both proxies save on Terminal-Bench; only rtk regresses. The comparison page opens with this caveat and the six tasks are queued for re-run at low concurrency; the tables will be regenerated then.

## What the study establishes

- The agent is **~98% cached** on TB too, so cache-read is again the largest cost term — but **cache-write**, a rounding error on SWE-bench, is the deciding term on TB's ~1.7M-token contexts.
- **Cost tracks agent steps**, `r = 0.95` across every arm and both benchmarks.
- **One cache-write costs 11.5 cache-reads** — so mutating already-cached content is the expensive mistake, not sending too many tokens.
- **Unique token removal is 0.02–0.13% of the billed total.** Content compaction is not where the money is.
- **`cache_control` is metadata, not hashed content** — moving a breakpoint is free and lossless.

`docs/results/improvement-plan.md` carries that synthesis and the prioritized work it implies.

## Also in this PR

`swebench.py`: captures and dumps move from a fixed `/tmp` path to the run's own jobs-root. `start_proxy` unlinks the capture path, so launching any new run truncated the capture an earlier analysis was computed from — which is how a 472-request capture was lost mid-analysis. Adds the `cacheonly` arm, which removes no content tokens at all, so any cost delta vs `off` is purely breakpoint placement.

## Reproduction

`docs/results/REPRODUCE.md` §7 (baseline: task-list generation, `n=24` concurrency feasibility on this box, the timeout/budget methodology, task-metadata extraction) and §7b (the three framework-arm commands). Every credential is read from the environment at run time; nothing is embedded.

## Verification

- `CGO_ENABLED=1 go build -tags cg_skeleton ./...` and `go test -tags cg_skeleton ./...` green (docs/harness only — no library change).
- All 89 tasks carry a scored outcome; timeouts count as reward-0 failures rather than being dropped.

### DIFF
--- a/deploy/harbor/gen_tb_docs.py
+++ b/deploy/harbor/gen_tb_docs.py
@@ -1,0 +1,188 @@
+#!/usr/bin/env python3
+"""Generate the Terminal-Bench 2.0 BASELINE results doc from a rows.json + task
+metadata, mirroring the SWE-bench result pages (totals + cache-aware cost model +
+per-task table) but adding TB-specific breakdowns by difficulty and category and a
+first-class treatment of timeouts — a baseline characterization of *where* the
+claude-code agent is strong/weak, so the later compaction arms have a like-for-like
+reference.
+
+All 89 attempted tasks are included. A task that hit AgentTimeout/VerifierTimeout is a
+genuine failure (reward 0, agent did not finish within its 1.5× wall-clock budget on
+this ~26 s/request gateway) and is flagged as a `timeout`; its partial spend still
+counts toward cost. The primary solve rate is solved / attempted (the standard
+Terminal-Bench metric); a secondary rate over completed-only tasks is also reported.
+
+Usage:
+  gen_tb_docs.py <rows.json> <out.md> [--meta /tmp/tb-runs/task_meta.json] [--summary summary.json]
+"""
+import argparse, json
+from collections import defaultdict
+
+IN, OUT, CREAD, CWRITE = 2e-6, 10e-6, 0.2e-6, 2.5e-6  # same cache-aware price model as the SWE study
+
+
+def billed(r):
+    return (r.get("fresh_input", 0) * IN + r.get("cache_read", 0) * CREAD +
+            r.get("cache_write", 0) * CWRITE + r.get("completion_tokens", 0) * OUT)
+
+
+def agg(rows):
+    n = len(rows)
+    solved = sum(1 for r in rows if (r.get("reward") or 0) >= 1)
+    cost = sum(billed(r) for r in rows)
+    steps = sum(r.get("steps", 0) or 0 for r in rows)
+    wall = sum(r.get("agent_wall_s", 0) or 0 for r in rows)
+    return n, solved, cost, steps, wall
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("rows"); ap.add_argument("out")
+    ap.add_argument("--meta", default="/tmp/tb-runs/task_meta.json")
+    ap.add_argument("--summary", default="")
+    ap.add_argument("--label", default="baseline", help="arm label for the page header")
+    ap.add_argument("--kind", default="baseline", choices=["baseline", "arm"],
+                    help="baseline emits the full narrative; arm emits a compact per-arm page")
+    a = ap.parse_args()
+    rows = json.load(open(a.rows))
+    rows.sort(key=lambda r: r["task"])
+    meta = {}
+    try:
+        meta = json.load(open(a.meta))
+    except Exception:
+        pass
+    for r in rows:
+        m = meta.get(r["task"], {})
+        r["_diff"] = m.get("difficulty", "unknown")
+        r["_cat"] = m.get("category", "unknown")
+        r["_timeout"] = bool(r.get("exception"))
+
+    scored = [r for r in rows if not r["_timeout"]]     # completed within budget
+    timeouts = [r for r in rows if r["_timeout"]]
+
+    is_base = a.kind == "baseline"
+    L = []
+    L.append(f"# Full results — {a.label} (Terminal-Bench 2.0, 89 tasks)\n")
+    if is_base:
+        L.append("Baseline arm: **no compaction** — the `claude-code` agent on `aws/claude-sonnet-5`, "
+                 "run LIVE through the harness against Terminal-Bench 2.0's 89 tasks. Routing goes "
+                 "through the context-guru `off` transparent passthrough proxy (identical plumbing to "
+                 "the compaction arms; zero content change), so this is the like-for-like reference the "
+                 "framework arms are measured against. Cache-aware billed input cost (fresh $2/M · "
+                 "cache-read $0.20/M · cache-write $2.50/M) + output $10/M, recomputed from each trial's "
+                 "own token tiers — the same model as the SWE-bench study. See [REPRODUCE.md](REPRODUCE.md).\n")
+    else:
+        L.append(f"Full per-task results for the **{a.label}** arm on Terminal-Bench 2.0 (`claude-code` on "
+                 "`aws/claude-sonnet-5`, live). Same cache-aware cost model and 4× budget as the other arms. "
+                 "For the four-way analysis (cost decomposition, per-component, verdict) see the "
+                 "**[Terminal-Bench comparison](terminal-bench-comparison.md)**; the reference arm is the "
+                 "**[baseline](terminal-bench-baseline.md)**. See [REPRODUCE.md](REPRODUCE.md).\n")
+
+    n = len(rows)
+    solved = sum(1 for r in rows if (r.get("reward") or 0) >= 1)
+    cost = sum(billed(r) for r in rows)
+    steps_c = sum(r.get("steps", 0) or 0 for r in scored)
+    tcr = sum(r.get("cache_read", 0) for r in rows)
+    tcw = sum(r.get("cache_write", 0) for r in rows)
+    tfresh = sum(r.get("fresh_input", 0) for r in rows)
+    tout = sum(r.get("completion_tokens", 0) for r in rows)
+    hit = 100 * tcr / max(tcr + tcw + tfresh, 1)
+    wall_c = sum(r.get("agent_wall_s", 0) or 0 for r in scored)
+
+    L.append("## Totals\n")
+    L.append("| attempted | solved | solve rate | completed | timed out | total billed cost | mean steps* | cache-hit |")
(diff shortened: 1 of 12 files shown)


OsherElhadad:
## Two corrections pushed — both found by trying to reproduce the published totals

While preparing the follow-up re-run I could not reconstruct the baseline's **$100.81** from `rows-off-final.json`, which sums to **$71.44**. Chasing that gap surfaced two real errors.

### 1. The cost conclusion was wrong, and the direction flips

The baseline is a **two-stage** merge: `merge_tb.py` output *plus* an 11-task `--agent-mult 4.0` rerun overlaid on top. With both stages applied the tokens reproduce the published figure exactly (215,971,427 / 4,011,068 / 58,893 / 4,746,887 = $100.81). That merge step was undocumented; it is now written up in REPRODUCE §7 with the snippet.

With the baseline correctly reconstructed, six trials are **degenerate** — the baseline aborted in 2-6 steps while the compaction arms did the real work:

| task | baseline | context-guru |
|---|--:|--:|
| `mteb-leaderboard` | $0.08 (4 steps) | $6.05 (147 steps) |
| `polyglot-rust-c` | $0.08 (3 steps) | $2.81 (50 steps) |
| `extract-moves-from-video` | $0.02 (2 steps) | $2.69 (112 steps) |

Those three are **$11.5 of apparent regression** on their own. On the **83 clean tasks**:

| | baseline | context-guru | delta |
|---|--:|--:|--:|
| billed model cost | $100.17 | $87.47 | **−12.7%** |
| + context-guru's own haiku cost | — | $90.34 | **−9.8%** |
| solved | 54 | **56** | +2 |
| total steps | 3,160 | **2,899** | **−8.3%** |

So Terminal-Bench does **not** invert the SWE-bench result. context-guru is cheaper, solves more, and takes *fewer* steps — the same shape as its −13% SWE win. The **+1.7%** headline was an artifact of six trials where the baseline didn't run.

To be clear about what is *not* an error: the $102.55 figure is correct as printed — it is model cost ($99.29) plus context-guru's own LLM cost ($3.26). The mistake was comparing it against a baseline total contaminated by degenerate trials.

The raw 89-task tables are **kept as measured**; the correction box, Headline, Verdict, and Bottom line now state the clean-set conclusion and why the two differ. The mechanism finding survives intact and is the durable result: **cache-write, a rounding error on SWE-bench, is the deciding term on TB's ~1.7M-token contexts.**

### 2. The `xdedup` premise is refuted — the plan's §C1 was wrong

The improvement plan called cross-turn dedup "the biggest untapped token lever" on a measured **39.8× re-send factor**. The factor is real; the interpretation was not. Measured on the **raw request captures** (1,325 requests / 51 sessions — not the change-log dumps, which only record messages a component already acted on and so structurally cannot answer this):

- **232 of 232** re-sent large outputs live at exactly **one stable message index** for their whole session. Independently re-checked on a separate pass: of **77** distinct >4 KB tool outputs, **0 ever appeared at a second index**.
- **1124/1124 = 100%** of consecutive turn pairs have the entire previous turn as a byte-identical prefix.
- **Zero** compaction/rewind events on swebench.

The agent **appends**; it does not re-send. Those 5.46M tokens sit in the cached prefix and already bill at the **cache-read** rate. An `xdedup` component would have no legal opportunity to act (`TailOnly` correctly refuses), and rewriting them anyway would convert reads into writes at **11.5×** — causing the exact harm it was meant to prevent. The genuinely duplicated cases are already caught by `dedup` (23 acts, **0 tokens missed** above the size gate).

§C1 is retracted in place with the evidence; C2 (recurrence-aware floor) and C3 (freeze everywhere) are unaffected. Issue #27 is closed as measurement-refuted, with one caveat left open: `capture-tb` showed 19 turns where the message count shrank, and aggressive compaction is the one regime that could make the premise viable — worth re-testing then.

### Methodological lessons now recorded in the docs

**A trial where the baseline aborts is not a measurement** — exclude it rather than average it. And **per-arm totals hide this**, so a per-task paired comparison is the only honest default. Both are in REPRODUCE and the comparison page.

The six-baseline re-run remains queued; it will tighten these figures but is no longer needed to state the conclusion.

OsherElhadad:
## Orchestrator cross-check of the final Terminal-Bench arm

I re-derived the numbers from the raw trial directories (`result.json` + `agent/trajectory.json`, 88 scored trials) rather than from the agent's analyzer, using a deliberately **different** degenerate-exclusion rule, to see whether the conclusion depends on the analysis choices.

### It survives, but the single-task sensitivity is worse than reported

| | agent's rule (paired, 7 excluded) | my rule (baseline ≤6 steps or exception, 12 excluded) |
|---|--:|--:|
| clean tasks | 81 | 76 |
| aggregate cost delta | **−16.4%** | **−13.7%** |
| median per-task | −7.8% | **−6.5%** |
| tasks cheaper | 49/81 | 44/76 |
| solved (cg vs baseline) | 61 vs 53 | 60 vs 53 |

Same direction, same order of magnitude, under two independent exclusion rules — that is the robustness that matters.

**But the leave-one-out is more severe than the agent's figure suggested.** On my clean set:

| | aggregate |
|---|--:|
| all 76 clean tasks | −13.7% |
| **minus `path-tracing`** | **−2.8%** |
| minus `reshard-c4-data` | −11.2% |
| minus `protein-assembly` | −12.7% |

**`path-tracing` alone carries roughly 80% of the headline** (it saved $7.84 on one task). The agent reported −16.4% → −9.1% for that exclusion; on my stricter clean set the drop is −13.7% → −2.8%. Either way the honest summary is the **median: about −6.5% to −7.8% on a typical task**, with the aggregate dominated by a handful of very large trials.

That is not a criticism of the run — it is the §F-1.9 rule (*a sum over heterogeneous tasks can be one task*) doing its job, and the agent added the leave-one-out and median specifically because it caught this pattern early. It just needs to be the headline framing rather than a footnote when this lands in the docs.

### What I consider solidly established

**Cache-write is the clean mechanistic result.** cache-write/cache-read is **1.77–1.86%, indistinguishable from baseline's 1.86%**, where the previous arm ran at **2.86% (+54%)**. That is the plan's "cache-write tax" eliminated, and cache-write is the term the published comparison names as *deciding* on Terminal-Bench. It does not depend on any exclusion rule.

**Reward improved on both rules** (+7 to +8 solved), though at n=1 per task the flip churn (+11/−3) matters more than the net, and all three losses were verified as **capability failures rather than information loss** — HTTP 404, a wrong-cased flag, and a rejected non-fast-forward push. Notably two of the three used *fewer* steps than baseline, which argues against the "compaction hid something, agent redid work" mechanism.

**Own-LLM cost is $0** against the previous arm's $2.97, with `llm_calls` verified at 0 across 3,274 requests.

### Mechanisms verified fired, and one that could not be

- **#42 (cmdfilter)** — firing rate **0.89% → 27.30%**, unique tokens **886 → 31,624 (35.7×)**. I recomputed both from the raw counters.
- **#33 (SSE fast path)** — 43.5% buffered against a pre-fix unconditional 100%. The agent correctly declined to quote the earlier 26% figure, since that was measured on different traffic composition.
- **#40 (freeze TTL)** — **unverifiable in this arm, by construction**: its only callers are `mask`, `failed_run` and `extract_llm`, all three excluded by the chosen config. `frozen_*` all zero is the *expected* result, and no favourable cost number should borrow credit from it.
- **`cachesplit`** — **zero legal opportunity on TB**: the Agent SDK never appends the git/env snapshot the CLI does, so all 73 captured requests carry 3 system blocks and no volatile-tail marker. The −34.1% split win came from SWE-bench CLI traffic.

### Regressions that belong in the published table

1. **`system-administration`: +17.2% cost and −2 solved** — the only category losing on both axes, and the clearest genuine regression.
2. **`security`: +25.6%** — better than the previous arm's +121%, still the wrong direction.
3. **`fresh_input` ~3.8× baseline** — worth ~$0.20 vs $0.05, under 0.3% of the bill, but directionally wrong.
4. **Small-task inflation**: `optimization` +311% (n=1), `games` +50.7%, `model-training` +28.1% — exactly the §1b pattern where marker overhead isn't recovered below ~1M prompt tokens.

Also carrying forward as a stated limitation: **headroom's and rtk's TB artifacts have been pruned from disk**, so those two columns must be cited from the published comparison and labelled as such, not re-derived alongside the others.

OsherElhadad:
Superseded by #59 — this branch rebased onto `main` after all fifteen code PRs landed, plus the merged-system fifth arm you asked to wait for. Branch protection blocks force-push, hence the new branch.

Per your instruction this was held as a draft until the improved results existed. They now do: **61 solved vs baseline 53**, **cache-write back to baseline parity** (1.86% against the previous arm's 2.86%), **$0** own-LLM cost, and added latency down from 449.8 ms to 38.5 ms.

One framing decision worth flagging before you read it: the **−16.4% aggregate is dominated by a single task**. Removing `path-tracing` gives −9.1%, and my own re-derivation with a stricter degenerate rule gave −13.7% dropping to −2.8% on the same exclusion. The doc leads with the **median per-task figure of −7.8%** and publishes both, because they answer different questions and differ by 9 points.

Issue #27, opened 2026-08-10
feat(components): add xdedup — cross-turn duplicate tool-output references (39.8x measured re-send factor)


OsherElhadad:
## Problem statement

`dedup` only catches duplication **within the current request**. Agents re-read the same file across turns, and each re-read arrives as a fresh tool output that `dedup` has no memory of. Measured on Terminal-Bench: mean re-send factor **39.8×**, and in the worst single case a source file of ~**21,957 tokens** was re-sent **167 times ≈ 3.67M tokens**.

This is the **largest token lever in the codebase** and it needs no LLM. Ideal behavior sends the full content once plus 166 tiny references — approaching a **~99.3%** reduction on that output.

Do not use that theoretical figure as evidence. Reproduce representative cases and measure them.

## Current behavior

`components/offload/dedup.go` collapses the *later* duplicate within one request and keeps the first. That is cache-friendly by construction (it mutates the new tail, not the cached prefix) but scoped to a single `req.Input`. Nothing carries a content hash across turns, so turn N+1's copy of the same file looks novel.

## Motivation

Content compaction has been a small lever so far — unique token removal is 0.02–0.13% of the billed total, because the request is ~99.95% cached and the same removal is re-counted every turn. Cross-turn dedup is different in kind: it attacks content that is **genuinely re-sent as new bytes**, and every avoided re-send is a real cache-write avoided. One cache-write costs 11.5 cache-reads, so this compounds. See `docs/results/improvement-plan.md` §C1.

## Desired behavior

A new registered component, `xdedup`. Per session, maintain roughly:

```
contentHash -> (firstSeenMsgIdx, markerKey, metadata...)
```

When a tool output's exact content hash was already sent in an **earlier turn**, replace the repeat with something like:

```
[same as output at step N] <<cg:HASH>>
```

The original must remain recoverable through the existing restoration path.

## Requirements

- cross-**turn**, not merely intra-request;
- session-aware, with no accidental references across unrelated sessions;
- content-hash based;
- frozen/stable representation — once a repeat is referenced, it stays referenced byte-identically on later turns (otherwise the fix causes the cache-writes it was meant to prevent);
- cache-friendly: must respect `Ctx.TailOnly` / `MaxCachedIdx` and never rewrite a message already in the cached prefix;
- deterministic;
- safe restoration;
- bounded storage;
- lifecycle compatible with both sync and async modes (**#31**);
- dashboard metrics for its savings;
- tests for duplicate detection and for restoration.

[...]

PR #34, opened 2026-08-10
perf(extract_llm): economic gate, global result cache, and the honest verdict (#28)

### LINKED ISSUE #28: perf(extract_llm): prompt-cache the preamble, global result cache, economic gate, and value-based triggering (measured ~8x underwater)
## Problem statement

`extract_llm` is the only component that spends money to save money, and on Terminal-Bench it lost: **271 calls, $3.26, ~1,592 s cumulative latency (~450 ms/req)** against ~197,548 unique tokens saved — measured **~8× underwater** against the cache-aware value of those tokens. Worse, roughly **93% of its realized value came from the replay cache**, not from the LLM: most calls re-derived a result the system already had, or should have had.

It also has the worst ergonomics of any component: a hand-tuned `min_tokens` threshold that must be re-picked per workload.

## Current behavior

- **No prompt caching.** `internal/cheapmodel/anthropic.go:32-48` sends the whole prompt as a **single user message** — no `system` block, no `cache_control`. The ~**852-token invariant preamble** is billed as fresh input on every call.
- **Result cache is session-scoped.** Keys carry a session prefix, so identical content in a different session is re-derived. Measured: **82 of 103** unique contents recurred across sessions.
- **No economic gate.** The component decides on size and cadence (`extract_llm.go:66-71`, `:127-128` — `llm_every_n_requests`, `llm_max_per_request`), never on expected value. Note it *does* already carry a partial insight (`:82-90`): on a caching backend, skeletonizing small outputs "saves almost nothing yet costs the compaction LLM + one-time cache-write transitions → +30% billed cost." That reasoning is right and needs to become an actual gate.
- **Trigger is a raw token threshold.** `min_tokens` + `trigger.min_request_tokens`, both manual.

## Motivation

The request is ~99.95% cached, so a token removed from a cached region is worth `$0.20/M`, not `$2/M`. An LLM call that costs ~$0.012 must therefore remove a *lot* of cached tokens to break even — and on a prompt-caching backend it frequently cannot. Either the component becomes cost-aware or it should be off by default on caching backends. See `docs/results/improvement-plan.md` §B3.

### A. Prompt-cache the fixed preamble
Move the invariant instructions into a cacheable stable `system` block with `cache_control`, so repeated calls pay cache-read (`$0.20/M`) instead of fresh input on ~852 tokens. Requires extending `cheapmodel` beyond its current single-user-message shape (`anthropic.go:48`) — keep the interface honest for the OpenAI backend too (`internal/cheapmodel/openai.go`), which has no explicit breakpoints. **Measure the actual effect**; do not assume it.

### B. Investigate reusing the agent's own cached prefix
If the extraction model is the same as the agent model, and it is semantically safe, and the provider's caching rules allow it, and it does not poison the agent's own cache — investigate appending the extraction instruction as a final user message after the existing stable prefix, so extraction reuses the already-cached context.

**Do not implement this blindly.** Prototype, measure, and decide on evidence. Reasons it may be wrong: it would send a ~1.7M-token context to answer a question about one tool output (cache-read is cheap but not free); it risks a cache-write on the agent's prefix, which at 11.5× a read is exactly the mistake this whole workstream is about; and it couples the compaction model to the agent model. Record the decision either way.

### C. Global content-hash result cache
Re-key extraction results primarily on **content hash**, globally, where semantically safe. The same exact content with the same extractor semantics should reuse the previous result.

The key must include everything that materially affects the result: content hash, prompt/extractor **version**, model, and the relevant component configuration. Getting this wrong means silently serving a stale extraction after a prompt change — add the version field first.

Contrast with **#27**, which is deliberately session-scoped: `xdedup` makes a *conversational reference* ("same as step N") that only makes sense in-session, while this caches a *derived result* that is context-free.

### D. Economic gate
Only call the LLM when `expected_saving_$ > expected_call_$`. Compute the call cost from actual model pricing and token usage rather than hard-coding (`~$0.012/call` is today's estimate, not a constant). The gate should account for: expected number of future reuses, provider caching (which changes a saved token's value 10×), current context size, expected compression ratio, extraction model cost, remaining session horizon where estimable, and whether the content is likely to recur.

Where caching already makes extraction pointless, suppress it. On non-caching backends or high-value repeated content, allow it.

### E. Better triggering
No fragile per-workload threshold. Investigate: context pressure, context growth rate, repetition, expected reuse, content type, previous component effectiveness, expected savings, model context-window pressure, amortized economics, and frequency of similar outputs. It must not run every step unnecessarily, and must fire when value is actually likely. Ship sensible defaults.

### F. Metrics
Expose: extraction calls; calls avoided by result cache; calls suppressed by the economic gate; prompt-cache hit/read behavior; extraction LLM cost; gross savings; **net savings after extraction cost**; average latency; global cache hit rate; trigger reason.

Net-after-cost is the honest headline — the current `/stats` shape reports the tool's LLM cost (`metrics/metrics.go:213-215`) separately from savings, so nothing shows the component underwater.

## Relevant code locations

- `components/offload/extract_llm.go` — `:55` trigger, `:62` `llmSeen`, `:66-71` config, `:82-90` the existing caching insight, `:90-108` defaults/construction, `:127-128` cadence, `:277` concurrency.
- `internal/cheapmodel/anthropic.go:16-48` (single-user-message shape — the blocker for A), `openai.go`, `usage.go` (token accounting for the cost model).
- `internal/extract/` — `extract.go:52-54` (`ContentKey`), `cache.go` (result cache), `prompt.go` (the preamble to split), `starlark.go`.
- `components/component.go:101-135` — `Ctx`, `CacheAware` (the signal the economic gate needs).
- `metrics/metrics.go:90-100`, `:213-215` — per-component stats and LLM cost.

## Proposed architecture

1. `cheapmodel` gains a system-block + cache-control capable call shape; Anthropic sets a breakpoint, OpenAI ignores it.
2. `prompt.go` splits into invariant preamble (cached) and per-call variable part.
3. Result cache key becomes `sha256(content) + extractorVersion + model + configFingerprint`, global namespace.
4. An `expectedValue()` helper reading `Ctx.CacheAware`, request size, and observed historical compression ratio, gating the call.
5. Trigger derived from context pressure + recurrence rather than a bare threshold; old keys kept working.

## Alternatives considered

- **Delete `extract_llm`.** Defensible on today's numbers, and the honest fallback if A–E do not get it above water. But it is the only component that can compress novel prose/log shapes no deterministic rule anticipates, so fix it first.
- **Keep it but off by default on caching backends.** Effectively what the economic gate does, expressed as a default rather than a computation. Weaker: it cannot exploit the cases where extraction *is* worth it.
- **Batch multiple outputs per call.** Amortizes the preamble, but couples unrelated outputs and complicates attribution. Consider only if A does not deliver.

## Observability implications

Everything in F, plus: expose the **trigger reason** per activation (an operator's first question is "why did this run?"), and log a suppression reason when the gate declines. Feed the dashboard's per-component economics panel (**#30**) so it is obvious whether the component earns its place.

Result cache namespace changes from session-scoped to global. Bound it. On a key-schema change, old entries must be inert (miss), never mis-served — version the key.

## Backward compatibility

Existing configs must keep working; `min_tokens` stays honored if set explicitly, with the smarter trigger as the default when it is not. Changing `codesmart`'s behavior is benchmark-visible and must be measured. Cache-key change invalidates existing entries once — acceptable, note it.

## Configuration design

Prefer defaults over knobs. Retain `model`, `strategy`; keep `min_tokens`/`trigger` as explicit overrides; add nothing that is only meaningful to someone who has read the source.

## Testing plan

- Unit: the preamble is sent as a cached system block; the request shape is correct for Anthropic and degrades cleanly for OpenAI.
- Unit: result cache hit across two different sessions with identical content.
- Unit: a prompt/extractor version bump misses rather than serving stale.
- Unit: the economic gate suppresses when cache-aware and the output is small; permits on a non-caching backend; permits for high-reuse content.
- Unit: the cost model matches a known token count × known price.
- Unit: the trigger does not fire every step on a growing context.
- `go test -race` (the component runs concurrent per-output calls, `:277`).

## Real-world benchmark plan

2–3 Terminal-Bench tasks where `extract_llm` was most active (from the `codesmart` dump) and 2–3 SWE-bench tasks. Record commit SHA, config, task ids, extraction calls, calls avoided, calls suppressed, prompt-cache reads, extraction cost, gross and unique tokens saved, **net after cost**, latency added, plus all four token tiers, total billed cost, steps, and reward. The bar is explicit: **is `extract_llm` economically positive after these changes?** If it is not, say so and recommend disabling it by default.

## Acceptance criteria

- [ ] Preamble prompt-cached; measured effect reported (tokens + dollars).
- [ ] B prototyped and a documented decision recorded, with numbers.
- [ ] Global content-hash result cache with a version-inclusive key; cross-session reuse demonstrated.
- [ ] Economic gate implemented from real pricing; suppresses when uneconomic; reason logged.
- [ ] Trigger works with defaults, no per-workload tuning.
- [ ] All F metrics exposed, including net-after-cost.
- [ ] Benchmarked on real tasks; a clear verdict on whether the component is now positive.
- [ ] `go test -race` green.

## Documentation updates

`docs/components/extract_llm.md` (rewrite — economics, triggering, caching, the honest verdict), `docs/reference/config.md`, `docs/reference/presets.md` if `codesmart`/`agent`/`general` change, and update improvement plan §B3 from hypothesis to result.

## Dependencies

Independent of the others in implementation, but its benchmark numbers are cleaner after **#25** removes cache-write noise. Feeds **#30** (per-component economics). Its global-cache design should be contrasted with **#27**'s deliberately session-scoped index.

OsherElhadad:
Closes #28

`extract_llm` is the only component that spends money to save money, and on Terminal-Bench it lost: **271 calls, $3.26, ~1,592 s of added latency** against ~197,548 unique tokens saved — **~8× underwater** once those tokens are priced at the rate they would actually have been billed. ~93% of its realized value came from the replay cache, not the LLM.

The cause is arithmetic. A request to a caching backend is ~99.95% cached, so a token removed from a cached region saves the **cache-read** rate (`$0.30/MTok`), not the fresh-input rate (`$3/MTok`) — a **10× haircut**.

## What's implemented

**A. Prompt-cache the preamble — and the measurement that matters.** The ~1,463-token invariant contract now goes in a stable `system` block with `cache_control` (a leading system message on the OpenAI backend, which has no explicit breakpoints). **Measured against the gateway, this is INERT on `claude-haiku-4-5`**, whose minimum cacheable prefix is 4096 tokens — a sub-minimum breakpoint is silently ignored:

| Prefix | Model | Result |
|---|---|---|
| ~1.5k | `claude-haiku-4-5` | `write=0 read=0` — **inert** |
| ~4.5k | `claude-haiku-4-5` | `write=5401` then `read=5401` |
| ~1.5k | `claude-sonnet-5` | `write=2653` then `read=2653` |

Shipped because it is free and correct where it wins, but `/stats` exposes `prompt_cache_read_tokens` so nobody infers a cache win from placement.

**B. Reusing the agent's cached prefix — PROTOTYPED AND REJECTED, with numbers.** It works mechanically (a 103,019-token prefix read from cache, no write, no invalidation) but cache-read is cheap, not free:

| Prefix size | Cost of one extraction | vs a dedicated call ($0.004) |
|---|---|---|
| 103,019 tok | $0.03398 | 8.5× |
| 500,000 tok | $0.15307 | 38.3× |
| 1,700,000 tok | $0.51307 | **128.3×** |

At ~1.7M contexts with 4 concurrent calls/turn that is ~$2.04/turn vs ~$0.016. Plus: risks a cache-write on the agent's own prefix (11.5× a read), and couples the compaction model to the agent model.

**C. Global content-hash result cache.** Re-keyed on `sha256(content + prompt version + model + config fingerprint)` with no session prefix — 82 of 103 unique contents recurred across sessions. A version/model/config change **misses** rather than serving stale. Old entries inert once (noted); a session-scoped entry is honored as a migration read. Contrast #27's deliberately session-scoped index: reference vs derived result.

**D. Economic gate** (on by default). `expected_saving = removed × (1 + expected replays) × per-token value` vs the analytic size-aware cost of a call. Per-token value from `Ctx.CacheAware`; ratio learned from results; cost from real pricing (`CHEAP_MODEL_PRICE_*`), never a constant. Every decision records a reason.

**E. Derived triggering.** Context pressure + growth rate replace `min_tokens`, which stays honored when set explicitly. A merely growing context no longer fires every step.

**F. Metrics.** `/stats` gains an `extract` block: calls, calls avoided, calls suppressed, prompt-cache behavior, cost, gross value, and **NET value after cost** — the honest headline, previously impossible to see because cost lived in a field separate from savings. Purely additive; every existing key keeps its name so `deploy/harbor/*.py` keeps parsing.

## Two deadlocks found by benchmarking

Both the same shape — a pessimistic estimate suppressing every call, so nothing was observed, so the estimate could never correct:

1. **Flat per-call cost** priced everything at the $0.012 average (~5× true cost on small outputs) and suppressed everything. Actual: $0.0024/call. Cost is now **analytic and size-aware**, correct on the first call with zero observations.
2. **Compression ratio** defaulted to 0.45; measured reality is **~0.12** (31–254 tokens removed per call on 400–2,000-token outputs). Corrected — and a **bounded 3-call exploration budget** now lets the tracker learn whether a workload actually compresses.

Also fixed: `/compact` hard-coded the context window as unknown, silently disabling every fraction-based trigger *and* the new pressure logic on that path, so offline eval measured a different component than ships.

## Benchmark results

Replay of real captures, `aws/claude-haiku-4-5`. `forced` = pre-#28, `gated` = new default.

**Terminal-Bench (20 requests):**

| Arm | Backend | Calls | Cost | Saved | **NET** | Avg latency |
|---|---|---|---|---|---|---|
| forced | caching | 4 | $0.0086 | 1,220 | **−$0.0082** | 9,726 ms |
| **gated** | caching | 3 | $0.0065 | 0 | **−$0.0065** | 6,966 ms |
| forced | non-caching | 6 | $0.0109 | 6,664 | **+$0.0091** | 10,287 ms |
| **gated** | non-caching | 7 | $0.0232 | **17,286** | **+$0.0287** | **4,654 ms** |

**SWE-bench (19 requests):**

| Arm | Backend | Calls | Cost | Saved | **NET** | Avg latency |
|---|---|---|---|---|---|---|
| forced | caching | 2 | $0.0098 | 793 | **−$0.0095** | 8,173 ms |
| **gated** | caching | 2 | $0.0092 | 806 | **−$0.0089** | 5,060 ms |
| forced | non-caching | **13** | $0.0593 | 726 | **−$0.0571** | 5,150 ms |
| **gated** | non-caching | **2** | $0.0090 | **2,451** | **−$0.0016** | 3,754 ms |

## The verdict — stated plainly

**On a non-caching backend the gate is a clear win**: net +$0.0287 vs +$0.0091 (3.2×) with 2.6× more tokens saved at half the latency; on SWE-bench it cut 13 calls to 2 and reduced waste **97%**.

**On a caching backend `extract_llm` is still net-negative even after #28.** The gate reduces the loss and cuts latency ~30%, but it cannot make the component pay when saved tokens bill at the cache-read rate and break-even is ~30,500 tokens/output against a largest-observed output of 2,053. **Recommendation: disable it by default on caching backends** — prefer `codesafe` (no LLM pass) or drop `extract_llm` from `codesmart` for caching traffic. This is documented as the verdict rather than dressed up as a win.

Break-even figures are pinned by test, so drift fails CI instead of silently invalidating the documented verdict.

## Verification

- `CGO_ENABLED=1 go build -tags cg_skeleton ./...` ✅
- `CGO_ENABLED=1 go test -tags cg_skeleton ./...` ✅
- `go test -race` ✅ (the component runs concurrent per-output calls)
- `make lint` ✅
- Tests: cached-system-block shape on both backends, cross-session cache reuse, version bump misses rather than serving stale, gate suppression when cache-aware, permission on non-caching and for recurring content, cost model vs known tokens × known price, documented break-even sizes, trigger not firing every step, and regressions for both deadlocks.

## Docs

`docs/components/extract_llm.md` rewritten (economics, triggering, caching, measured tables, honest verdict), `docs/reference/config.md` (pricing env), `docs/reference/presets.md` (`codesmart`/`aggressive` behavior change + verdict).

Assisted-By: Claude Opus 5

### DIFF
--- a/components/all/llm_test.go
+++ b/components/all/llm_test.go
@@ -135,7 +135,11 @@ func TestSummarizeEmptyResponseSkips(t *testing.T) {
 // TestExtractRLMUsesModel: strategy=rlm currently maps to code and still runs the
 // model's filter (not silently deterministic).
 func TestExtractRLMUsesModel(t *testing.T) {
-	off := newComp(t, "extract_llm", "strategy: rlm\nmin_tokens: 1\nmodel:\n  source: config\n")
+	// economic_gate: false — this is a MECHANISM test (does the model-written filter
+	// run and reduce?), and its small fixture output is genuinely uneconomic, so the
+	// #28 gate would correctly suppress the call. Gate economics are tested in
+	// components/offload/extract_econ_test.go against the dollar figures directly.
+	off := newComp(t, "extract_llm", "strategy: rlm\nmin_tokens: 1\neconomic_gate: false\nmodel:\n  source: config\n")
 	st := store.NewMemory(store.Options{})
 	pad := strings.Repeat("padding ", 40) // so reduction beats the marker cost (D1 guard)
 	body := `[{"id":1,"name":"keep this one ` + pad + `"},{"id":2,"name":"drop it ` + pad + `"}]`
@@ -157,7 +161,11 @@ func TestExtractRLMUsesModel(t *testing.T) {
 // TestExtractCodeUsesModel: the code strategy runs the model's Starlark filter and
 // keeps only the matching records (a contained subset), with a marker.
 func TestExtractCodeUsesModel(t *testing.T) {
-	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\nmodel:\n  source: config\n")
+	// economic_gate: false — this is a MECHANISM test (does the model-written filter
+	// run and reduce?), and its small fixture output is genuinely uneconomic, so the
+	// #28 gate would correctly suppress the call. Gate economics are tested in
+	// components/offload/extract_econ_test.go against the dollar figures directly.
+	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\neconomic_gate: false\nmodel:\n  source: config\n")
 	st := store.NewMemory(store.Options{})
 	pad := strings.Repeat("padding ", 40) // so reduction beats the marker cost (D1 guard)
 	body := `[{"id":1,"name":"keep this ` + pad + `"},{"id":2,"name":"drop this ` + pad + `"},{"id":3,"name":"keep that ` + pad + `"}]`
@@ -215,7 +223,11 @@ func TestDeterministicExtractCollapsesRepeats(t *testing.T) {
 // extract_llm with no model available is a clean no-op (deterministic collapse is a
 // separate component now — extract_llm never silently falls back to it).
 func TestExtractLLMNilModelSkips(t *testing.T) {
-	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\nmodel:\n  source: config\n")
+	// economic_gate: false — this is a MECHANISM test (does the model-written filter
+	// run and reduce?), and its small fixture output is genuinely uneconomic, so the
+	// #28 gate would correctly suppress the call. Gate economics are tested in
+	// components/offload/extract_econ_test.go against the dollar figures directly.
+	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\neconomic_gate: false\nmodel:\n  source: config\n")
 	st := store.NewMemory(store.Options{})
 	body := strings.Repeat("some log line\n", 40)
 	req := &bschemas.BifrostChatRequest{Input: []bschemas.ChatMessage{
--- a/components/all/reuse_test.go
+++ b/components/all/reuse_test.go
@@ -78,7 +78,11 @@ func TestSummarizeReusesCheckpoint(t *testing.T) {
 // TestExtractReusesResultCache: the same large tool output re-sent on a later turn
 // reuses the prior compaction — no second model call — and is still reduced.
 func TestExtractReusesResultCache(t *testing.T) {
-	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\nmodel:\n  source: config\n")
+	// economic_gate: false — this is a MECHANISM test (does the model-written filter
+	// run and reduce?), and its small fixture output is genuinely uneconomic, so the
+	// #28 gate would correctly suppress the call. Gate economics are tested in
+	// components/offload/extract_econ_test.go against the dollar figures directly.
+	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\neconomic_gate: false\nmodel:\n  source: config\n")
 	st := store.NewMemory(store.Options{})
 	filter := "data = json.decode(INPUT)\nOUTPUT = json.encode([r for r in data if \"keep\" in r[\"name\"]])\n"
 	cm := &countingModel{resp: filter}
--- a/components/all/xglobal_test.go
+++ b/components/all/xglobal_test.go
@@ -119,3 +119,104 @@ func TestExplicitMinTokensConfigStillReduces(t *testing.T) {
 		t.Fatalf("explicit min_tokens must still reduce (skipped=%v calls=%d)", rep.Skipped, cm.calls)
 	}
 }
+
+// Cross-session reuse must be gated on RECOVERABILITY. In the default rewrite mode the
+// containment proof is deliberately skipped, so a cached result can be a lossy rewrite
+// steered by ANOTHER session's goal. That is tolerable only while `expand` can recover the
+// original — with marker_mode: off there is no way back, so a second session must NOT
+// inherit the first session's lossy rewrite.
+func TestNoCrossSessionReuseOfIrreversibleRewrite(t *testing.T) {
+	off := newComp(t, "extract_llm",
+		"strategy: code\nmin_tokens: 1\neconomic_gate: false\nmarker_mode: off\nmodel:\n  source: config\n")
+	st := store.NewMemory(store.Options{})
+	filter := "data = json.decode(INPUT)\nOUTPUT = json.encode([r for r in data if \"keep\" in r[\"name\"]])\n"
+	cm := &countingModel{resp: filter}
+	pad := strings.Repeat("padding ", 40)
+	body := `[{"id":1,"name":"keep this ` + pad + `"},{"id":2,"name":"drop this ` + pad + `"}]`
+
+	runIn := func(session string) {
+		req := &bschemas.BifrostChatRequest{Input: []bschemas.ChatMessage{
+			userMsg("find the keep records"), toolMsg(body),
+		}}
+		c := &components.Ctx{Ctx: context.Background(), Session: session, Store: st,
+			Model: components.ModelSpec{Static: cm}}
+		var rep components.Report
+		if _, err := off.Offload(req, &rep, c); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	runIn("session-A")
+	first := cm.calls
+	if first == 0 {
+		t.Fatal("first session should have called the model")
+	}
+	runIn("session-B")
+	if cm.calls == first {
+		t.Fatal("an irreversible lossy rewrite must NOT be reused across sessions " +
+			"(no expand path to recover the original)")
+	}
+}
+
+// The same content in the SAME session must still be reused even when irreversible — that
+// costs nothing extra and keeps the request prefix byte-stable.
+func TestSameSessionReuseStillWorksWhenIrreversible(t *testing.T) {
+	off := newComp(t, "extract_llm",
+		"strategy: code\nmin_tokens: 1\neconomic_gate: false\nmarker_mode: off\nmodel:\n  source: config\n")
+	st := store.NewMemory(store.Options{})
+	filter := "data = json.decode(INPUT)\nOUTPUT = json.encode([r for r in data if \"keep\" in r[\"name\"]])\n"
+	cm := &countingModel{resp: filter}
+	pad := strings.Repeat("padding ", 40)
+	body := `[{"id":1,"name":"keep this ` + pad + `"},{"id":2,"name":"drop this ` + pad + `"}]`
+
+	run := func() {
+		req := &bschemas.BifrostChatRequest{Input: []bschemas.ChatMessage{
+			userMsg("find the keep records"), toolMsg(body),
+		}}
+		c := &components.Ctx{Ctx: context.Background(), Session: "one-session", Store: st,
+			Model: components.ModelSpec{Static: cm}}
+		var rep components.Report
+		if _, err := off.Offload(req, &rep, c); err != nil {
+			t.Fatal(err)
+		}
+	}
+	run()
+	after := cm.calls
+	run()
+	if cm.calls != after {
+		t.Fatalf("same-session reuse must avoid a second call, calls went %d -> %d", after, cm.calls)
+	}
+}
+
+// SHIPPING DECISION, end to end: on a cache-aware request the component must decline by
+// default however attractive the economics look, and `allow_on_caching_backend: true` must
+// hand control back to the gate.
+func TestCachingBackendDisabledByDefaultInPipeline(t *testing.T) {
+	filter := "data = json.decode(INPUT)\nOUTPUT = json.encode([r for r in data if \"keep\" in r[\"name\"]])\n"
+	// A big output, well above any break-even, so only the default can be declining it.
+	pad := strings.Repeat("padding word here ", 4000)
+	body := `[{"id":1,"name":"keep this ` + pad + `"},{"id":2,"name":"drop this ` + pad + `"}]`
+
+	runWith := func(cfg string) int {
+		off := newComp(t, "extract_llm", cfg)
+		cm := &countingModel{resp: filter}
+		req := &bschemas.BifrostChatRequest{Input: []bschemas.ChatMessage{
+			userMsg("find the keep records"), toolMsg(body),
+		}}
+		c := &components.Ctx{Ctx: context.Background(), Session: "s", Store: store.NewMemory(store.Options{}),
+			Model: components.ModelSpec{Static: cm}, CacheAware: true, MaxCachedIdx: -1,
+			CtxWindow: 200_000}
+		var rep components.Report
+		if _, err := off.Offload(req, &rep, c); err != nil {
+			t.Fatal(err)
+		}
+		return cm.calls
+	}
+
+	if n := runWith("strategy: code\nmin_tokens: 1\nmodel:\n  source: config\n"); n != 0 {
+		t.Errorf("caching backend must be off by default, got %d calls", n)
+	}
+	if n := runWith("strategy: code\nmin_tokens: 1\nallow_on_caching_backend: true\nmodel:\n  source: config\n"); n == 0 {
+		t.Error("allow_on_caching_backend: true must permit a clearly-economic call")
+	}
+}
--- a/components/all/xglobal_test.go
+++ b/components/all/xglobal_test.go
@@ -1,0 +1,121 @@
+package all_test
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	bschemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/rossoctl/context-guru/components"
+	"github.com/rossoctl/context-guru/schema"
+	"github.com/rossoctl/context-guru/store"
+)
+
+// TestExtractResultCacheHitsAcrossSessions is the headline acceptance criterion for issue
+// #28 part C: identical content in a DIFFERENT session must reuse the prior extraction
+// instead of paying for it again. Before the global re-key the result cache carried a
+// session prefix, so the second session re-derived a result the system already had —
+// measured wasteful on 82 of 103 unique contents.
+func TestExtractResultCacheHitsAcrossSessions(t *testing.T) {
+	// economic_gate: false isolates the CACHE behavior under test from the gate's
+	// (separately tested) spending decision.
+	off := newComp(t, "extract_llm", "strategy: code\nmin_tokens: 1\neconomic_gate: false\nmodel:\n  source: config\n")
+	st := store.NewMemory(store.Options{}) // one store, as a real proxy has
+	filter := "data = json.decode(INPUT)\nOUTPUT = json.encode([r for r in data if \"keep\" in r[\"name\"]])\n"
+	cm := &countingModel{resp: filter}
+	pad := strings.Repeat("padding ", 40)
+	body := `[{"id":1,"name":"keep this ` + pad + `"},{"id":2,"name":"drop this ` + pad + `"}]`
+
+	runIn := func(session string) *bschemas.BifrostChatRequest {
+		req := &bschemas.BifrostChatRequest{Input: []bschemas.ChatMessage{
+			userMsg("find the keep records"), toolMsg(body),
+		}}
+		c := &components.Ctx{Ctx: context.Background(), Session: session, Store: st,
+			Model: components.ModelSpec{Static: cm}}
+		var rep components.Report
+		if _, err := off.Offload(req, &rep, c); err != nil {
+			t.Fatal(err)
+		}
+		return req
+	}
+
+	req1 := runIn("session-A")
+	if cm.calls != 1 {
+		t.Fatalf("first session must call the model once, calls=%d", cm.calls)
+	}
+	out1 := schema.MessageText(req1.Input[1])
+	if strings.Contains(out1, "drop this") {
+		t.Fatal("first session should have reduced the output")
+	}
+
+	// A DIFFERENT session, same content. This is the case the session-scoped key missed.
+	req2 := runIn("session-B-completely-different")
+	if cm.calls != 1 {
+		t.Fatalf("a different session must REUSE the cached extraction (no new model call), calls=%d", cm.calls)
+	}
+	out2 := schema.MessageText(req2.Input[1])
+	if strings.Contains(out2, "drop this") {
+		t.Fatal("cross-session reuse must still drop the non-keep record")
+	}
+
+	// A third, also free.
+	runIn("session-C")
+	if cm.calls != 1 {
+		t.Fatalf("every later session must reuse, calls=%d", cm.calls)
+	}
+}
+
+// The gate must actually suppress in a real pipeline run on a cache-aware request with a
+// small output — the Terminal-Bench losing case, end to end rather than in unit isolation.
+func TestExtractGateSuppressesInPipelineWhenCacheAware(t *testing.T) {
+	off := newComp(t, "extract_llm", "strategy: code\nmodel:\n  source: config\n")
+	st := store.NewMemory(store.Options{})
+	filter := "data = json.decode(INPUT)\nOUTPUT = json.encode([r for r in data if \"keep\" in r[\"name\"]])\n"
+	cm := &countingModel{resp: filter}
+	pad := strings.Repeat("padding ", 40) // ~400 tokens: far below the ~12.7k cached break-even
+	body := `[{"id":1,"name":"keep this ` + pad + `"},{"id":2,"name":"drop this ` + pad + `"}]`
+
+	req := &bschemas.BifrostChatRequest{Input: []bschemas.ChatMessage{
+		userMsg("find the keep records"), toolMsg(body),
+	}}
+	c := &components.Ctx{Ctx: context.Background(), Session: "s1", Store: st,
+		Model: components.ModelSpec{Static: cm}, CacheAware: true, MaxCachedIdx: -1,
+		CtxWindow: 1_000_000}
+	var rep components.Report
+	if _, err := off.Offload(req, &rep, c); err != nil {
+		t.Fatal(err)
+	}
+	if cm.calls != 0 {
+		t.Fatalf("a small output on a cache-aware request must not be worth a call, calls=%d", cm.calls)
+	}
+	if schema.MessageText(req.Input[1]) != body {
+		t.Fatal("a suppressed candidate must be left verbatim (fail open)")
+	}
+}
+
+// Backward compatibility: an existing config that pins min_tokens must keep working
+// unchanged — the smarter trigger is the DEFAULT only when nothing was configured.
+func TestExplicitMinTokensConfigStillReduces(t *testing.T) {
+	// A pinned min_tokens plus the pre-#28 gate setting reproduces old behavior exactly.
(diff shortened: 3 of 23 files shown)


OsherElhadad (commented):
## Independent review — verdict: **do not merge as-is.** Fix H1–H3 first.

I re-derived the economics myself and ran the suite. The headline conclusion holds and the
engineering is mostly careful, but there are three defects that change what the numbers mean —
one of which makes the gate's own reuse signal fire on itself.

### What I verified independently (not taken on trust)
- **Break-even arithmetic is CORRECT.** Re-derived from scratch: `cost(s) = (1463 + min(s,5000) + 200)/1e6 × $1.00 + 200/1e6 × $5.00`, value `= s × 0.12 × 7 × per-token`. At $0.30/MTok cached I get **30,500 tokens**; at $3.00/MTok fresh, **1,800**. Both match the docs and the pinned test exactly. `TestBreakEvenSizesMatchTheDocumentedVerdict` genuinely fails on drift (I confirmed the bands are tight: 26k–35k against an actual 30,500, and it recomputes rather than asserting a constant).
- `go build -tags cg_skeleton ./...` ✅ · `go test -tags cg_skeleton ./...` ✅ · `go test -race -count=5` on `components/... internal/extract/... internal/cheapmodel/... metrics/...` ✅ (no races; `ratioTracker` is mutex-guarded and `exploring()` correctly increments under the lock) · `make lint` (go vet) ✅.
- No credentials, tokens, or private gateway URLs introduced by this diff. The `sk-proxy` literals and IBM gateway hostnames in `deploy/harbor/*` are pre-existing and are placeholders/base URLs, not secrets. PR body is clean.
- `/stats` is additive-only; `TestSnapshotJSONKeysAreBackwardCompatible` pins every key `deploy/harbor/measure.py:140-146` and `analyze.py:90` actually read. `extract` is `omitempty`. Confirmed by grep of the readers.
- `min_tokens` back-compat: the raw-YAML `probe` before defaults is the right technique, and `codesmart` sets `min_tokens: 3000` explicitly (`config/config.go:160`), so the shipped default preset keeps its old trigger. Good.
- `SystemModel` degrades correctly: `completeSplit` concatenates for a non-implementing client, `Complete` omits `system` entirely, OpenAI invents no `cache_control`. Anthropic request shape is valid against the real API contract, and one system block cannot exhaust the 4-breakpoint cap (this is a separate request from the agent's, so no interaction with `cacheinject`).
- **The A finding is plausible and I'd accept it.** haiku-4-5's minimum cacheable prefix is 4096 tokens; sonnet-class is 1024. A 1,463-token preamble is therefore genuinely inert on haiku and genuinely caches on sonnet. `write=0 read=0` is exactly the documented failure mode (silently ignored, no error). `prompt_cache_read_tokens` is properly wired end-to-end (`anthropic.go` → `recordUsageCache` → `CacheUsage()` → `proxy.go:571` → `/stats`), and OpenAI's `cached_tokens` is correctly subtracted out of `prompt_tokens` so the two backends mean the same thing.

**I independently agree with "net-negative on caching backends."** The 15× gap between a 30,500-token break-even and a 2,053-token largest-observed output is not a tuning problem — it is structural, and no prior correction closes it.

---

## H1 (high) — `markSeenContent` poisons the gate's own recurrence signal

`components/offload/extract_llm.go:388` marks content seen **immediately after the gate allows it**, in the same pass. But `hasSeenContent` is the input to `expectedReuses(seenBefore=true) → 6` vs `4`, a 40% valuation bump — and the mark is written on *first* sight, before any evidence of recurrence exists.

The mark is also global and, per `state.go:80`, has no TTL of its own beyond the store's. So the first encounter of any content permanently reclassifies it as "recurring", and every later encounter — in any session — reads at the inflated 6-reuse rate. That inflation is *only* visible in the direction that spends money.

Worse, `TestGatePermitsHighReuseContent` is built on exactly this distinction (34,000 tokens is chosen to sit between the recurring and non-recurring break-evens), so the one test that exercises recurrence never exercises the production path that sets the flag.

Concretely: `evaluateGate` decides at 34,000 tokens that recurring content pays and non-recurring does not. In production every candidate that survives to the gate is marked, so on its second appearance a 34,000-token output that has recurred exactly zero times reads as `seenBefore`. The gate then permits a call the arithmetic says loses.

**Fix:** mark content seen when it is *observed*, not when it is *selected* — move `markSeenContent` above the gate (or above the floor check) so the flag means "encountered before" rather than "the gate liked it before". Better: store a count and require ≥2 sightings before treating it as recurring, since "seen twice" is the actual evidence `expectedReuses` claims to be using. As written the docs' 82/103 cross-session recurrence rate is being asserted a priori rather than measured.

## H2 (high) — the exploration budget is per-process, not per-session or per-workload

`maxExploreCalls = 3` (`extract_econ.go:307`) is enforced on `ratioTracker.explored`, and the tracker is a field on the `ExtractLLM` value built once by `config.Build` and held for the process lifetime (`proxy.go:108` stores one `*Pipeline`). So:

1. **The budget is global across every session the proxy ever serves.** A long-lived proxy spends its 3 exploration calls on the first session's traffic and every subsequent session inherits a prior it can never revise. That is precisely the self-justifying-prior failure the PR says it fixed — the fix bounds the loop but reintroduces it at process scope.
2. **`minRatioSampleTokens = 1500` is reached by one medium output.** A single 2,000-token candidate ends exploration permanently, so the "learned" ratio is an n=1 estimate for the entire process. `TestGateExploresThenSettles` asserts the budget is bounded but never asserts it is *sufficient*, and `TestRatioTrackerLearnsFromObservations` deliberately uses a 20,000-token observation, so nothing covers the n=1 case.
3. **The ratio is a single global mean.** Two interleaved workloads — one compressible JSON traffic, one incompressible file reads — average to a number wrong for both, with no way to separate them.

Note the direction of the over-fire risk you asked me to check: `ratio()` is `removed/total` over *accepted* extractions only. A workload where the first observation compresses well (say a JSON array at 0.5) latches a ratio 4× the measured prior, which multiplies expected saving 4× and drops the cached break-even from 30,500 to ~7,000 — for every session afterwards. The gate can absolutely over-fire; nothing bounds `ratio()` from above, and there is no test that a high early observation doesn't blow the break-even open. **Add a cap** (`min(observed, 0.45)` would bound it at the old prior) and scope the tracker per session, or require a real sample (say 10k tokens / 5 calls) before the observed ratio displaces the default.

## H3 (high) — `/stats` net-value figure attributes the whole pipeline's savings to `extract_llm`

`proxy.go:574` computes `grossValue` from **`snap.SavedTokens`** — the pipeline-wide `before − after` across *every* component (`metrics.go:265`). `format`, `dedup`, `failed_run`, `cmdfilter`, `mask`, `extract`, `collapse` all contribute. Cost, correctly, is `extract_llm`'s alone.

So the "honest headline" is inflated by however much the free deterministic components saved — which on `codesmart` is most of it. On the very benchmark this PR reports, `extract_llm` contributed a small fraction of total savings; `net_value_usd` would show the component comfortably positive while the component-scoped arithmetic in this same PR proves it is negative. That inverts the PR's own conclusion in the one field an operator would actually read.

`metrics/extract.go` already tracks the right number — `xGrossSaved`, fed by `RecordExtractionSaving`. Use `s.GrossSavedTokens × perSavedTok`, not `snap.SavedTokens`. Note `TestExtractSnapshotExposesAllCounters` passes `grossValue` as a literal, so no test catches the wrong source at the call site — add one asserting the value derives from the extract-scoped counter.

Two smaller problems in the same block: `agentFreshPerMTok`/`agentCacheReadPerMTok` are duplicated as fresh constants in `proxy.go:583-586` with a comment explaining the two layers "may legitimately be priced differently" — but nothing can price them differently, since both are compile-time constants that must agree for the numbers to be comparable. And the `CacheMode == "off"` test misses `"auto"`-resolved-to-off, so a non-caching request under `auto` is valued at the cached rate.

---

## Medium

**M1 — `Cfg.MinKeepRatio` is in the fingerprint but `Floor` is doubly volatile.** `cfgFingerprint` includes `Floor` (`extract.go:130`), and `extract_llm.go:255` sets `extCfg.Floor` from `pressureFloor(...)`, which varies with **context pressure** every request. So the global cache key changes as the window fills: 0.6%→0.3%→0.15%→0.05% of window are four distinct keys for identical content and identical semantics. The cache cannot hit across pressure bands, silently discarding much of the reuse this PR exists to capture. Floor only gates *which candidates are attempted*, not what a given result means — drop it from the fingerprint, or snap it to a coarse bucket.

I could not construct two configs with different extractions and the same key — the fingerprint covers every `Cfg` field that reaches a strategy. The bug is the opposite: over-keying.

**M2 — the version bump is manual, and the marker mode is missing from the key.** `PromptVersion = "v2"` with a comment saying BUMP THIS. Nothing enforces it — a `codeRules` edit with no bump serves stale extractions forever, and the comment is the only guard. A cheap mechanical fix: derive the version from `sha256(codeContract + codeRules + codeDeletionRules + codeExample)[:8]` so it bumps itself. Separately, `e.mode` (`parseMarkerMode`) is **not** in the key even though `apply()` uses it to build the replacement text — two marker modes over the same content produce different output under one key. It's applied outside the cached value, so not stale-serving today, but it is a latent trap.

**M3 — cross-session reuse is safe, but only because of an invariant nothing tests.** I traced this properly. The cached value is `out[k].projected`, produced by `runStarlark` → `execStarlarkSummary` executing a model-written filter **over the tool output alone**. The Starlark program's only input is `INPUT` = that content. `goal` and `keepIDs` shape the *program*, not the recorded result, and both are derived from the requesting session's conversation. So: the program is session-influenced, and its output is cached globally under a key that does not include `goal`.

That is defensible for a deletion-only projection (`IsContained` proves the result is a subset of the content — no foreign bytes can enter), and it is the right call. **But `rewrite` defaults to `true`** (`extract_llm.go:136`), and `validateExtraction` explicitly skips the containment proof in rewrite mode (`extract.go:268`). In the default configuration a model can therefore *reword* content under the influence of session A's goal, and that rewording is served verbatim to session B. It is derived from B's own bytes so it isn't a disclosure, but it is A's interpretation of B's data — and the summary (`putSummaryGlobal`) is free-form model prose with no containment check at all.

Not a blocker, but the tenancy argument in the docs claims "context-free derived result" and that is only true in deletion-only mode. Either exclude `rewrite: true` results from the global namespace, or state the narrower guarantee. Add a test that a `goal` difference cannot change a cached deletion-only result — the reuse safety case currently rests on it and nothing pins it.

**M4 — PR #37's 17.8s/0-tokens-saved case is only partly prevented.** I cross-checked. The gate would suppress most of those calls (small outputs, cache-aware ⇒ below break-even), and `RecordExtractionCall` now makes the latency visible, so this is real progress. But three gaps remain: the **exploration budget deliberately permits up to 3 calls that the arithmetic says lose**, each up to `llmCallTimeout = 15s` (`extract_llm.go:29`) — that alone can reproduce ~17.8s across 2 calls, which is exactly the reported symptom; the gate prices dollars and never wall-clock, so a task against its time budget has no lever; and `llmConcurrency` calls can run in parallel per request. **Add a latency circuit-breaker**: skip the LLM path when observed mean latency exceeds a threshold, or make exploration respect a deadline. The gate as built cannot prevent the failure #37 found.

**M5 — the methodology gap is real and the verdict partly rests on it.** Deterministic replay isolates economics cleanly and I'd defend it for the cost question. But it cannot measure reward, and `rewrite: true` (the default) is *lossy and unverified by construction*. So the component's headline risk — removing content the agent needed — is unmeasured, while the PR's recommendation (disable on caching backends) happens to be the risk-reducing direction. For a *disable* recommendation that asymmetry is acceptable and I'd let it ship. It is not acceptable as support for the "clear win" on non-caching backends: that arm recommends *more* LLM rewriting on reward-unmeasured evidence. Either soften that claim or land one Harbor task with reward.

**M6 — the `/compact` window fix is right but untested.** `proxy.go:220-230` is a genuine and important fix; without it offline eval measured a component that never fires pressure logic. There is no test that `/compact` resolves a window (no `Windows` reference in `proxy/*_test.go`), so it can silently regress to `0` and re-invalidate every future replay-based measurement — the exact class of bug being fixed. Also note `h.opts.Windows == nil` still yields `window = 0`, and `replay.py:106`/`replay2.py:153` both drive `/compact`, so whether the benchmark numbers in this PR reflect the fix depends on `Windows` being wired in the replay harness. Worth stating explicitly in the PR.

## Low

- **L1** `metrics/extract.go` counters are process-global while `xReasons` is unbounded — reasons are a closed set today, but `RecordExtractionReason` is exported and takes an arbitrary string.
- **L2** `round4` in `metrics/extract.go:152` reimplements rounding via int64 truncation; `math.Round(f*10000)/10000` is one line and correct at the boundaries.
- **L3** `analyticBaseline` hardcodes 2000 tokens as the reconciliation denominator; the `0.1 < ratio < 10` guard is sensible but the magic constant deserves a name.
- **L4** `prevTokens` (`extract_llm.go:75`) grows unbounded — one entry per session for process lifetime. Same for `llmSeen`. Pre-existing for the latter; the new map compounds it.
- **L5** The docs are genuinely good and describe implemented behaviour including the negative verdict — a credit worth stating. I could not run `mkdocs build --strict` (mkdocs not installed here), so that check is unverified by me.
- **L6** The cache-key change invalidating existing entries once *is* noted (`extract.go` comment + docs). Old entries miss rather than mis-serve: keys are `cg:xres:`-prefixed and `k1`-versioned, and the pre-#28 session-scoped migration read goes through the old `resultKey` path. That part is correct.

---

### On the shipping question

Ship the **verdict and the measurement infrastructure** — they are the valuable part, and the willingness to publish a negative result about your own component is the right instinct. But the component should be **off by default on caching backends now**, not merely documented as inadvisable: `codesmart` is the proxy default and still contains `extract_llm`, so the default configuration ships a component this PR proves is net-negative, with a doc note as the only guard. A `codesafe`-style default for caching traffic — or a hard `CacheAware ⇒ skip unless min_tokens is explicit` — matches the evidence better than a recommendation in prose.

The `codesmart`/`aggressive`/`agent`/`general` presets all still list it. If the verdict is real, the presets should reflect it.

OsherElhadad:
## Orchestrator verification of the fix round — and the headline retraction is the important part

All three high-severity defects are fixed and I verified the two structural ones:

- **H3** — `metrics/extract.go:133` now takes `perSavedTokenUSD` (a **rate**) and applies it internally to `GrossSavedTokens`. `proxy/proxy.go:581` passes the rate, not a pre-multiplied total. Passing a pipeline-wide figure is no longer *expressible*, which is the right shape for this fix — a regression test can be bypassed, an impossible API cannot.
- **Shipping decision** — `AllowOnCachingBackend` in `components/offload/extract_llm.go:100` with the decline in `extract_econ.go`. This is the guard being **in code, not prose**, which is what I asked for.

Full suite green, `gofmt` clean.

### The retraction matters more than the fix

Fixing H3 **invalidated this PR's original headline.** The reported "+$0.0287 net win" on Terminal-Bench was *produced by the bug* — pipeline-wide savings credited to one component. With honest attribution the best result on the available captures is **break-even, not profit.**

That is now **three** of this PR's benchmark results traced to measurement artifacts: the flat $0.012/call cost, the 0.45 compression-ratio prior, and this attribution error. Each pointed the same direction — making the component look better than it is — and each was found by continuing to check rather than by stopping at a favourable number.

I want to be precise about what this PR now establishes, because it is a genuinely useful negative result and it should not be softened:

> On prompt-caching backends `extract_llm` cannot pay for itself. Break-even needs ~30,500 tokens per output against a largest-observed 2,053 — a 15× structural gap, not a tuning problem. The component now makes **zero calls and costs nothing** there. On non-caching traffic the gate is a strict improvement in every arm (68% less waste on TB; 26 calls → 1 on SWE), but still only reaches break-even on these captures. A genuine positive result requires outputs above ~1,800 tokens on non-caching traffic, which neither capture contains.

### Preset note

`extract_llm` stays listed in `aggressive`/`agent`/`general`/`codesmart`. That is now **correct** and better than removing it: the component itself declines on caching backends, so the guard travels with the component rather than depending on which preset someone picked or whether a future preset edit remembers the constraint. A user on non-caching traffic still gets the (gated, strictly-improved) behaviour.

### On the medium items

All three taken, and the reasoning on cross-session reuse is better than what I asked for: gating on **recoverability** rather than restricting to deletion-only correctly identifies that the risk is a lossy rewrite steered by another session's goal, and that reversibility is the actual mitigation — with same-session fallback when the marker mode or store can't guarantee it. Deriving `PromptVersion` from a hash of the prompt constants closes the footgun I flagged; keeping `semanticsVersion` manual for validation-gate changes is the right split.

The 6 s latency brake also addresses #37's measured failure (17.8 s / 2 calls / 0 tokens saved contributing to a wall-clock timeout), which was the cross-PR concern.

Verdict from my side: ready to merge once the remaining reviewer comments are addressed. The component ships honest, disabled where it loses money, and instrumented so the next person can tell.

OsherElhadad:
Superseded by #51.

This branch's history was rewritten to rebase onto `main` at `5768b00` (nine merges: #33, #36, #40, #41, #42, #43, #44, #49, #50), and force-pushing a protected branch is blocked — so the work moved to `feat/i28-extract-llm-rebased`. Same pattern as #37 → #42.

#51 carries this PR's content plus:

- **#40's unified result-cache key** adopted on both the session-scoped and global paths — one JSON value per decision, so a replay can never emit half a decision. The global namespace keeps the pair together for the same reason.
- **Cache-safety ordering restored**: `getResult` (depth-safe) → tail gate → `getResultGlobal`. Only the same-session replay may bypass the gate; a cross-session hit at depth would mutate a cached prefix that holds the *original*, which is what #40 removed `repairLostResult` to prevent. `TestGlobalCacheHitIsNotSplicedAtDepth` was confirmed to fail when the ordering is flattened.
- **The shipping guard re-verified**, with a new end-to-end test (`TestNoDefaultConfigRunsExtractLLMOnCachingBackend`) covering the bare defaults and the `codesmart` preset, so `allow_on_caching_backend` cannot be lost silently in a future rebase.
- **Corrected economics.** The final Terminal-Bench arm excluded `extract_llm` entirely and independently re-derived them: the 197,548 saved tokens sit in the cached prefix, so at cache-read price they are worth $0.0395 against $3.26 and 1,592,467 ms — **82× underwater**, not the 8× in the improvement plan, which priced them as fresh input. The verdict here was right and understated.

Closing in favor of #51. Review discussion above still applies; all of it is carried forward.

PR #35, opened 2026-08-10
feat(proxy): async mode — cache-safe deferred compaction (HELD: benefit unmeasured since #34)

### LINKED ISSUE #25: fix(cache): freeze-store lifetime — sliding TTL, longer default, and no cache-destructive regression on a store miss
## Problem statement

A frozen compaction can expire mid-task, causing context-guru to hand the provider a *different* representation of an already-cached prefix. The provider must then re-write the whole suffix. This is the leading suspect for the Terminal-Bench cache-write regression (**4.01M → 6.53M tokens**).

## Current behavior

`store/store.go:122-136` — `Get` enforces the TTL and refreshes **LRU recency**, but never refreshes `e.expires`:

```go
func (m *Memory) Get(key string) ([]byte, bool) {
	m.mu.Lock()
	defer m.mu.Unlock()
	el, ok := m.items[key]
	if !ok {
		return nil, false
	}
	e := el.Value.(*entry)
	if m.now().After(e.expires) {
		m.remove(el)
		return nil, false
	}
	m.ll.MoveToFront(el)     // <-- recency refreshed
	return e.payload, true   // <-- e.expires NOT refreshed
}
```

`Put` sets `e.expires = m.now().Add(m.ttl)` (`:111`, `:115`); the default TTL is **1800s** (`store/store.go:85`). So an entry dies 1800 s after it was last *written*, no matter how often it was successfully read.

The consequence, via the freeze-and-replay path (`components/offload/state.go:47-81`): `freeze` stores a `(session, component, contentKey) → replacement` decision, and `reapplyFrozen` replays it so a message the provider already cached is re-sent **byte-identically**. When the frozen entry expires, `reapplyFrozen` misses, the component re-decides from scratch (or declines), and the message flips representation — `full → compacted` or `compacted → full` — inside the cached prefix.

Terminal-Bench tasks routinely run **1–4 h** of wall clock. At the measured cadence that is roughly **~69 turns** before a frozen decision dies, and then again every ~69 turns after. A revert at mid-depth re-writes `W` tokens at `$2.50/M` where they would otherwise have been read at `$0.20/M` — a premium of `W × $2.30/M` for **zero** benefit.

## Motivation

Working hypothesis, to be **validated rather than assumed**: eliminating ~15 mid-depth reverts removes on the order of **2.5M cache-write tokens ≈ $6.3**, which would account for most of the observed 4.01M → 6.53M gap. One cache-write costs **11.5 cache-reads** (`($2.50 − $0.20) / $0.20`), so cache-write avoidance is a far larger lever on this workload than content reduction (unique token removal is 0.02–0.13% of the billed total). See `docs/results/improvement-plan.md` §A2.

### A. Sliding TTL
Refresh `expires` on a successful `Get`. A decision that is still actively being replayed is by definition still live; expiring it is what causes the harm.

### B. Raise the default TTL
1800 s is well under a realistic long-horizon agent task. Default to **~10000 s** unless the code review turns up a stronger value. Keep `ttl_seconds` configurable (`store/store.go:64`).

### C. A store miss must never force a cache-destructive regression
This is the design question, not a one-liner. Today `reapplyFrozen` returning `false` means "no frozen decision", which is indistinguishable from "the decision existed and was lost". Investigate the architecture and decide the correct fail behavior. Note the tension with the fail-open invariant in `CLAUDE.md`: fail-open on *compaction* means forwarding the original, but for an already-established compaction the safe direction is the opposite — **keep the representation the provider cached**. Enumerate what state would have to survive for that to be possible, and where.

Related, in the same area: `apply/apply.go:155-163` derives `maxCachedIdx = prevLen(st, sessionID) - 1`, so a `prevLen` store miss yields `-1`, and `Ctx.TailOnly` (`components/component.go:131-135`) then returns **true for every index** — fail-*open*, permitting mutation of the cached prefix. Measured on **11.2%** of Terminal-Bench requests. Consider whether this belongs in this issue or a follow-up; at minimum, document the interaction.

## Relevant code locations

- `store/store.go:105-136` — `Put`/`Get`, TTL and LRU; `:83-96` defaults; `:64` config struct.
- `components/offload/state.go:47-81` — `frozenKey`, `freeze`, `reapplyFrossen`/`reapplyFrozen`.
- `apply/apply.go:155-163`, `:307-315` — `maxCachedIdx`, `prevLen`/`putLen`.
- `components/component.go:101-135` — `Ctx.CacheAware`, `MaxCachedIdx`, `TailOnly`.
- Users of freeze-replay: grep `reapplyFrozen` — only `mask` and `failed_run` currently call it, which is itself a finding (see the "freeze everywhere" note in the improvement plan).

## Proposed architecture

1. Sliding TTL in `Memory.Get` (2 lines).
2. Default `ttl` 1800 s → 10000 s, still overridable.
3. A distinguishable "was frozen, now lost" signal so the caller can choose to preserve the cached representation instead of silently re-deciding. Options to weigh: a separate longer-lived index of `(session, contentKey) → representation-fingerprint`; making frozen entries exempt from LRU eviction while their session is live; or a per-session generation counter that makes a lost decision detectable.

## Alternatives considered

- **Raise the TTL only.** Cheapest, but a long enough task still hits the wall, and LRU eviction under a 1000-entry cap can drop an entry before its TTL anyway.
- **Never expire frozen entries.** Unbounded growth; the cap exists for a reason.
- **Persist to disk/SQLite.** Larger change; may be right eventually, but not required to fix this.

## Observability implications

Emit counters for: frozen-decision hits, misses, misses that caused a representation flip, and estimated cache-write tokens attributable to those flips. Without this the fix cannot be verified in a benchmark run — and per the improvement plan we are currently **blind on our largest cost line**. Surface these in `/stats` under the component and in the new dashboard (#30).

Possible new key namespace alongside `cg:frz:`. Keep bounded. If frozen entries gain eviction exemption, cap that exemption per session so a pathological session cannot pin the whole cache.

## Backward compatibility

Behavior-only change; no config or API break. A longer default TTL raises steady-state memory — quantify it (entries × payload size × 1000-entry cap) in the PR.

## Configuration design

Keep `store.TTLSeconds` / `ttl_seconds`. If a "never revert an established compaction" policy is added, expose it as a documented boolean with a safe default rather than an implicit behavior change.

## Testing plan

- Unit: `Get` on an entry near expiry extends its life; an entry never read still expires; the LRU cap still evicts.
- Unit: a frozen decision survives a simulated long session (fake clock, ≫1800 s of replays).
- Unit: on a forced store miss, an already-established compaction does not flip representation.
- Integration: a synthetic 100-turn session asserting the forwarded prefix stays byte-identical across the old expiry point.
- `go test -race` on the store.

## Real-world benchmark plan

The hypothesis **must** be validated with real runs, not accepted. Pick 2–3 long-horizon Terminal-Bench tasks from the arms where cache-write was worst, plus 2–3 SWE-bench tasks as a no-regression control. Record for each: commit SHA, config, model, task id, fresh input / cache-read / cache-write / output tokens, total billed cost, steps, reward, latency, expands, reverts, and the new frozen-miss counters. Report before/after per task. **Do not cherry-pick**: if a task regresses, investigate it.

## Acceptance criteria

- [ ] `Get` refreshes `expires`; unit-tested with a fake clock.
- [ ] Default TTL raised; still configurable; memory impact quantified.
- [ ] A store miss cannot flip an already-established compaction back to a cache-destructive representation (or, if that proves architecturally wrong, the reason is documented and the alternative implemented).
- [ ] Frozen hit/miss/flip counters exist and are visible in `/stats`.
- [ ] Measured on real tasks: cache-write tokens down, cost down, reward not worse. Numbers in the PR body.
- [ ] `go test -race ./store/... ./components/...` green.

## Documentation updates

`docs/reference/config.md` (TTL default), a note in `docs/design.md` on freeze lifetime and the fail direction for an established compaction, and `docs/results/improvement-plan.md` §A2 updated from hypothesis to measured result.

## Dependencies

None — this is the **first** issue to land; every other benchmark measurement is cleaner once cache-write noise is removed. #30 (dashboard) consumes the new counters. Coordinate with #27 (`xdedup`), which will also want freeze-replay.

### LINKED ISSUE #31: feat(proxy): three operating modes — sync, async (cache-safe deferred compaction), and observe (measure without enforcing)
## Problem statement

Compaction is unconditionally **synchronous**: `proxy/proxy.go:364-373` calls `apply.BodyFull` inline before forwarding, and the caller waits. Measured cost on Terminal-Bench: **~450 ms/req** added, ~1,592 s cumulative on one arm, almost all of it `extract_llm`'s model call.

That forces two bad choices on a user. To get savings they must accept latency on every request. And to find out whether context-guru helps *their* workload, they must enforce it in production and compare against history — there is no way to evaluate it safely first.

## Motivation

- **Latency.** Synchronous compaction is the right default for correctness but the wrong default for a long-lived agent session, where a decision computed once is replayed for many turns. There is no reason the *first* turn must pay for a compaction that benefits turn 5 onward.
- **Adoption.** "Run it and see what it would have saved, without touching my requests" is the single most useful thing a compaction proxy can offer, and neither reference implementation has it: headroom has no observe/shadow/dry-run mode at all (its `token` vs `cache` modes are both enforcing; the only control arm is a 10% output-shaper holdout). This is a genuine differentiator, not a copy.
- **Evidence.** Every conclusion in `docs/results/` came from running full paired benchmark arms. Observe mode makes that comparison available to any user on their own traffic.

### Mode 1 — `sync` (current behavior, remains the default)
Before an LLM request: context-guru compacts, the caller waits, the compacted request is sent. Prioritizes immediate savings; adds compaction latency to the request path.

### Mode 2 — `async`
Compaction happens asynchronously. The agent continues making LLM calls while context-guru computes a new compacted/frozen state; once ready, **subsequent** turns use it.

**The cache requirement is the hard part.** By default, while waiting for async compaction, the newest un-compacted steps must **not** be provider-cached in a way that later forces us to rewrite a cached suffix once the compaction arrives. Design this against real Anthropic prompt-caching semantics: a cache-write costs **11.5×** a cache-read (`($2.50 − $0.20)/$0.20`), so a naive async implementation that lets the uncompacted tail get cached and *then* replaces it converts a read into a write and is strictly worse than `sync`. That failure mode is not hypothetical — it is precisely what tripled headroom's cache-write on Terminal-Bench (12.37M vs baseline 4.01M) by rewriting the live zone.

Configuration must allow changing this, but the **safe default protects cache-write economics**.

Design for: stable prefix · mutable tail · cache breakpoints · compaction generation/version · stale async results arriving after newer turns · newer turns arriving while compaction runs · cancellation and coalescing · **at most one useful compaction job per session/version** · race conditions · session isolation.

Prior art worth studying rather than reinventing: headroom's `BackgroundCompressor` is a single bounded `asyncio.Queue` + one drain task, dedup-by-key with the pending slot claimed **before** the job becomes observable (so dedup is atomic against a concurrent enqueue), no request-coupled deadline, fail-open on every path ("lost savings, never lost correctness"), and a `stats()` exposing `{queued, pending, processed, dropped, errors}`. Its instructive weakness: the dashboard shows only `queued`, so `dropped` — the "we silently gave up savings" counter — is invisible. Expose the whole tuple.

### Mode 3 — `observe`
Context-guru asynchronously computes what it **would** have done and never applies it. The agent receives the untouched original conversation.

Reports: potential tokens saved · potential dollar savings · potential per-component contributions · hypothetical cache implications · hypothetical context-guru overhead · **actual** baseline usage · projected optimized usage.

The dashboard must make it unmistakable: *"You are currently in observe mode. context-guru did not modify requests. Here is what it could have saved."*

This requires careful metric semantics so observed/hypothetical numbers can **never** be confused with actual enforced savings. That is the primary correctness risk of this issue — a mis-labeled hypothetical number is worse than no number.

## Current behavior / relevant code locations

- `proxy/proxy.go:353-383` — the synchronous block: panic-recover backstop, `apply.BodyFull`, `RecordAddedLatency`, `expand.Inject`, then `h.serve`.
- `apply/apply.go:115-126` — top-level fail-open recover; `:127-173` — messages parse, session resolve (`session.Resolve`), `resolveCacheAware`, `maxCachedIdx = prevLen(st, sessionID) - 1` (`:161`) and the deferred `putLen` (`:162`); `:164-173` `Ctx` construction. **`putLen` running in a `defer` is a direct async hazard**: two concurrent turns of one session would race on the per-session length.
- `components/component.go:101-135` — `Ctx.CacheAware`, `MaxCachedIdx`, `TailOnly` (note `MaxCachedIdx < 0` ⇒ `TailOnly` true for **all** indices, i.e. fail-open into mutating the cached prefix — coordinate with **#25**).
- `components/offload/state.go:47-81` — `freeze`/`reapplyFrozen`: the existing "compute once, replay byte-identically" mechanism, which is the natural substrate for async generations.
- `store/store.go` — TTL/LRU; generation state lives here (**#25** governs its lifetime).
- `metrics/metrics.go:70` — one mutex-guarded `Aggregator`; `:196-221` `Snapshot`. Mode must become a dimension here.
- `proxy/proxy.go:68-72` — `Options.InjectExpand`, the precedent for a tri-state option.
- `session/` — session identity.

## Proposed architecture

1. **Mode as an explicit `Options` field** threaded onto `Ctx`, defaulting to `sync`. Never inferred.
2. **A per-session compaction generation.** A request records the generation it was built from. An async result that targets a stale generation is **discarded**, not applied — this is the single most important invariant, and it must be tested with `-race`.
3. **One worker pool, bounded, owned by the proxy** — not per-request goroutines. Enqueue is dedup-by-`(session, generation)` with the pending slot claimed before the job is observable. Bounded queue; drop rather than block; count drops. Cancel on session end; coalesce supersessions.
4. **Async cache policy.** Until a compaction lands, the uncompacted tail must be excluded from the cached region — i.e. do not place a breakpoint at or beyond the not-yet-compacted tail, so the eventual replacement rewrites nothing the provider committed to. `cacheinject` already owns breakpoint placement (`components/reformat/cacheinject.go`), and `apply/prefixsplit.go` shows the body-level precedent for this kind of surgery. Make the policy explicit and configurable, with the cache-safe choice as default.
5. **Observe mode** runs the same pipeline on a **copy**, discards the output, and records results into a strictly separate metric namespace. The enforced path must be provably untouched: assert byte-identity of the forwarded body in a test.
6. **Metrics carry the mode**: at minimum `sync_enforced`, `async_enforced`, `observe_hypothetical`, distinct in the payload and filterable in the dashboard.

## Alternatives considered

- **Async via unmanaged goroutines per request.** Rejected explicitly by the requirement, and rightly: goroutine leaks, stale writes over newer generations, no cancellation, unbounded fan-out under load.
- **Observe mode as a config flag on the existing sync path** (compute, log, discard, still synchronous). Much simpler and worth considering as a **first increment** since it needs none of the async machinery — but it keeps the latency it was meant to avoid, so it is a stepping stone, not the destination.
- **Sampling instead of observe** (compact 10% of traffic). Cheaper but answers a different question and perturbs real requests; the output-shaper holdout in headroom shows the pattern's limits.
- **A separate replay tool over captured traffic.** Zero production risk, and `CONTEXT_GURU_CAPTURE` already supports offline replay — but it cannot report *live* projected savings, which is the point.

## Observability implications

Mode-tagged everything. Observe mode needs its own vocabulary in the payload (`potential_*` / `projected_*`), never sharing a key with an enforced metric. Async needs `{queued, pending, processed, dropped, errors, stale_discarded}` and a "savings realized on turn N+k" figure so deferred value is attributable rather than invisible. Dashboard (**#30**) must filter and compare by mode and display the observe-mode banner.

Per-session generation counters and pending-job state, bounded and session-isolated. Observe-mode results must be stored in a way that cannot be summed into enforced totals — separate columns or an explicit mode column that every aggregate query filters on. Getting this wrong silently inflates the product's headline claim, so it is a correctness requirement, not a nicety.

## Backward compatibility

Default `sync` reproduces today's behavior exactly, byte for byte — assert this with a golden test. `/stats` keeps its current shape (the harnesses in `deploy/harbor/*.py` parse it); new mode fields are additive. Existing configs unchanged.

## Configuration design

```yaml
mode: sync            # sync | async | observe
async:
  cache_uncompacted_tail: false   # safe default: protect cache-write economics
  max_queue: 256
```

One option per real decision. The cache-policy escape hatch exists because a non-caching backend genuinely does not need the protection.

## Testing plan

- **`go test -race` throughout** — this is a concurrency issue and race tests are mandatory, not optional.
- Unit: `sync` output byte-identical to pre-change (golden).
- Unit: observe mode leaves the forwarded body byte-identical while still recording hypothetical savings.
- Unit: an async result for a stale generation is discarded, never applied.
- Unit: two concurrent turns of one session do not corrupt `prevLen`/generation state (today's deferred `putLen` is the hazard).
- Unit: enqueue dedup is atomic against a concurrent enqueue of the same key.
- Unit: a full queue drops and counts, and never blocks the request path.
- Unit: worker cancellation on session end; no goroutine leak (`goleak`-style assertion or an explicit count).
- Unit: with `cache_uncompacted_tail: false`, no breakpoint is placed at/beyond the un-compacted tail.
- Unit: observe-mode metrics cannot be summed into enforced totals (aggregate query test).

## Real-world benchmark plan

2–3 SWE-bench and 2–3 Terminal-Bench tasks **per mode** (9–18 trials), plus a real Claude Code session in each mode. Record per trial: commit SHA, mode, config, model, task id, fresh/cache-read/cache-write/output tokens, total billed cost, context-guru's component cost, steps, reward, latency (**added latency per mode is the headline for async**), expands, reverts, savings, and the async queue counters.

Specific questions to answer with numbers:
- Does `async` actually reduce added latency without increasing cache-write? (If cache-write rises, the cache policy is wrong.)
- Does `async` reach the same steady-state savings as `sync`, just later?
- Does `observe` add measurable latency to the enforced path? (It must not.)
- Do observe-mode projected savings match what `sync` actually achieved on the same tasks? That agreement is the validation of the whole mode — report the discrepancy honestly.

## Acceptance criteria

- [ ] Three modes implemented; `sync` is the default and byte-identical to today (golden test).
- [ ] Async uses a bounded, owned worker pool with generation-versioned results; stale results discarded; dedup/coalescing atomic; cancellation clean; no goroutine leaks.
- [ ] Async's default cache policy provably avoids rewriting a cached suffix; measured cache-write does not exceed `sync`.
- [ ] Observe mode never modifies a request (byte-identity test) and reports the full projected set.
- [ ] Observe/hypothetical metrics are namespace-separated and cannot be aggregated into enforced savings.
- [ ] Mode is a filterable dimension in `/stats` and the dashboard; the observe-mode banner is unmistakable.
- [ ] Async queue counters exposed in full, including `dropped` and `stale_discarded`.
- [ ] `go test -race` green across proxy, apply, components, store.
- [ ] Benchmarked in all three modes on real tasks; observe-mode projections compared against `sync`'s actuals, with discrepancies reported.

## Documentation updates

New `docs/how-to/operating-modes.md` (when to use each, the async cache trade-off, how to read observe-mode numbers), `docs/reference/config.md`, `docs/design.md` (generations, job lifecycle, cache policy, fail-open behavior per mode), `docs/dashboard.md` (mode filtering + observe banner), `mkdocs.yml`, README.

## Dependencies

Interacts with **#25** (frozen-state lifetime is the substrate for async generations; `MaxCachedIdx` fail-open must be resolved or async inherits it). **#27**'s `xdedup` must be lifecycle-compatible with async/observe. **#30** must render all three modes with distinct semantics — coordinate the metric namespace early. Consider landing observe mode first: it is the smaller, lower-risk half and delivers the adoption story without the concurrency surface.

OsherElhadad:
> ## 🚧 DRAFT — held pending a benchmark, not pending code
>
> **Observe mode landed separately in #43.** This branch is async-only, and it is held
> because its measured benefit has evaporated — not because anything here is unfinished or
> known-broken. All seven blocking review findings plus four follow-ups (S1–S11) are fixed;
> gates are green including `-race -count=5`.
>
> **Why held:**
>
> - The headline latency win (1,599 ms → 25 ms) came almost entirely from deferring
>   `extract_llm`, and **#34 hard-declines `extract_llm` on caching backends**. On the
>   primary workload there is now nothing expensive left to defer. Terminal-Bench already
>   measured exactly that: zero LLM calls, async 26.8 ms vs sync 26.9 ms.
> - The cache-write evidence is **withdrawn** (see the corrections section): the tail
>   protection was inert on claude-code, which sets its own breakpoints.
> - `dropped` and `stale_discarded` — the counters guarding the correctness invariant — are
>   exercised by tests only, never by production load.
>
> **What would unblock it:** a 50-task paired arm with `async.strip_caller_breakpoints: true`
> showing (a) cache-write no higher than `sync`'s, and (b) a real latency reduction on
> traffic that still makes compaction model calls. That is a measurement, not a code change.
>
> **The benefit is conditional and currently unmeasured on claude-code.** With default
> settings async declines to defer on every turn where the agent has cache-written the tail,
> so it is deliberately inert rather than falsely protective — `async_tail_unprotected_turns`
> reports that. `strip_caller_breakpoints: true` is required for it to do anything there.
>
> Kept rather than discarded because the hard parts — per-session generations, the bounded
> pool, the copy-on-write store overlay, the decline-rather-than-pretend cache policy —
> survived a hostile review intact. This is the ready-to-measure branch.
>
> **Note for #40/#43:** the `store.Buffer` → `FrozenLost` structural forwarding lives on THIS
> branch, because `store.Buffer` does. #43 does not need it (observe's shadow store is a
> plain `*store.Memory`, which implements `FrozenLost` natively), so nothing is lost by this
> sitting in draft.

Closes #31.

Compaction was unconditionally synchronous, which forced two bad trades: to get
savings you accepted ~450 ms/req of latency, and to find out whether context-guru
helps *your* workload you had to enforce it in production and compare against
history. This adds an explicit `mode:` with three settings.

`sync` remains the default and is byte-identical to today.

## The three modes

| Mode | Request path | Notes |
|---|---|---|
| **`sync`** *(default)* | Runs the pipeline inline; the caller waits. | Unchanged, asserted by a golden test. |
| **`async`** | Runs inline **with no model clients**, replaying decisions an earlier turn froze; enqueues the expensive compaction for later turns. | Cache-safe by default (below). |
| **`observe`** | Does **not** run the pipeline at all. Forwards the original, byte for byte. | A copy runs off-path against a buffer that is never committed. |

```yaml
mode: async
async: { cache_uncompacted_tail: false, max_queue: 256, workers: 1 }
```

## The hard part: async's cache requirement

A cache-write costs **11.5x** a cache-read (`($2.50 − $0.20)/$0.20`). So a naive
async implementation is *strictly worse than sync*: it lets the un-compacted tail
get provider-cached, then replaces it when the compaction lands, and the provider
re-writes a span it had committed to. That is exactly what tripled headroom's
cache-write on Terminal-Bench (12.37M vs 4.01M baseline) by rewriting the live zone.

By default no breakpoint is placed at or beyond the tail a pending compaction will
replace: `cacheinject` drops those positions and anchors at the highest safe index
instead, so the whole stable prefix is still written and nothing the provider
commits to is later rewritten. `async.cache_uncompacted_tail: true` is the escape
hatch for a backend confirmed not to cache, where the protection buys nothing.

> **Review correction.** The first version pruned only the positions `cacheinject` itself
> wanted, so breakpoints the *agent* set survived — and claude-code sets its own on the
> newest message, inside exactly the protected span. The protection was a silent no-op on
> the primary workload: async paid the 11.5x rewrite *and* lost a slot while reporting
> success. It now either strips those (`async.strip_caller_breakpoints: true`) or
> **declines to defer the turn**, counted as `async_tail_unprotected_turns`. With
> claude-code on defaults, async is deliberately inert rather than falsely protective.

The protection needed its own bool rather than a sentinel index: index 0 is a
legitimate value ("no breakpoint anywhere"), so no integer is free to mean "off".
A false default also makes an unset field cost a missed optimisation rather than a
wrong request — the opposite of `MaxCachedIdx`'s `-1` (see #25).

## Generations: the invariant everything rests on

An async result lands at an unpredictable later moment, by which time the agent may
have taken another turn and another job may have committed. Applying a result built
from a superseded snapshot is how a compaction proxy corrupts a cached prefix.

`modes.Tracker` holds, per session under one lock, the compaction generation — which
advances on **every turn** — and the cached-prefix boundary. A request records the
generation it was built from; a deferred job writes into a `store.Buffer` (copy-on-write
over the real store) and that buffer is committed **only** if the session is still on that
turn — checked under the same lock that advances it. A stale result is discarded and
counted as `stale_discarded`.

> **Review correction.** The generation originally advanced only on *commit*, so a job from
> turn 1 still read its own generation as current after eight later turns and committed
> against a long-superseded transcript: the guard could only fire on a dedup collision,
> never on staleness. The documented invariant was not the implemented one. Fixed, with the
> honest consequence now stated in the docs — at agent turn rates async discards much of
> what it computes, and `stale_discarded` is the number that tells you whether deferral
> suits a workload.

The buffer is what makes "discard" possible at all: a deferred run writes frozen
decisions and stashes as it goes, so throwing the result away after the fact is not
an option. Running it against an overlay makes the whole result one atomic,
discardable unit — `Commit()` flushes it, and not calling `Commit()` is the discard.

## The worker pool

One bounded queue plus a fixed worker count owned by the proxy, not a goroutine per
request. The shape is headroom's `BackgroundCompressor`, ported and extended:

- dedup by `(session, generation)`, pending slot claimed **before** the job is
  observable in the queue, so dedup is atomic against a concurrent enqueue;
- a full queue **drops** and counts, never blocks — the request already went out;
- jobs run under the pool's context, not the request's (which is cancelled when the
  response is written);
- fail-open on every path, including a panicking job;
- the **whole** counter tuple exposed, `dropped` and `stale_discarded` included.
  headroom's dashboard shows only `queued`, hiding precisely the "we silently gave up
  savings" counters.

## Observe: measure without enforcing

Byte-identity is **structural**, not a property of careful copying: the request path
never touches the pipeline, and also skips `expand.Inject` — injecting a tool
declaration would be a modification.

Observe results live in physically separate accumulators serialized under
`potential_*` / `projected_*`, sharing no key with any enforced metric. In observe
mode every enforced aggregate is zero by construction, which is the machine-readable
form of "context-guru did not modify requests". A mislabelled hypothetical silently
inflates the headline savings claim, so this is a correctness boundary and a test
asserts no enforced aggregate can reach an observe result.

headroom has no observe/shadow/dry-run mode at all — its `token` and `cache` modes
are both enforcing — so this is a genuine differentiator rather than a port.

## Two bugs the benchmark surfaced

Worth calling out, because both were found by running it rather than reading it:

1. **Deferred runs were double-counted.** They emitted reports stamped `async`, so
   their savings entered the enforced rollups even though nothing was forwarded — and
   entered again when a later turn replayed the decision on-path. `Report`/`RunReport`
   now carry `Deferred` and the aggregator drops those; off-path work shows up as
   `async_deferred_runs` / `async_deferred_ms_total` / the queue tuple, and savings are
   credited only where they were realized.
2. **`cacheinject` on a deferred run corrupted turn state.** Its per-message
   divergence digests are *turn* state, and a deferred job commits some turns after the
   one it was built from, so committing its digests replayed turn N's over turn N+2's
   and made the next turn compute the wrong divergence point. It now skips deferred
   runs entirely (its breakpoints were pointless there anyway — the body is discarded).

## `/stats`

Additive only; every pre-existing field keeps its name and shape, because
`deploy/harbor/*.py` parses this payload (a test pins the full pre-existing key set).
New: `mode`, `sync_enforced`, `async_enforced`, `async_queue` (the full tuple),
`async_deferred_runs`, `async_deferred_ms_total`, `async_realized_saved_tokens`, and
the observe hypotheticals.

## Overlap with #25

`prevLen` was read then written back in a `defer`, so two concurrent turns of one
session raced on it. `Tracker.Turn` now reads the previous length and records the new
one in a single locked call. That is the minimal form of the same fix #25 owns; if
#25 lands first, this reduces to using its state instead. **Merge-order note:** I did
not touch #25's files. The legacy store-backed path is left intact for library
callers and `/compact`, which supply no tracker.

## Tests

`go test -race` throughout (the whole suite, not just the new packages):

- `sync` byte-identical to the legacy entry point, and the unset mode identical to
  explicit `sync` (golden);
- observe leaves the forwarded body byte-identical while still recording;
- a stale-generation result is discarded end to end, and concurrent commits at one
  generation apply exactly once;
- concurrent turns of one session corrupt neither `prevLen` nor the generation —
  through the tracker and through the real handler;
- enqueue dedup is atomic against 32 concurrent enqueues of one key;
- a full queue drops, counts, and never blocks;
- no goroutine leak on `Pool.Stop` or `Handler.Close`, both idempotent;
- with the default policy, no breakpoint at/beyond the un-compacted tail — and the
  escape hatch restores normal placement;
- observe metrics unreachable from every enforced aggregate, and deferred runs
  counted as neither enforced nor hypothetical;
- async savings actually **arrive on a later turn** (2 committed compactions
  realizing 4,948 tokens across five turns of one session).

`CGO_ENABLED=1 go build/test -tags cg_skeleton ./...` green, `gofmt -l` clean,
`go vet` clean, `mkdocs build --strict` green.

## Docs

New `docs/how-to/operating-modes.md` (when to use each, the cache arithmetic, how to
read the queue counters and the observe numbers — including what observe *cannot*
tell you: cache effects are projected not measured, no expand bounce is exercised,
and off-path measurement still spends cheap-model tokens). `docs/design.md` gains
generations, the job lifecycle, the cache policy and fail-open per mode.
`docs/reference/config.md`, `mkdocs.yml`, README.

## Review corrections (read before the benchmark comments)

An independent review found six semantic defects — the concurrency primitives held, the
*meaning* did not. Three were places where the code and these docs disagreed, which is
worse than a bug because the docs were the spec. All six are fixed here; the two benchmark
comments below predate the fixes, and two of their claims are **retracted**:

1. **The −17.8%/−45% cache-write result is no longer claimed as causal.** The numbers stand
   as measurements, but the tail protection was inert on that workload, and two further
   defects (zero breakpoints on a session's first turn; total suppression under
   `cache_mode: off`) lower cache-write for arithmetic rather than policy reasons. The arms
   were also unpaired — 49% more cache-read, 6 mean steps apart, async re-run separately
   after a port collision. Mechanism is now verified by test; the *effect* needs a 50-task
   paired re-run with `strip_caller_breakpoints: true`.
2. **`async_realized_saved_tokens` = 15,962 "the entire enforced saving" is retracted.** The
   counter fired on every async turn that saved anything, so it re-reported the inline
   saving and equality with the total was guaranteed by construction. Now gated on a
   compaction having actually landed, with a test asserting a *strict* subset. Corrected
   measurement: `realized=4125` of `saved=4948`, with one legitimate stale discard.

Also fixed: the protected span was off by one turn (guarding this turn's new messages
rather than the previous turn's tail that the pending job actually replaces); observe's real
off-path model spend is labelled rather than implied hypothetical; unproductive sessions
stop buying cheap-model calls; eviction can no longer let a surviving in-flight job commit
over a recreated session; `Stop()` is bounded at 2s instead of inheriting the cheap model's
5-minute client timeout (which `main.go`'s deferred `Close()` would hang on); and
`store.Buffer` forwards #40's optional `FrozenLost` structurally so the wrapper does not
disable it.

Gates re-run after the fixes: build, `test -tags cg_skeleton`, **`-race -count=5`**,
`gofmt -l`, `go vet`, `mkdocs build --strict` — all clean.
`docs/results/operating-modes.md` carries both retractions inline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,35 @@ are in **[docs/components.md](docs/components.md)** and **[docs/results/componen
 | `smartcrush` | Offload | keeps anchor items of a long JSON array, drops the middle |
 | `summarize` | Offload (LLM) | compresses the middle of the trajectory into one summary (run alone) |
 
+## Operating modes
+
+`sync` (the default) compacts inline and the caller waits. Two other modes trade that off:
+
+| Mode | The request path | Use it when |
+|---|---|---|
+| **`sync`** *(default)* | Compacts inline; the caller waits (~450 ms/req measured on Terminal-Bench). | You want the full saving from turn one. |
+| **`async`** | Forwards immediately, replaying decisions an earlier turn computed; the expensive compaction runs off-path and benefits later turns. | Latency on the request path matters more than saving on turn one. |
+| **`observe`** | Forwards the request **untouched, byte for byte**, and reports what compaction *would* have saved. | You want to evaluate context-guru on your own traffic without enforcing it. |
+
+```yaml
+mode: async
+async:
+  cache_uncompacted_tail: false   # safe default: protect cache-write economics
+```
+
+Two things worth knowing before reaching for `async`: a cache-write costs **11.5x** a
+cache-read, so by default context-guru refuses to place a cache breakpoint on a tail a
+pending compaction is going to replace — caching it and then replacing it is what
+tripled headroom's cache-write on Terminal-Bench and would make `async` strictly worse
+than `sync`. And observe-mode numbers are reported under their own `potential_*` /
+`projected_*` keys that share no name with an enforced metric, so a hypothetical can
+never be read as a realized saving.
+
+`observe` is a genuine differentiator, not a port: headroom has no observe/shadow/
+dry-run mode at all — its `token` and `cache` modes are both enforcing.
+
+Details in [docs/how-to/operating-modes.md](docs/how-to/operating-modes.md).
+
 ## Integrate
 
 | Option | What | Where |
@@ -182,7 +211,8 @@ Details in [docs/integrations.md](docs/integrations.md).
 
 ## Docs
 
-- [docs/design.md](docs/design.md) — architecture: component model, fail-open pipeline, store, session, expand loop, metrics.
+- [docs/design.md](docs/design.md) — architecture: component model, fail-open pipeline, store, session, expand loop, metrics, operating modes.
+- [docs/how-to/operating-modes.md](docs/how-to/operating-modes.md) — sync vs async vs observe: when to use each, async's cache trade-off, how to read observe numbers.
 - [docs/components.md](docs/components.md) — every registered component: how it works, live before→after, lossiness, config, best use.
 - [docs/integrations.md](docs/integrations.md) — proxy gateway vs AuthBridge plugin, with request paths.
 - [docs/setup.md](docs/setup.md) — setup + a concrete SWE-bench run through the eval-containers gateway.
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -112,7 +112,19 @@ func BodyWithModelWindow(ctx context.Context, pipe *components.Pipeline, st stor
 // cache-awareness when the backend is a prompt-caching provider or the request
 // already carries cache_control breakpoints; "on" forces it; "off" restores the
 // legacy compact-everything behavior (correct for confirmed non-caching backends).
-func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, provider bschemas.ModelProvider, body []byte, explicitSession string, bypass bool, models components.ModelSpec, window int, cacheMode string) (result []byte, changedBody bool) {
+func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, provider bschemas.ModelProvider, body []byte, explicitSession string, bypass bool, models components.ModelSpec, window int, cacheMode string) ([]byte, bool) {
+	r := BodyOpts(ctx, pipe, st, Opts{
+		Provider: provider, Body: body, Session: explicitSession, Bypass: bypass,
+		Models: models, Window: window, CacheMode: cacheMode,
+	})
+	return r.Body, r.Changed
+}
+
+// BodyOpts is the full entry point: everything BodyFull takes plus the operating mode
+// (#31) and the per-session generation snapshot async mode needs. Hosts that support
+// modes call this; BodyFull is the positional shim every other caller keeps using.
+func BodyOpts(ctx context.Context, pipe *components.Pipeline, st store.Store, o Opts) (res Result) {
+	body, provider, bypass := o.Body, o.Provider, o.Bypass
 	// Top-level fail-open backstop: the per-component recover in pipeline.runOne only
 	// covers component code. A panic anywhere else on the rewrite path (normalize, the
 	// sjson splice, rebuildCountChanged, a marshal) must NOT 500 the client — forward
@@ -120,13 +132,26 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	// the whole entry point, not just inside components.
 	defer func() {
 		if r := recover(); r != nil {
-			slog.Error("context-guru: recovered from panic in BodyFull; forwarding original request", "panic", r)
-			result, changedBody = body, false
+			slog.Error("context-guru: recovered from panic in BodyOpts; forwarding original request", "panic", r)
+			res = Result{Body: body}
 		}
 	}()
+	mode := o.Mode
+	if mode == "" {
+		mode = components.ModeSync
+	}
+	models := o.Models
+	// Async, on the REQUEST path: replay only decisions that are already computed. The
+	// expensive part of a compaction is the LLM call, which is the entire reason async
+	// exists, so the inline pass gets no model clients and every NeedsModel component
+	// degrades to its deterministic path or no-ops (that degradation is already a
+	// documented contract). The off-path job (Deferred) gets the clients.
+	if mode == components.ModeAsync && !o.Deferred {
+		models = components.ModelSpec{}
+	}
 	msgsRaw := gjson.GetBytes(body, "messages")
 	if !msgsRaw.Exists() || !msgsRaw.IsArray() {
-		return body, false
+		return Result{Body: body}
 	}
 
 	// Volatile-tail split, before anything else touches the body. This is a
@@ -142,7 +167,7 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 
 	norm, slots := normalize(provider, msgsRaw.Array())
 	if len(norm) == 0 {
-		return body, systemSplit // keep the split even with nothing to compact
+		return Result{Body: body, Changed: systemSplit} // keep the split even with nothing to compact
 	}
 
 	if debugTraffic {
@@ -150,27 +175,62 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	}
 	chat := &bschemas.BifrostChatRequest{Provider: provider, Input: norm}
 	sys, firstUser := systemAndFirstUser(norm)
-	sessionID := session.Resolve(explicitSession, sys, firstUser)
-	cacheAware := resolveCacheAware(cacheMode, provider, body)
+	sessionID := session.Resolve(o.Session, sys, firstUser)
+	cacheAware := resolveCacheAware(o.CacheMode, provider, body)
 	maxCachedIdx := -1
 	if cacheAware && !bypass {
 		// Messages present on the previous turn of this session are already committed
 		// to the provider cache; only the new tail is being cache-written this turn.
 		// Restrict supersession/age offloaders to that tail so they never mutate the
 		// cached prefix. Growth-based (dialect-agnostic; needs no cache_control mapping).
-		maxCachedIdx = prevLen(st, sessionID) - 1
-		defer putLen(st, sessionID, len(norm))
+		//
+		// The boundary comes from the Tracker when the host supplies one: it reads the
+		// previous length and records this turn's in ONE locked call, which is what
+		// removes the concurrent-turn race the old read-then-deferred-write had
+		// (#31/#25). Without a tracker (library callers, /compact) the legacy store path
+		// stands — same numbers, same race, no behavior change for them.
+		switch {
+		case o.PrevLen != nil:
+			maxCachedIdx = *o.PrevLen - 1
+		case o.Tracker != nil:
+			pl, gen := o.Tracker.Turn(sessionID, len(norm))
+			maxCachedIdx = pl - 1
+			res.PrevLen = pl
+			res.Generation = gen
+		default:
+			maxCachedIdx = prevLen(st, sessionID) - 1
+			defer putLen(st, sessionID, len(norm))
+		}
+	} else if o.Tracker != nil {
+		res.Generation = o.Tracker.Gen(sessionID)
+	}
+	// Async cache policy: while a compaction for this session is queued but not landed,
+	// the un-compacted tail is about to be REPLACED, so no breakpoint may be committed
+	// at or beyond it (see components.Ctx.NoCacheAtOrAfter). CacheUncompactedTail=true
+	// is the escape hatch for a confirmed non-caching backend, where the protection buys
+	// nothing.
+	tailPending, noCacheAt := false, 0
+	if mode == components.ModeAsync && !o.Deferred && !bypass && !o.CacheUncompactedTail {
+		tailPending = true
+		if noCacheAt = maxCachedIdx + 1; noCacheAt < 0 {
+			noCacheAt = 0
+		}
 	}
 	c := &components.Ctx{
-		Ctx:          ctx,
-		Session:      sessionID,
-		Store:        st,
-		Model:        models,
-		Bypass:       bypass,
-		CtxWindow:    window,
-		CacheAware:   cacheAware,
-		MaxCachedIdx: maxCachedIdx,
+		Ctx:              ctx,
+		Session:          sessionID,
+		Store:            st,
+		Model:            models,
+		Bypass:           bypass,
+		CtxWindow:        o.Window,
+		CacheAware:       cacheAware,
+		MaxCachedIdx:     maxCachedIdx,
+		Mode:             mode,
+		Deferred:         o.Deferred,
+		TailCachePending: tailPending,
+		NoCacheAtOrAfter: noCacheAt,
 	}
+	res.Session = sessionID
 
 	// Canonical form of each normalized message BEFORE the pipeline, so a
 	// count-changing component (summarize) can be mapped back to the body.
@@ -179,7 +239,7 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 		normPre[i], _ = json.Marshal(norm[i])
 	}
 
-	pipe.Run(chat, c)
+	res.Run = pipe.Run(chat, c)
 
 	// A component changed the message count (summarize restructures the transcript
 	// to [msg0, <summary>, last-K]). Rebuild the messages array preserving each
@@ -188,9 +248,11 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	if len(chat.Input) != len(norm) {
 		nb, ok := rebuildCountChanged(body, msgsRaw.Array(), normPre, slots, chat.Input)
 		if !ok && systemSplit {
-			return body, true // keep the split even when the rebuild declined
+			res.Body, res.Changed = body, true // keep the split even when the rebuild declined
+			return res
 		}
-		return nb, ok || systemSplit
+		res.Body, res.Changed = nb, ok || systemSplit
+		return res
 	}
 
 	out := body
@@ -208,14 +270,16 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 			}
 			var err error
 			if out, err = sjson.SetBytes(out, s.path, newText); err != nil {
-				return body, false
+				res.Body = body
+				return res
 			}
 			changed = true
 			changes = append(changes, mkChange(s.path, s.preText, newText))
 		default: // wholeMessage
 			post, err := json.Marshal(chat.Input[i])
 			if err != nil {
-				return body, false
+				res.Body = body
+				return res
 			}
 			if bytes.Equal(post, s.pre) {
 				continue // unmodified — keep the original bytes verbatim (I1)
@@ -227,7 +291,8 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 				continue
 			}
 			if out, err = sjson.SetRawBytes(out, s.path, post); err != nil {
-				return body, false
+				res.Body = body
+				return res
 			}
 			changed = true
 			var pm bschemas.ChatMessage
@@ -238,7 +303,8 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	if changed && dumpPath != "" {
 		dumpChanges(c.Session, changes)
 	}
-	return out, changed
+	res.Body, res.Changed = out, changed
+	return res
 }
 
 // resolveCacheAware decides whether cache-aware compaction is active for this
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -177,6 +177,14 @@ func BodyOpts(ctx context.Context, pipe *components.Pipeline, st store.Store, o 
 	sys, firstUser := systemAndFirstUser(norm)
 	sessionID := session.Resolve(o.Session, sys, firstUser)
 	cacheAware := resolveCacheAware(o.CacheMode, provider, body)
+	// Turn accounting is independent of cache mode: the generation counts TURNS, and a
+	// turn happens whether or not the backend caches. Deriving it inside the cache-aware
+	// branch left every generation at 0 with cache_mode: off, which both disabled the
+	// stale guard and collided with 0's use as "nothing pending".
+	if o.Tracker != nil && !o.Deferred {
+		pl, gen := o.Tracker.Turn(sessionID, len(norm))
+		res.PrevLen, res.Generation = pl, gen
+	}
 	maxCachedIdx := -1
 	if cacheAware && !bypass {
 		// Messages present on the previous turn of this session are already committed
@@ -193,42 +201,52 @@ func BodyOpts(ctx context.Context, pipe *components.Pipeline, st store.Store, o 
 		case o.PrevLen != nil:
 			maxCachedIdx = *o.PrevLen - 1
 		case o.Tracker != nil:
-			pl, gen := o.Tracker.Turn(sessionID, len(norm))
-			maxCachedIdx = pl - 1
-			res.PrevLen = pl
-			res.Generation = gen
+			maxCachedIdx = res.PrevLen - 1 // recorded above, in one locked call
 		default:
 			maxCachedIdx = prevLen(st, sessionID) - 1
 			defer putLen(st, sessionID, len(norm))
 		}
-	} else if o.Tracker != nil {
-		res.Generation = o.Tracker.Gen(sessionID)
 	}
 	// Async cache policy: while a compaction for this session is queued but not landed,
 	// the un-compacted tail is about to be REPLACED, so no breakpoint may be committed
 	// at or beyond it (see components.Ctx.NoCacheAtOrAfter). CacheUncompactedTail=true
 	// is the escape hatch for a confirmed non-caching backend, where the protection buys
 	// nothing.
+	//
+	// Three conditions beyond "async", each one a bug found in review:
+	//
+	//   - cacheAware. With cache_mode: off there is no cached prefix to protect and no
+	//     boundary to protect it at, so blocking breakpoints would suppress caching
+	//     forever for nothing (the two knobs interacted backwards).
+	//   - a boundary that exists. On a session's FIRST turn prevLen is 0, so the whole
+	//     request is "tail" and blocking it wrote zero breakpoints — on precisely the
+	//     turn whose job is to write the prefix. There is also nothing to protect yet:
+	//     no compaction is pending, because no earlier turn enqueued one.
+	//   - the tail a pending job will actually replace. The job enqueued by the PREVIOUS
+	//     turn targets that turn's tail, which by now sits at or below the boundary.
+	//     Blocking from the boundary up protected this turn's new messages, which no
+	//     pending job is going to touch — off by one turn, and it protected the wrong
+	//     span. The doomed span starts where the previous turn's own tail started.
 	tailPending, noCacheAt := false, 0
-	if mode == components.ModeAsync && !o.Deferred && !bypass && !o.CacheUncompactedTail {
+	if mode == components.ModeAsync && !o.Deferred && !bypass && !o.CacheUncompactedTail &&
+		cacheAware && o.PendingFrom > 0 {
 		tailPending = true
-		if noCacheAt = maxCachedIdx + 1; noCacheAt < 0 {
-			noCacheAt = 0
-		}
+		noCacheAt = o.PendingFrom
 	}
 	c := &components.Ctx{
-		Ctx:              ctx,
-		Session:          sessionID,
-		Store:            st,
-		Model:            models,
-		Bypass:           bypass,
-		CtxWindow:        o.Window,
-		CacheAware:       cacheAware,
-		MaxCachedIdx:     maxCachedIdx,
-		Mode:             mode,
-		Deferred:         o.Deferred,
-		TailCachePending: tailPending,
-		NoCacheAtOrAfter: noCacheAt,
+		Ctx:                    ctx,
+		Session:                sessionID,
+		Store:                  st,
+		Model:                  models,
+		Bypass:                 bypass,
+		CtxWindow:              o.Window,
+		CacheAware:             cacheAware,
+		MaxCachedIdx:           maxCachedIdx,
+		Mode:                   mode,
+		Deferred:               o.Deferred,
+		TailCachePending:       tailPending,
+		NoCacheAtOrAfter:       noCacheAt,
+		StripCallerBreakpoints: o.StripCallerBreakpoints,
 	}
 	res.Session = sessionID
 
@@ -240,6 +258,7 @@ func BodyOpts(ctx context.Context, pipe *components.Pipeline, st store.Store, o 
 	}
 
 	res.Run = pipe.Run(chat, c)
+	res.TailUnprotected = c.TailUnprotected()
 
 	// A component changed the message count (summarize restructures the transcript
 	// to [msg0, <summary>, last-K]). Rebuild the messages array preserving each
--- a/apply/modes_test.go
+++ b/apply/modes_test.go
@@ -121,10 +121,11 @@ func TestStaleAsyncResultIsDiscardedEndToEnd(t *testing.T) {
 		Mode: components.ModeAsync, Tracker: tr,
 	})
 
-	// A newer turn's compaction lands first, advancing the generation.
-	if !tr.CommitIfCurrent("s", inline.Generation, func() {}) {
-		t.Fatal("could not advance the generation")
-	}
+	// A newer TURN ships, superseding the job's snapshot. This is the realistic path and
+	// the one that used to be broken: the generation advanced only on commit, so a job
+	// from turn 1 read its own generation as current no matter how many turns had
+	// shipped, and committed against a transcript long since replaced.
+	tr.Turn("s", 99)
 
 	// Now the older job finishes.
 	buf := store.NewBuffer(base)
--- a/apply/modes_test.go
+++ b/apply/modes_test.go
@@ -1,0 +1,211 @@
+package apply_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"strings"
+	"sync"
+	"testing"
+
+	bschemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/rossoctl/context-guru/apply"
+	"github.com/rossoctl/context-guru/components"
+	_ "github.com/rossoctl/context-guru/components/all"
+	"github.com/rossoctl/context-guru/config"
+	"github.com/rossoctl/context-guru/modes"
+	"github.com/rossoctl/context-guru/store"
+)
+
+func modePipe(t *testing.T, yaml string) *components.Pipeline {
+	t.Helper()
+	cfg, err := config.LoadBytes([]byte(yaml))
+	if err != nil {
+		t.Fatal(err)
+	}
+	p, err := cfg.Build(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return p
+}
+
+func dupJSON(t *testing.T) []byte {
+	t.Helper()
+	dump := strings.Repeat("a verbose repeated tool output line\n", 60)
+	b, err := json.Marshal(map[string]any{
+		"model": "gpt-x",
+		"messages": []map[string]any{
+			{"role": "user", "content": "go"},
+			{"role": "tool", "tool_call_id": "a", "content": dump},
+			{"role": "tool", "tool_call_id": "b", "content": dump},
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return b
+}
+
+// TestModeDefaultsToSync: an Opts with no Mode must produce exactly what the explicit
+// sync mode does, which is what the positional BodyFull has always produced.
+func TestModeDefaultsToSync(t *testing.T) {
+	pipe := modePipe(t, "pipeline: [dedup]\n")
+	body := dupJSON(t)
+
+	legacy, changed := apply.BodyFull(context.Background(), pipe, store.NewMemory(store.Options{}),
+		bschemas.OpenAI, body, "s", false, components.ModelSpec{}, 0, "auto")
+	if !changed {
+		t.Fatal("the legacy entry point compacted nothing; the comparison is vacuous")
+	}
+	res := apply.BodyOpts(context.Background(), modePipe(t, "pipeline: [dedup]\n"),
+		store.NewMemory(store.Options{}),
+		apply.Opts{Provider: bschemas.OpenAI, Body: body, Session: "s", CacheMode: "auto"})
+	if !bytes.Equal(legacy, res.Body) {
+		t.Fatalf("BodyOpts default differs from BodyFull\n legacy: %s\n opts:   %s", legacy, res.Body)
+	}
+}
+
+// TestAsyncInlinePassMakesNoModelCall: the whole point of async is that the request
+// path does not wait on an LLM. The inline pass must therefore be handed no model,
+// while the deferred pass is.
+func TestAsyncInlinePassMakesNoModelCall(t *testing.T) {
+	var mu sync.Mutex
+	calls := 0
+	m := fakeModel{fn: func() string {
+		mu.Lock()
+		calls++
+		mu.Unlock()
+		return "the agent read a file twice"
+	}}
+	// keep_last/min_tokens/start_from_message lowered so the tiny fixture is eligible;
+	// the point of the test is WHICH pass gets a model client, not the gating.
+	yaml := "pipeline: [summarize]\ncomponents:\n  summarize:\n    keep_last: 1\n    min_tokens: 10\n    start_from_message: 1\n    model:\n      source: config\n"
+	body := dupJSON(t)
+
+	apply.BodyOpts(context.Background(), modePipe(t, yaml), store.NewMemory(store.Options{}), apply.Opts{
+		Provider: bschemas.OpenAI, Body: body, Session: "s",
+		Models: components.ModelSpec{Static: m}, Mode: components.ModeAsync,
+	})
+	mu.Lock()
+	inline := calls
+	mu.Unlock()
+	if inline != 0 {
+		t.Fatalf("the async inline pass made %d model call(s) — the latency it exists to remove", inline)
+	}
+
+	apply.BodyOpts(context.Background(), modePipe(t, yaml), store.NewMemory(store.Options{}), apply.Opts{
+		Provider: bschemas.OpenAI, Body: body, Session: "s",
+		Models: components.ModelSpec{Static: m}, Mode: components.ModeAsync, Deferred: true,
+	})
+	mu.Lock()
+	deferred := calls
+	mu.Unlock()
+	if deferred == 0 {
+		t.Fatal("the deferred pass made no model call either — no compaction would ever be computed")
+	}
+}
+
+// TestStaleAsyncResultIsDiscardedEndToEnd wires the real pieces the proxy wires: a
+// deferred run writes into a Buffer, the session advances underneath it, and the
+// commit is refused — so not one byte of the stale result reaches the live store.
+func TestStaleAsyncResultIsDiscardedEndToEnd(t *testing.T) {
+	base := store.NewMemory(store.Options{})
+	tr := modes.NewTracker(0)
+	pipe := modePipe(t, "pipeline: [dedup]\n")
+	body := dupJSON(t)
+
+	// Turn 1 records the generation the deferred job will be built from.
+	inline := apply.BodyOpts(context.Background(), pipe, base, apply.Opts{
+		Provider: bschemas.OpenAI, Body: body, Session: "s", CacheMode: "on",
+		Mode: components.ModeAsync, Tracker: tr,
+	})
+
+	// A newer turn's compaction lands first, advancing the generation.
+	if !tr.CommitIfCurrent("s", inline.Generation, func() {}) {
+		t.Fatal("could not advance the generation")
+	}
+
+	// Now the older job finishes.
+	buf := store.NewBuffer(base)
+	prev := inline.PrevLen
+	res := apply.BodyOpts(context.Background(), pipe, buf, apply.Opts{
+		Provider: bschemas.OpenAI, Body: body, Session: "s", CacheMode: "on",
+		Mode: components.ModeAsync, Deferred: true, PrevLen: &prev,
+	})
+	if !res.Changed || buf.Writes() == 0 {
+		t.Fatal("the deferred run produced nothing; the discard test proves nothing")
+	}
+	if tr.CommitIfCurrent("s", inline.Generation, buf.Commit) {
+		t.Fatal("a STALE async result was applied")
+	}
+	// The buffer still holds every write, which IS the proof that none reached the live
+	// store: Commit is the only path there, and it never ran.
+	if buf.Writes() == 0 {
+		t.Fatal("the buffer drained even though the commit was refused")
+	}
+	// And the generation really had moved on, so the discard was not vacuous.
+	if tr.Gen("s") == inline.Generation {
+		t.Fatal("the generation never advanced, so nothing was ever stale")
+	}
+}
+
+// TestBufferIsolatesUntilCommit: the buffer is what makes "discard a stale result"
+// possible at all — without it a deferred run's writes land as it goes and cannot be
+// taken back.
+func TestBufferIsolatesUntilCommit(t *testing.T) {
+	base := store.NewMemory(store.Options{})
+	base.Put("pre", []byte("existing"))
+	buf := store.NewBuffer(base)
+
+	buf.Put("k", []byte("v"))
+	buf.MarkSticky("s", "id")
+	if _, ok := base.Get("k"); ok {
+		t.Fatal("a buffered write reached the base store before Commit")
+	}
+	if v, ok := buf.Get("k"); !ok || string(v) != "v" {
+		t.Fatal("the buffer cannot read its own write")
+	}
+	if v, ok := buf.Get("pre"); !ok || string(v) != "existing" {
+		t.Fatal("the buffer does not fall through to the base store")
+	}
+	if _, ok := base.Sticky("s")["id"]; ok {
+		t.Fatal("a buffered sticky mark reached the base store before Commit")
+	}
+
+	buf.Commit()
+	if v, ok := base.Get("k"); !ok || string(v) != "v" {
+		t.Fatal("Commit did not flush")
+	}
+	if _, ok := base.Sticky("s")["id"]; !ok {
+		t.Fatal("Commit did not flush sticky marks")
+	}
+	if buf.Writes() != 0 {
+		t.Fatal("Commit did not drain the buffer")
+	}
+
+	// Discard: never committed, never visible.
+	d := store.NewBuffer(base)
+	d.Put("gone", []byte("x"))
+	if _, ok := base.Get("gone"); ok {
+		t.Fatal("an uncommitted write is visible")
+	}
+}
+
+// TestSessionOfMatchesApply: the async worker keys jobs by the session id apply
+// resolves, so the two must agree — including on the content-hash fallback.
+func TestSessionOfMatchesApply(t *testing.T) {
+	pipe := modePipe(t, "pipeline: [dedup]\n")
+	body := dupJSON(t)
+	res := apply.BodyOpts(context.Background(), pipe, store.NewMemory(store.Options{}),
+		apply.Opts{Provider: bschemas.OpenAI, Body: body}) // no explicit session
+	if got := apply.SessionOf(bschemas.OpenAI, body, ""); got != res.Session {
+		t.Fatalf("SessionOf resolved %q, apply used %q", got, res.Session)
+	}
+}
+
+// --- helpers ----------------------------------------------------------------
+
+type fakeModel struct{ fn func() string }
+
+func (f fakeModel) Complete(context.Context, string) (string, error) { return f.fn(), nil }
--- a/apply/opts.go
+++ b/apply/opts.go
@@ -1,0 +1,94 @@
+package apply
+
+import (
+	bschemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/rossoctl/context-guru/components"
+	"github.com/rossoctl/context-guru/modes"
+	"github.com/rossoctl/context-guru/session"
+	"github.com/tidwall/gjson"
+)
+
+// Opts is BodyOpts' input: everything the positional BodyFull takes, plus the
+// operating mode (#31) and the per-session generation state async mode needs.
+//
+// A struct rather than a 13th positional argument: the parameter list was already at
+// the limit of readability, and modes add three fields that only one host sets.
+type Opts struct {
+	Provider bschemas.ModelProvider
+	Body     []byte
+	// Session is the host-supplied session id ("" => content hash).
+	Session string
+	Bypass  bool
+	Models  components.ModelSpec
+	// Window is the model's resolved context window (max input tokens; 0 = unknown).
+	Window int
+	// CacheMode is "auto" (default) | "on" | "off" — see resolveCacheAware.
+	CacheMode string
+
+	// Mode is the operating mode. Empty means components.ModeSync, so a caller that
+	// does not know about modes gets exactly today's behavior.
+	Mode components.Mode
+	// Deferred marks the OFF-PATH async run: nothing it produces is forwarded, it
+	// exists to populate the frozen state later turns replay. Only the async worker
+	// sets it, and it is what re-enables the LLM components the inline async pass
+	// deliberately withholds.
+	Deferred bool
+	// CacheUncompactedTail disables async's tail cache protection. In async mode the
+	// tail beyond the cached prefix is by construction content a not-yet-landed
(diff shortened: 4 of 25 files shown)


OsherElhadad:
## Benchmark results — all three modes

Commit `a0c7253` (branch head at time of run), `codesmart` pipeline, `claude-code` on
`aws/claude-sonnet-5` through the IBM gateway. Cache-aware billed cost recomputed from
each trial's token tiers.

**Scale up front:** 2 SWE-bench tasks per mode at n=1, plus one real Claude Code session
per mode. Enough to validate the mechanism and answer the latency and cache questions;
**not** enough for a cost or solve-rate claim. Treat the billed-cost column as tracking
trajectory length, not mode.

### SWE-bench Verified — 2 tasks, n=1

| | `sync` | `async` | `observe` |
|---|---|---|---|
| solved | 2/2 | 2/2 | 2/2 |
| mean steps | 15.5 | 21.5 | 22.5 |
| **added latency / req** | **1,599.4 ms** | **25.3 ms** | **0.062 ms** |
| content savings (enforced) | 0.82% | 4.17% | — (0 by construction) |
| projected savings | — | — | 6.40% |
| cache-read | 1,464,729 | 2,186,100 | 2,300,699 |
| **cache-write** | **52,287** | **42,980** | 127,589 |
| cache-write per 1M cache-read | 35,697 | **19,661** | — (not enforcing) |
| cache-hit rate | 96.55% | **98.07%** | 94.74% |
| fresh input | 54 | 78 | 86 |
| output | 10,249 | 10,711 | 11,522 |
| billed cost | $0.5263 | $0.6519 | $0.8945 |
| context-guru's own LLM cost | $0.0122 (1) | $0.0435 (4) | $0.0779 (7) |
| off-path compaction time | — | 80.8 s | 75.0 s |
| deferred compactions committed | — | 1 | 0 (never commits) |
| `async_realized_saved_tokens` | — | 15,962 | — |
| queue `{dropped, stale_discarded}` | — | `{0, 0}` | `{0, 0}` |

### The four questions

**1. Does async reduce added latency without increasing cache-write?**

Latency: **1,599.4 ms → 25.3 ms, 63x**. Visible per component — `extract_llm` costs
15,014 ms on sync's request path and 71.3 ms *cumulative across 42 requests* on async's,
with `acted=0` inline. The model call is genuinely off the hot path; 80.8 s of
compaction ran there instead.

Cache-write: **it went down.** 52,287 → 42,980 absolute, on ~49% *more* cache-read
traffic — normalised, **19,661 cache-write tokens per 1M cache-read against sync's
35,697 (−45%)**, hit rate 96.55% → 98.07%.

That is the failure mode the issue warned about *not* occurring. A policy that was
caching the un-compacted tail and then rewriting it could not produce this arm. I claim
only the direction, not the magnitude — n=1, two differently-shaped trajectories.

**2. Does async reach the same steady-state savings as sync, just later?**

It reached *more* here (4.17% vs 0.82%), but that is trajectory noise, not a finding.
The load-bearing number is `async_realized_saved_tokens` = 15,962 = **the entire
enforced saving**, from 1 committed deferred compaction. Every token async saved was
saved by a later turn replaying an earlier turn's off-path work. The deferral works end
to end on real traffic.

**3. Does observe add measurable latency to the enforced path?**

**No — 0.062 ms/req** against sync's 1,599.4 ms. Structural, not tuned: the request path
never runs the pipeline. Confirmed independently in a live Claude Code session (0.209 ms
vs sync's 28.964 ms).

Observe is not free in other respects — it moved 75.0 s of compaction off-path and spent
$0.0779 of cheap-model tokens measuring. It costs money and CPU, not request latency.

**4. Do observe's projections match sync's actuals?**

Answering this honestly found **two real bugs**, and was the most valuable thing the
benchmark did. Both are fixed in this PR with tests.

- First comparison: **9.53% projected vs 0.82% enforced**, an 11x overstatement. The
  observe job ran without the session tracker, so its cached-prefix boundary was
  unknown, the tail gate never fired, and 50 `extract_llm` candidates passed where sync
  allowed 5. A projection that ignores cache-awareness projects what a *cache-blind*
  proxy would do.
- Fixing that exposed the opposite error: observe then **under**-projected ~3x, having
  run against a discarded buffer, losing the frozen decisions offloaders replay every
  turn — where most of the sustained saving lives.

After both fixes, on **identical traffic** through the real handler, projection and
actual agree **exactly: 10,020 tokens / 23.06% each**. A test pins it and fails at ratio
0.33 if the shadow store is removed.

On the benchmark arms the residual gap is **6.40% projected vs 0.82% enforced**, and I
am not explaining it away:
- the arms are different trajectories (22.5 vs 15.5 mean steps) — observe saw 46
  requests / 492,652 baseline tokens, sync saw 35 / 244,319. Not the same conversations.
- observe's projection never pays a bounce (nothing is offloaded, so `wasted_tokens` is
  structurally 0). It is an **upper bound** on content savings, and documented as one.
- 2 tasks at n=1 cannot separate a real bias from trajectory noise.

The controlled same-traffic test is the strong evidence; the arms are consistent with it
but too small to confirm it independently. A 50-task paired run is the honest next step.

### Real Claude Code sessions (one per mode)

| | `sync` | `async` | `observe` |
|---|---|---|---|
| requests (enforced) | 4 | 5 | **0** |
| `sync_enforced` / `async_enforced` | 4 / 0 | 0 / 5 | 0 / 0 |
| added latency / req | 28.964 ms | 17.797 ms | **0.209 ms** |
| baseline tokens | 6,025 | 8,124 | 6,025 *(as `actual_baseline_tokens`)* |
| queue `{dropped, stale_discarded}` | — | `{0, 0}` | `{0, 0}` |
| task answered correctly | yes | yes | yes |

Observe's `actual_baseline_tokens` = 6,025 is *exactly* sync's `tokens_before` = 6,025 on
the same prompt — the hypothetical namespace accounts for identical traffic identically,
measured independently.

### Namespace separation, verified in production

From the observe arm's live `/stats`: enforced side `requests: 0`, `saved_tokens: 0`,
`sync_enforced: 0`, `async_enforced: 0`, `components: {}` — all zero, all empty;
hypothetical side `observe_hypothetical_requests: 46`,
`actual_baseline_tokens: 492652`, `projected_optimized_tokens: 461112`,
`potential_saved_tokens: 31540`, `potential_components: {…}` — fully populated. No
aggregate over the enforced rollups can reach a hypothetical.

### Terminal-Bench 2.0

Ran `sync` (26.9 ms/req added, 1.02% savings, 60 pipeline runs) and `async` (19.7 ms/req
added, 77 requests) on 2 cache-sensitive tasks. The gap is small there because
`extract_llm` made **zero** calls on these tasks — with no model call in the pipeline
there is nothing expensive to defer, which is itself the useful negative result: async's
benefit is proportional to how much LLM work the pipeline does. Neither arm solved these
two (hard) tasks, and one hit an environment-build exception, so I am not quoting
reward numbers from it.

### Bugs the benchmark found that the tests did not

All four passed the unit suite while live:
1. deferred runs double-counted into the enforced rollups;
2. `cacheinject` committing stale per-turn digests from an off-path run;
3. observe overstating 11x (missing cache boundary);
4. observe understating 3x (discarded frozen state).

Each now has a test that fails without its fix.

### Not established

- The cache-write result's *magnitude* (direction is solid).
- Any per-mode cost or solve-rate claim.
- Async under concurrency pressure: `dropped` and `stale_discarded` were 0 on every
  arm, so those paths are exercised only by tests, never yet by production load.

Full write-up in `docs/results/operating-modes.md`.

OsherElhadad:
## Terminal-Bench 2.0 — the cache-write result replicates

Second benchmark, different traffic shape, and it independently reproduces the finding
that matters most for this PR.

| | `sync` | `async` |
|---|---|---|
| added latency / req | 26.9 ms | 26.8 ms |
| cache-read | 3,144,887 | 6,311,918 |
| cache-write | 68,211 | 83,222 |
| **cache-write per 1M cache-read** | **21,689** | **13,185 (−39.2%)** |
| cache-hit rate | 97.87% | 98.70% |
| pipeline runs | 60 | 110 |
| context-guru's own LLM calls | **0** | **0** |
| off-path compaction time | — | 2.5 s |
| `async_realized_saved_tokens` | — | 1,156 |
| queue `{dropped, stale_discarded}` | — | `{0, 0}` |

**Normalised cache-write fell 39.2% under async here, against −45% on SWE-bench.** Two
benchmarks, two traffic shapes, same direction. That is the strongest evidence I have
that the tail policy is doing its job — a policy that was caching the un-compacted tail
and then rewriting it could not produce either arm.

**And a useful negative result:** `extract_llm` made **zero** model calls on these tasks,
so there was nothing expensive to defer and async's added latency is identical to sync's
(26.8 vs 26.9 ms). Async buys back the compaction model call, so on traffic that never
triggers one it buys nothing — and, importantly, costs nothing either. Worth knowing
before anyone enables async expecting a universal win: the benefit is proportional to how
much LLM work your pipeline actually does.

No reward numbers from this arm. Two hard tasks, an environment-build exception in each
configuration, and trajectories that diverged sharply (30 vs 55.5 mean steps) — at this
scale the step count drives the cost column entirely, not the mode.

The observe arm was still running at hand-off; `docs/results/operating-modes.md` carries
the full write-up including everything above.

OsherElhadad:
## All nine trials complete — TB observe closes it out

| Terminal-Bench | `sync` | `async` | `observe` |
|---|---|---|---|
| **added latency / req** | 26.9 ms | 26.8 ms | **0.076 ms** |
| **cache-write per 1M cache-read** | **21,689** | **13,185 (−39.2%)** | — |
| cache-hit rate | 97.87% | 98.70% | 97.15% |
| content savings (enforced) | 1.02% | 0.11% | — (0 by construction) |
| projected savings | — | — | **0%** |
| enforced requests | 60 | 110 | **0** |
| queue `{dropped, stale_discarded}` | — | `{0, 0}` | `{0, 0}` |

**Observe projected 0%, and that is the result I'd point a reviewer at first.** It is a
negative control, and a stronger form of evidence than the SWE-bench agreement: on
traffic where sync achieved almost nothing (1.02%), observe correctly reported almost
nothing rather than inventing a headline. A mode that only ever agreed on high-savings
traffic would be much weaker evidence that its projections mean anything. It also
correctly reported the overhead sync *would* have added (9.1 ms/req — small here because
the pipeline made no model calls) while itself costing 0.076 ms.

That gives three independent lines of evidence on projection accuracy:
- **exact** agreement on identical traffic through the real handler (10,020 tokens /
  23.06% both sides, controlled test);
- **correct near-zero** agreement on Terminal-Bench (1.02% actual → 0% projected);
- consistent-but-noisy on SWE-bench (0.82% vs 6.40%), where the arms took different
  trajectories and observe's projection is a structural upper bound — documented as such
  rather than explained away.

### Final tally

- Nine trials: 2 SWE-bench + 2 Terminal-Bench tasks per mode at n=1, plus one live Claude
  Code session per mode. All nine complete; nothing left running.
- Async's cache-write fell on **both** benchmarks (−45% normalised on SWE-bench, −39.2%
  on Terminal-Bench). The headroom failure mode did not reproduce.
- `dropped` and `stale_discarded` stayed at `0` across all nine — no queue pressure was
  ever reached, so those paths remain test-only. Still the top gap.
- Four bugs found by benchmarking that the passing test suite missed, each now with a
  test that fails without its fix.

`docs/results/operating-modes.md` carries the full write-up, including a "what is not
established" section.

OsherElhadad:
## Orchestrator verification of the cache-write claim — it holds, with a caveat the PR should state

I re-derived the headline number from the row files rather than taking it on trust, and hit a discrepancy worth recording.

### What tripped me up

`/tmp/i31-runs/swe/` contains all three arms, and its async arm gives **41,136** cw/1M-read — *worse* than its sync arm's 35,697. The PR's 19,661 comes from a different directory, `swe-as3/`. So the headline pairs `sync` from one run against `async` from another.

**I checked whether the discarded arms were dropped for being unfavourable, and they were not.** Both are genuinely invalid:

```
swe/rows-...-async.json      astropy-12907  reward=None steps=None exception=True
swe-async2/rows-...async.json  both tasks   reward=0.0  steps=2    exception=True (wall 2.3s / 2.6s)
```

A nulled trial and a pair of 2-step aborts. `swe-as3` is the only clean async arm (2/2 solved, no exceptions). Discarding the others is correct — and is exactly the degenerate-trial exclusion this project had to learn the hard way on the Terminal-Bench study.

### Paired per task, which is the honest framing

| task | steps sync→async | cw/1M-read sync→async | reward |
|---|---|--:|---|
| `astropy-12907` | 12 → 10 | 58,983 → **27,207** | 1.0 → 1.0 |
| `astropy-14365` | 19 → 33 | 23,625 → **17,877** | 1.0 → 1.0 |

**Both tasks move the same direction, and — importantly — the one whose step count nearly doubled (19 → 33) still shows lower normalised cache-write.** That is the case where trajectory noise would most plausibly explain the result away, and it doesn't. Terminal-Bench replicates independently (`tb3`: 21,689 → 13,185, −39%, again with async taking *more* steps: 60 → 111).

So the direction survives per-task pairing on both benchmarks, in both cases with async on the longer trajectory. That is stronger evidence than the aggregate table conveys.

### What the PR should add

State plainly that **sync and async came from separate runs** (`swe/` and `swe-as3/`), that two async arms were discarded for `exception: true` trials with the specific reason, and give the per-task paired table above rather than only the aggregate. As written, a reader who opens `swe/` will find an async arm that contradicts the headline and no explanation. That is a documentation gap, not a data problem — but it is the kind that erodes trust in an otherwise well-evidenced result.

### On the remaining open concerns

- **`dropped`/`stale_discarded` = 0 across all nine trials** is the gap I'd rank highest. The stale-discard path is the single most important correctness invariant in this PR, and production load never exercised it. Tests cover it; that's acceptable to merge on, but it should be called out as untested-in-anger rather than listed among passing metrics.
- The **negative control on TB projection** (1.02% actual → 0% projected, because `extract_llm` made zero calls) is a genuinely good validation and is currently underplayed. A projection that correctly reads near-zero when there is nothing to project is worth more than the exact-agreement case.
- The `extract_llm`-made-zero-calls finding also cross-confirms **#34**, which independently concluded the component is net-negative on caching backends, and **#37**, which measured it burning 17.8 s across 2 calls for 0 tokens saved. Three independent lines of evidence now point the same way.

Verdict from my side: the mechanism and the direction are sound and replicated. Fix the provenance disclosure before merge.

OsherElhadad (commented):
## Independent review — verdict: **request changes**

I did not write this code. I built a scratch worktree at `feat/i31-modes` (d550bbe), ran
`go test -race ./...` (green), `-race -count=5` on `modes/apply/store/metrics` and
`-count=3` on `proxy` (green, no flakes), `go build`, `go vet`, `gofmt -l` (all clean).
The concurrency primitives themselves are sound — I could not construct a data race, a
lost update in `Tracker`, a dedup window in `Pool`, or a goroutine leak on `Stop`.

The problems are not races. They are **semantic**: the generation does not mean what the
PR says it means, the cache protection has three holes that turn it into a cache-write
*regression* on the exact backend it was written for, and two of the headline numbers are
tautologies. Each finding below has a runnable reproduction I ran against this branch.

---

### S1 — The tail cache protection is defeated by any agent that sets its own breakpoints (claude-code does)

`components/reformat/cacheinject.go:155-168`. The policy only deletes blocked positions
from the `want` set. Caller-set breakpoints are handled *after*, at line 172-178, and are
never removed:

```
c.MaxCachedIdx=21, NoCacheAtOrAfter=22, agent set its own breakpoint on msg 29
=> breakpoints = [19 21 29]   // 29 is inside the doomed tail
```

The cacheinject doc comment (v2's whole design) says "keeps every breakpoint the caller
set". claude-code sets its own tail breakpoint on every turn. So on the primary target
workload the protection places its safe anchor at 21 *in addition to* the agent's unsafe
one at 29 — the doomed tail is still cache-written, and async is now strictly worse than
sync **plus** one wasted slot. This is the failure mode #31 explicitly warns about,
arriving through the one path the tests don't cover.

Fix: when `TailCachePending`, strip `cache_control` from blocked indices before counting
`existing`. That is the only version of this policy that is actually a policy.

### S2 — Every async session's first turn writes no cache at all

`apply/apply.go:206-212`. First turn: `prevLen=0` → `maxCachedIdx=-1` →
`noCacheAt = max(0, 0) = 0` → `CacheBlocked(i)` is true for **all** i, and the
"highest safe index" fallback is `-1`, so nothing is placed. Reproduced: 40 messages,
`breakpoints=[] skipped=true`. The existing `TestWholeRequestPendingPlacesNothing`
asserts this as *correct*, treating `NoCacheAtOrAfter == 0` as "the whole request is
doomed tail" — but on turn 1 nothing is pending yet, and turn 1 is precisely the turn
that must write the prefix so turn 2 can read it at 0.1x. Suppressing it bills the whole
prefix at 1.0x on turn 2. On short sessions (the common case) that is the *entire* cache
benefit gone.

Fix: on the first turn there is no pending compaction for this session, so
`tailPending` must be false. Gate on "a job is actually in flight for this session",
not on `mode == async`.

### S3 — `cache_mode: off` + `async` suppresses every breakpoint on every turn, forever

`apply/apply.go:206-212` sets `tailPending = true` whenever mode is async and not
bypassed — including when `cacheAware` is false, where `maxCachedIdx` never leaves `-1`.
Result: `NoCacheAtOrAfter` is permanently 0 and cacheinject places nothing, permanently.
Reproduced. Ironic given `cache_uncompacted_tail: true` exists as the escape hatch for
exactly this deployment; the two knobs interact backwards.

Fix: `if mode == ModeAsync && !o.Deferred && !bypass && cacheAware && !o.CacheUncompactedTail`.

### S4 — The protection is off by one turn: it blocks the wrong region

The blocked region is `[maxCachedIdx+1, ∞)` — *this* turn's new messages. But the
pending job was built from the **previous** turn's body, so the span it will replace is
the previous turn's tail, which by turn N+1 sits *below* the boundary and is
unprotected. Reproduced: turn N boundary 22 (job targets 22..29); turn N+1 boundary 30
places its top breakpoint at 29 — inside the span the still-pending job will rewrite.

So the protection blocks a region no pending job targets, and permits the region every
pending job does target. Combined with S1/S2/S3 I do not think the claimed mechanism
holds at all, which brings me to:

### S5 — "cache-write went down 17.8%" is not evidence the policy works

`docs/results/operating-modes.md:44-58` presents −17.8% cache-write as "the result the
policy was designed for". Given S2/S3, async's cacheinject frequently places **fewer or
zero** breakpoints — which lowers cache-write for the trivial reason that less is
written, while raising *uncached* input on later turns. The doc's own numbers are
consistent with that reading: cache-read went **up 49%** in the same arm. At n=1 over 2
tasks with different trajectories (15.5 vs 21.5 steps) this is not causal, and the
paragraph "this arm cannot be reconciled with a cache policy that was rewriting the live
zone" overclaims — it also cannot be distinguished from a policy that stopped writing
cache. Please either drop the causal claim or add the one measurement that separates
them: breakpoints placed per request, sync vs async.

### S6 — The generation counts commits, not turns, so `CommitIfCurrent` never rejects a genuinely stale result

This is the invariant the PR calls "the one everything rests on". `Tracker.gen` advances
only in `CommitIfCurrent`. A turn does **not** advance it. So a job built from turn 1
still sees `gen == 0` as current after turns 2…N have been forwarded, and commits.

Reproduced through the real handler (6 turns, `mask`, cache on):
`deferred_runs(committed)=3, StaleDiscarded=2`. The 2 discards are *dedup collisions*
(two jobs at the same gen), not "the session moved on". Every commit that did happen
wrote a result built from a snapshot the session had already moved several turns past —
the exact situation the guard was written to catch — and none was flagged.

The frozen decisions are content-hash-keyed so this is not corruption today, but the
guard does not provide the property claimed in the PR body, `docs/design.md`, and
`docs/how-to/operating-modes.md:85-89`. Either advance the generation on `Turn` (and
accept more discards), or restate the invariant honestly as "at most one commit per
generation" — which is what the code actually enforces.

### S7 — `async_realized_saved_tokens` is a tautology

`proxy/modes.go:59-62` calls `RecordRealized(res.Run.Saved())` on **every** async request
whose inline pass saved anything, with no check that any deferred job ever committed.
Reproduced on the **first** turn with a purely deterministic pipeline (`dedup`, no model,
nothing to defer): `saved=396, realized=396`. So
`async_realized_saved_tokens == saved_tokens` by construction, not by measurement.

That means `docs/results/operating-modes.md:70` — "`async_realized_saved_tokens` = 15,962
= the entire enforced saving … every token async saved was saved by a later turn
replaying an earlier turn's off-path job" — is circular. The equality is what the code
computes, not a finding. To make it real, credit only savings whose frozen decision was
written by a committed deferred run (e.g. tag replayed decisions, or diff against what
the inline pass would have saved with an empty store).

### S8 — Observe mode does contaminate two pre-existing enforced `/stats` fields

`TestObserveMetricsCannotBeSummedIntoEnforcedTotals` checks `requests`, `tokens_*`,
`saved_tokens`, `components`, `sync/async_enforced`. It does not check the *other*
enforced-namespace fields, and two are non-zero in observe mode:

- `cg_added_ms_avg` — `proxy.go:438` records added latency in every mode. Reproduced:
  `requests=0, cg_added_ms_avg=0.0087`. `deploy/harbor/swebench.py:305` reads this
  straight into `s["cg_added_ms_avg"]`, so an observe arm reports a real-looking
  overhead figure for requests it never touched.
- `llm_calls` / `llm_input_tokens` / `llm_output_tokens` — `proxy.go:620` reads the
  global `cheapmodel.Usage()`, which observe's off-path runs increment. `swebench.py`
  turns these into `cg_llm_cost`. The results doc itself reports observe spending
  $0.0779 of cheap-model tokens, so this is known — but it lands in the *enforced*
  vocabulary a harness reader sums.

Neither is a hypothetical mislabelled as a saving, so the headline claim survives. But
"in observe mode every enforced aggregate is zero by construction" is not true as
written. Either move these under `potential_*` in observe mode or amend the claim in the
PR body, `metrics.go:359-366`, and the how-to.

### S9 — An unproductive session re-runs a full off-path compaction on every turn, unbounded

The generation only advances on commit and `runOne` releases the pending slot
unconditionally, so a job that finds nothing to compact is re-enqueued at the same key
every turn. Reproduced: 10 turns of one session → **10** off-path jobs, gen still 0.
Each is a real cheap-model LLM call in production.

`TestAsyncDoesNotSpinOnAnUnproductiveTurn` only asserts the queue stays bounded and 12
requests were forwarded — it never counts jobs run, so it passes while the spin happens.
This is the answer to "what if turns arrive faster than compaction completes,
indefinitely": nothing bounds the *spend*, only the queue depth. Suggest a per-(session,
generation) attempt cap, or advance the generation on a completed-but-unproductive job.

### S10 — Tracker eviction lets a stale gen-0 result commit, and fails open into mutating the cached prefix

`modes/tracker.go:64-72` evicts an arbitrary session at the cap and recreates it at
`gen=0, prevLen=0`. Reproduced with `NewTracker(2)`: session A's turn-1 job (100 msgs)
commits over a session now at 400 msgs, because both read `gen == 0`. The comment claims
a forgotten session "restarts at generation 0 — correct, just missing the pending job's
savings"; it is not correct, it is the one thing the guard exists to prevent.

Worse, `prevLen` resetting to 0 makes `MaxCachedIdx = -1`, and `Ctx.TailOnly` returns
**true for every index** — so offloaders are free to mutate the whole already-cached
prefix. That is #25's fail-open direction, reachable here through eviction alone. At 1000
sessions with random eviction this is rare but not negligible.

Fix: LRU rather than arbitrary eviction (the tracker already has the map; `Turn` is the
touch point), and make eviction leave a tombstone or start recreated sessions at a
generation no in-flight job can hold.

### S11 — `Pool.Stop()` blocks for as long as the in-flight job takes

`modes/pool.go:191-201` cancels then `wg.Wait()`s. A running job that does not poll `ctx`
— and the cheap-model HTTP client has a **5-minute** timeout — holds Stop for its full
duration. Reproduced: Stop blocked 1.5 s on a `time.Sleep` job. `main.go` does
`defer h.Close()`, so process shutdown inherits it. `TestStopLeaksNoGoroutines` waits for
the job to finish *before* calling Stop, so it never exercises this. Suggest a bounded
wait in `Stop` (the jobs are pure savings; abandoning one is already the documented
trade) and a test that stops mid-flight.

---

### Minor

- `store/buffer.go:60-70` — `Sticky()` calls `b.Base.Sticky(session)` *before* taking
  `b.mu`. Not a race (both are individually locked) but it means a concurrent `Commit`
  can be observed half-applied through this one method. Worth a note; harmless today
  because only one worker touches a given buffer.
- `store/buffer.go` — an uncommitted Buffer is dropped with its job closure, so no leak.
  Verified. But a *dropped* enqueue never constructs one, and a job cancelled mid-run
  leaves the partial writes garbage-collected. Fine — no action, recording that I checked.
- `metrics.go:228-267` `observeComp` is a near-verbatim copy of the second half of
  `Component`. Two copies of the unique-savings attribution will drift. Extract the
  shared body and pass the target map.
- `Aggregator.Run` returns early on `r.Deferred` *before* the observe branch, so a
  deferred observe run (can't happen today — observe never sets Deferred) would silently
  vanish. Order these the same way as in `Component` for consistency.
- `docs/how-to/operating-modes.md:85-89` and `docs/design.md` state the S6 invariant as
  implemented. They need to match whatever S6 resolves to.
- Docs: I could not run `mkdocs build --strict` (mkdocs is not installed on this box), so
  the green claim is taken on trust. Nav entries and the relative link from
  `reference/config.md` check out by inspection.

### What is genuinely good

- `Tracker` is correct as a mutual-exclusion primitive; `CommitIfCurrent` running the
  commit under the generation lock is the right shape and I could not break it.
- `Pool`'s claim-before-observable dedup is real — the slot is taken under `mu` before
  the channel send, and rolled back on a full queue. Atomic as advertised.
- Panic containment verified: `runOne`'s `recover` is inside the deferred func that also
  releases the slot, so a panicking job neither kills the worker nor leaks its key.
- Structural observe byte-identity (never running the pipeline, skipping `expand.Inject`)
  is the right design, and `TestObserveForwardsByteIdenticalBody` genuinely asserts it.
- `sync` byte-identity: verified, and `TestCompactEndpointIgnoresMode` covers the
  `/compact` regression risk. `adapters/bifrost` still goes through `BodyFull`, whose
  shim is a faithful translation. No security issues found — no credential, token, or
  gateway URL in the new code, tests, docs, or PR body.

### Bottom line

The async machinery is well built. What it is *doing* is not what the PR says, in three
places that matter: the cache policy has holes on its primary target workload (S1–S4),
the stale-result invariant is not the one implemented (S6), and two headline numbers are
tautological (S5, S7). S1, S2, S3, S6 and S7 should block; S8–S11 are fixable follow-ups
if scoped. Every finding above has a reproduction I ran on this branch and can hand over.

OsherElhadad:
## Correction to my own earlier comment — the cache-write result is not yet causal

I verified the per-task pairing above and concluded the direction "survives per-task pairing on both benchmarks." **That was premature, and the independent review has a better explanation than mine.**

The reviewer found (and I confirmed at `components/reformat/cacheinject.go:157-166`) that the async cache protection **only deletes from context-guru's own `want` set** — caller-set breakpoints are counted afterwards and never stripped. So on claude-code traffic, which sets its own tail breakpoint, the doomed tail is still cache-written. The protection does nothing on the primary workload.

Combined with two more findings — every async session's first turn places **zero** breakpoints (`prevLen=0` → `NoCacheAtOrAfter=0` → everything blocked), and `cache_mode: off` + async suppresses breakpoints entirely — there is a much more mundane reason for lower cache-write in the async arms: **async often wrote fewer or no breakpoints at all.** Fewer writes because fewer breakpoints is not the mechanism this PR claims. And cache-read rose 49% in the same arm, which is what you'd expect if the prefix simply wasn't being anchored.

So my "both tasks move the same direction, including the one whose step count nearly doubled" observation is still factually true, but it does **not** support the causal reading I gave it. Two arms can both show lower normalised cache-write for a reason that has nothing to do with the cache policy working. I should have checked whether the policy actually fired before treating the outcome as evidence that it did — the same discipline I applied to the degenerate-baseline and `xdedup` cases, and I didn't apply it here.

**Revised position:** keep the numbers, relabel them as not-yet-causal, and re-measure after S1–S4 are fixed. The latency result (1,599 ms → 25 ms) is unaffected and stands on its own — that one doesn't depend on the cache policy at all.

Also retracting my characterisation of `async_realized_saved_tokens = 15,962 = the entire enforced saving` as validation of end-to-end deferral. I verified `proxy/modes.go:61` fires on *any* async request that saved anything, with no commit check, so the equality is circular by construction.

This is the fourth premise in this workstream to fall to "check whether the mechanism fired before believing the outcome" — after `xdedup`, the expand-tool registration, and `prefixpin`'s early-index churn. The pattern is consistent enough to be worth stating as a rule for the final write-up: **an aggregate moving in the predicted direction is not evidence the predicted mechanism operated.**

OsherElhadad:
## Review response — all seven blocking findings fixed, plus the four follow-ups

Thanks — this was a genuinely good review. The three where the code and the docs
disagreed were the worst of it, and you were right to rank them first: the docs were the
specification, so those weren't bugs, they were false claims.

I reproduced S1, S2, S6 and S7 before touching anything. S1 gave me `[19 21 29]` at
boundary 22, exactly as reported. S2 gave `[]`. S6 committed a turn-1 job after eight
later turns. S7 I fixed from the code path alone — the missing gate is plain once you look.

### The three code/doc disagreements

**S1 — protection was inert on the primary workload.** Fixed. The loop now also covers
caller-set breakpoints inside the protected span, and since removing a directive an agent
deliberately placed is a change to someone else's request, the host chooses:
`async.strip_caller_breakpoints: true` strips them, or `cacheinject` calls
`Ctx.DeclineTailProtection()` and the proxy **does not defer that turn**
(`async_tail_unprotected_turns`). I took your second option as the default — with
claude-code on defaults async is now deliberately inert rather than falsely protective,
and the counter says so. Two tests: one per branch.

**S6 — the generation now advances per turn.** This was the deepest one. Fixing it
surfaced a second defect you didn't see: turn accounting lived *inside* the cache-aware
branch, so with `cache_mode: off` every generation stayed 0, which both disabled the guard
and collided with 0's use as "nothing pending". Moved out — a turn happens whether or not
the backend caches.

I also had to make `CommitIfCurrent` idempotent per generation: it previously
self-superseded via `gen++`, and once that moved to `Turn` sixteen concurrent jobs at one
generation could all commit. Dedup prevents that in production, but the guard should be
exact alone.

The honest consequence is now in the docs rather than buried: at agent turn rates a
compaction taking tens of seconds is usually superseded before it lands, so async
discards much of what it pays for. `stale_discarded` is the number that decides whether
async suits a workload, and I stopped calling it a tuning nit.

**S7 — tautology gone.** Gated on `Tracker.Landed(session)`. The test now asserts
`realized` is a **strict** subset of `saved_tokens`, since equality is the signature of
the old bug. Corrected measurement: `realized=4125` of `saved=4948`, one legitimate stale
discard. Retracted from the PR body and `docs/results/operating-modes.md`.

### S2/S3/S4

All three were the same root shape — the protection firing with nothing to protect — so
they're one guard now: async **and** cache-aware **and** something actually pending, with
the span taken from `Opts.PendingFrom` (the *previous* turn's tail, tracked through
`Tracker.SetPending`/`ClearPending` on every terminal job path, including "ran and
produced nothing" so it can't latch on forever).

You were right that a test encoded S2. `TestWholeRequestPendingPlacesNothing` asserted
turn-1 suppression as correct; it's now `TestFirstTurnStillWritesThePrefix` asserting the
opposite.

### S5 — claim withdrawn

Agreed, and it's worse than unpaired arms: given S1 the mechanism wasn't running, so the
numbers can't be evidence for it. Kept as measurements, relabelled, with the reasons named
(inert protection; fewer breakpoints from S2/S3 lowering cache-write arithmetically;
unpaired arms). Mechanism is verified by test; the effect needs a 50-task paired re-run
with `strip_caller_breakpoints: true`. I did **not** re-run benchmarks to re-establish it —
that's a fresh 50-task arm, not something to slip in under a review fix.

### Follow-ups

- **S8** — split rather than blanket-fixed, because the two fields differ. `cg_added_ms_avg`
  is a *real* measurement of the enforced path, and it reading ~0 in observe **is** the
  headline; zeroing it would hide the result. `llm_calls`/`llm_input_tokens`/
  `llm_output_tokens` are real money really spent, so moving them into `potential_*` would
  be a worse lie than leaving them — they stay where cost tooling reads them, labelled by a
  new `observe_llm_notice`. The over-broad claim is corrected in the how-to and the test.
- **S9** — `Tracker.Barren`: three consecutive unproductive jobs and the session stops
  enqueueing; any productive job resets it. Flat count, not a backoff — marked `ponytail:`
  with the reasoning.
- **S10** — the tracker keeps an `issued` high-water mark across all sessions, so a
  recreated session starts *above* any generation a surviving job could hold. `prevLen`
  deliberately still resets to 0: it's a claim about what the provider cached, and after
  eviction we don't know. Read #40 first as you suggested — no conflict, and I added
  structural forwarding of `FrozenLost` on `store.Buffer`, since a component asserting
  `c.Store.(store.FrozenLoser)` sees the *wrapper* on every off-path run and would have
  silently taken the degraded path. Asserted on a locally-declared shape so it compiles
  before #40 and binds on merge with no edit.
- **S11** — `Stop()` bounded at 2s and now returns `bool`. Cancelling can't interrupt an
  in-flight HTTP call to the cheap model, and nothing waits on that result.

### Gates (re-run after the fixes)

`build` · `test -tags cg_skeleton` · **`-race -count=5`** · `gofmt -l` · `go vet` ·
`mkdocs build --strict` — all clean.

### One thing I'm leaving

Async's value is now visibly conditional: inert on claude-code unless
`strip_caller_breakpoints` is set, and much of what it computes is discarded at agent turn
rates. That's the honest state and it's documented as such. Whether it's worth shipping on
that basis is a call worth making explicitly rather than having the docs imply an
unconditional win — happy to split async out and land observe alone if you'd rather.

PR #36, opened 2026-08-10
fix(cache): write cacheinject's breakpoints to the wire as metadata

### LINKED ISSUE #32: fix(cache): cacheinject's breakpoints never reach the wire — 46 applied, 0 forwarded on real traffic
## Problem statement

**`cacheinject` applies breakpoints that never reach the wire on real Claude Code traffic.** Measured over 40 real captured requests: **46 breakpoints applied at the component level, 0 in the output body.** Every mark landed on a message the writeback layer then discarded.

This makes the repo's flagship cache component a no-op on its primary workload, and it explains the benchmark result that has been reported as "cacheinject is inert": it is inert not because placement has no headroom, but because **its output is thrown away before it is sent**.

## Root cause

`cacheinject` can only mark messages that carry `ContentBlocks`. On Claude Code traffic those are exclusively **`assistant` messages carrying `tool_use`**. Those are precisely the messages whose slot is not round-trippable, so `apply/apply.go:222-227` drops the change:

```go
if !s.lossless {
	// bifrost can't round-trip this message; splicing our re-marshal would
	// drop provider fields it doesn't model. Discard the change, keep the
	// original bytes. ponytail: correctness over the marginal saving here.
	continue
}
```

`s.lossless` is `jsonEqual(rawBody, bifrostRoundTrip)`. bifrost v1.7.0 drops `tool_use.id/name/input` on unmarshal:

```
ORIG:      {"role":"assistant","content":[{"type":"tool_use","id":"toolu_...","name":"Bash","input":{...}}]}
ROUNDTRIP: {"role":"assistant","content":[{"type":"tool_use"}]}
```

So `lossless == false` for every assistant message, and every `cacheinject` mark on one is silently dropped. The discard itself is **correct** — splicing that re-marshal would corrupt the request. The bug is the combination: a component whose only possible targets are exactly the messages that can never be written back.

Measured probe result: `onLosslessSlots = 0` in all 40 requests.

## Why this matters more than it looks

`docs/components/cacheinject.md` presents a four-rule cost-minimisation policy and PR #22 rewrote the component around it. On the benchmarked workload **none of that policy has ever taken effect.** Every conclusion drawn about breakpoint placement from these runs measured a component whose output was discarded — including the "placement headroom is ~0%" finding. That conclusion may well survive a real test, but it has not had one.

`/stats` does not reveal this: `Mutated > 0` (the component did change the in-memory request), so `cacheinject` is correctly excluded from `top_passthrough` (`metrics/metrics.go:244-247`) while contributing nothing. It looks like a working component that saves no tokens — which is exactly what a Reformat is supposed to look like.

## Second, independent defect: the breakpoint budget ignores `system` and `tools`

`cacheinject` counts existing breakpoints by walking `req.Input` only. Real Claude Code traffic puts 2 of its 3 breakpoints in the top-level **`system`** array: measured **1,771 of 1,794 requests carry exactly `(system=2, tools=0, messages=1)`**. So the component sees `existing=1` and computes `budget = 4 - 1 = 3`, when the true remaining budget is **1**.

Probe: on a synthetic 60-message request with one caller message-breakpoint, `cacheinject` applied **4** message breakpoints (indices 38, 40, 57, 59) -> **6 on the wire** once the 2 system ones are counted. Anthropic's hard cap is 4; over it, the request 400s.

This has not been observed in production **only because defect 1 suppresses the marks.** Fix defect 1 without fixing this and the result is a live 400.

## Desired behavior

1. `cacheinject`'s breakpoints reach the wire on real agent traffic.
2. The 4-breakpoint budget is computed against **all** breakpoint locations — `system`, `tools`, and `messages` — never exceeding the provider cap.
3. No request is ever corrupted to achieve either. The `!s.lossless` guard exists for a real reason and must not simply be removed.

## Relevant code locations

- `apply/apply.go:222-227` — the discard.
- `apply/apply.go:65-85` — `slotKind`, `slot.lossless`; `:394-437` `normalize` where `lossless` is computed.
- `apply/apply.go:201-246` — the per-slot writeback loop.
- `components/reformat/cacheinject.go:107-187` — `Reformat`; `:143-157` the budget computation that ignores `system`/`tools`; `:23-26` `maxBreakpoints = 4`, `lookbackBlocks = 20`.
- `apply/prefixsplit.go` — the precedent for body-level `system`-array surgery, and proof that touching `system` from `apply` is an accepted pattern.
- `apply/apply.go:276-310` — `hasCacheBreakpoint` already does structural gjson queries across `system`/`tools`/`messages`; the budget count should use the same approach.
- `metrics/metrics.go:244-247` — why this is invisible in `/stats`.

## Proposed architecture

`cache_control` is **metadata, not content** — adding it neither changes what the model reads nor what the provider hashes for content purposes. It therefore does not need bifrost's message model at all, which is what makes a clean fix possible.

**Preferred: move breakpoint placement to a body-level operation, like `splitVolatileTail`.** Set `cache_control` with a targeted `sjson` write at a precise path (`messages.<i>.content.<b>.cache_control`) on the **original raw bytes**, never via a re-marshal of a bifrost message. Nothing else in the message is touched, so the losslessness concern does not arise. The component then computes *where* the breakpoints should go (its real value) and hands back positions; `apply` performs the metadata write.

This also fixes defect 2 naturally, since a body-level operation can see `system` and `tools`.

**Alternative: a metadata-only writeback path** — extend `slot` with a third kind that permits writing a whitelisted set of metadata keys via `sjson` even when `lossless == false`, since such a write provably cannot drop a provider field. Narrower blast radius, but adds a concept to the writeback layer.

**Rejected: relax or remove `!s.lossless`.** It would corrupt `tool_use` messages. Not an option.

**Also worth doing regardless:** a diagnostic counter for "component changes discarded by the writeback layer", per component. This class of bug should be loud. Its absence is why this survived two full benchmark studies.

## Alternatives considered

- **Upgrade or patch bifrost so `tool_use` round-trips.** Fixes the root cause for all components, not just this one, and would let `cacheinject` work unchanged. But it is an upstream dependency change with a much wider blast radius, and the metadata write does not need it. Worth investigating in parallel and reporting on; not the path for this fix.
- **Have `cacheinject` target only user/tool messages.** Would produce *some* wire effect but abandons the policy's core rule (a breakpoint on the last block every turn), since the last message is frequently an assistant turn.

## Observability implications

- Per-component **discarded-change** counter, surfaced in `/stats` and the dashboard (#30). A component that mutates and is then discarded must not be indistinguishable from one that works.
- A wire-level breakpoint count (total across `system` + `tools` + `messages`) so exceeding the cap is caught in telemetry, not by a 400.

None. `cg:cidx:<session>` digest tracking is unaffected.

## Backward compatibility

This is a behavior change on every Anthropic-family request that previously had its marks silently dropped — i.e. the fix makes `cacheinject` start working. That is benchmark-visible and **must be measured, not assumed**. It also means: for the first time, the component's cost policy will actually be exercised, so the possibility that it makes things *worse* is real and must be tested. `/stats` shape stays additive.

## Configuration design

No new configuration. If the fix proves cacheinject harmful once live, the answer is to remove it from the default preset — a config change, not a new knob.

## Testing plan

- Unit: a `cacheinject` mark on an assistant `tool_use` message **reaches the output body**, and the message is otherwise byte-identical (assert `tool_use.id`, `name`, `input` all survive).
- Unit: total wire breakpoints never exceed 4 with `system=2, tools=0, messages=1` incoming — the exact real-traffic shape.
- Unit: the synthetic 60-message case that produced 6 marks now produces at most 4.
- Unit: budget counting sees `system` and `tools` breakpoints.
- Unit: a discarded change increments the new counter.
- Regression: replay real captured requests (`CONTEXT_GURU_CAPTURE` streams exist) and assert breakpoints appear on the wire and the request stays valid JSON with all provider fields intact.
- `go test -race`.

## Real-world benchmark plan

This is the first honest test of the cacheinject policy, so measure it properly. Use the `cacheonly` arm (`deploy/harbor/swebench.py` `CUSTOM_CONFIGS`) — it removes **no** content tokens, so any cost delta vs `off` is purely breakpoint placement. Run 2-3 SWE-bench and 2-3 Terminal-Bench tasks, `cacheonly` vs `off`, before and after this fix.

Record per trial: commit SHA, config, model, task id, **fresh input / cache-read / cache-write / output tokens**, cache-hit rate, total billed cost, steps, reward, latency, and the wire breakpoint count. The decisive question: **now that the policy actually reaches the provider, does it help, do nothing, or hurt?** Report whichever it is. If placement genuinely has ~0% headroom against claude-code's own breakpoints, this fix will show it — and that is a publishable result either way. Watch specifically for a cache-write increase, since a badly placed breakpoint converts reads into writes at 11.5x cost.

## Acceptance criteria

- [ ] `cacheinject` breakpoints verifiably reach the wire on real captured Claude Code traffic.
- [ ] Provider fields (`tool_use.id/name/input`) survive intact — asserted, not assumed.
- [ ] Wire breakpoint total never exceeds 4; budget counts `system` + `tools` + `messages`.
- [ ] Discarded-change counter exists per component and is visible in `/stats`.
- [ ] `cacheonly` vs `off` measured before/after on real tasks; the verdict on placement headroom reported honestly, including if it is zero or negative.
- [ ] `docs/components/cacheinject.md` corrected — it currently documents a policy that has never taken effect on the benchmarked workload.
- [ ] `go test -race` green.

## Documentation updates

`docs/components/cacheinject.md` (add the finding, correct any claim resting on the discarded-mark runs), `docs/design.md` (the writeback losslessness constraint and the metadata-write exception), `docs/results/improvement-plan.md` (A5 currently calls cacheinject "provably inert" — the reason is different from what was assumed and needs restating), and a note in the Terminal-Bench/SWE comparison pages that the `cacheinject` per-component rows measured a suppressed component.

## Dependencies

Blocks nothing, but **should land before #30's per-component economics panel** is trusted, and before any further conclusion is drawn about breakpoint placement. Independent of #25/#27/#28/#29. Related to #31: async mode's cache policy assumes breakpoint placement works.

OsherElhadad:
Closes #32

`cacheinject`'s breakpoints never reached the provider on Claude Code traffic. Measured over 40 captured requests: **46 breakpoints applied at the component level, 0 in the output body.** The repo's flagship cache component was a no-op on its primary workload, and every conclusion drawn about breakpoint placement from the benchmark studies measured a component whose output was discarded before it was sent.

## Failing-then-passing evidence

Both new tests fail on `main` and pass here. Run against the unmodified `apply.go` + `cacheinject.go`:

```
=== RUN   TestCacheinjectReachesTheWire
    apply_test.go:291: expected cacheinject's breakpoint to change the body
--- FAIL: TestCacheinjectReachesTheWire (0.06s)
=== RUN   TestWireBreakpointCapRealTrafficShape
    apply_test.go:360: 6 breakpoints on the wire — the provider caps at 4 and 400s above it
--- FAIL: TestWireBreakpointCapRealTrafficShape (0.00s)
```

The second failure is defect 2 firing on its own: **6 breakpoints**, which the provider rejects with a 400. That is why both defects had to land together — fixing the discard alone would have shipped a live 400.

Replaying real captured traffic through the two code paths:

| capture | requests | breakpoints reach the wire — before | after |
|---|--:|--:|--:|
| `capture-swe.jsonl` | 19 | 1 | **16** |
| `capture-tb.jsonl` | 73 | — | **70** |
| fresh live `off` arm (this PR's run) | 44 | — | **41** |

The 1-of-19 on `main` is the single message bifrost happens to round-trip; the replay test's gate therefore requires a majority, not merely nonzero.

## The decisive wire metric

Distribution of the total breakpoint count the provider sees, over the 44 requests captured live during this PR's benchmark run:

| | 3 breakpoints | 4 | 5+ |
|---|--:|--:|--:|
| inbound (what claude-code sends) | 44 | 0 | 0 |
| outbound (after this PR) | 3 | 41 | **0** |

Exactly one breakpoint added — the single slot that was actually free — and never a cap breach. This also confirms the issue's traffic measurement on fresh traffic: **45 of 45 requests carry exactly `(system=2, tools=0, messages=1)`.**

## Design choice: body-level metadata write

Took the issue's preferred option over the metadata-slot-kind alternative.

`cache_control` is **metadata, not content** — it changes nothing the model reads, so it needs no bifrost message model to express. When a component's *only* change to a message is an added `cache_control` key, `apply` writes that key at its exact path (`messages.<i>.content.<b>.cache_control`) on the **original raw bytes** via `sjson`. A write that reads no other field cannot drop one, so the losslessness concern does not arise rather than being traded away.

The `!s.lossless` guard is **not** relaxed. It prevents real corruption and stays exactly as strict:

- `metadataOnlyWrites` proves "only that" by diffing pre against post *with the added keys removed* and requiring equality. A text edit, a removed key, a changed block count, a role change — all still discarded.
- `applyMetaWrites` refuses if the raw body's block layout disagrees with the normalized view, so a key can never land on the wrong block, and never overwrites a breakpoint the caller set.
- `TestMetadataOnlyWritesRejectsNonMetadataChanges` pins all seven cases.

Rejected the metadata-only slot kind: it adds a concept to the writeback layer for the same effect, and a body-level operation is also what lets defect 2 be fixed at all, since it can see `system` and `tools`.

## Defect 2: the budget was blind in two ways, not one

The issue names `system`/`tools`. Replaying real captures surfaced a second blind spot the issue did not have: the agent's *message* breakpoint sits on a `tool_result` block, and bifrost drops `cache_control` there too. So on real traffic **all three** of claude-code's breakpoints were invisible to the component, not two — it computed `budget = 4 − 0` from what it could see.

Counting in the component could never be correct, so `apply` counts structurally from the raw body (`wireBreakpoints`, the same gjson-path approach as `hasCacheBreakpoint`) and passes the total as `Ctx.ExistingBreakpoints`. `apply` also counts its own output and logs an error on a breach, so a cap violation appears in telemetry instead of as a provider 400.

## Observability: discards are loud now

A mutated-then-discarded component was byte-indistinguishable from a working Reformat — `Mutated > 0`, correctly excluded from `top_passthrough`, contributing nothing. That is why this survived two full benchmark studies.

`Pipeline.RecordDiscards` attributes each thrown-away change back to the component that made it (via `Report.ChangedIdx`), surfacing as per-component `discarded_changes` and `top_discarded` in `/stats`. **Both fields are additive** — no existing `/stats` key is renamed or removed, so `deploy/harbor/*.py` keeps parsing unchanged. A `Discarded` report is attribution, not a run, and is asserted not to inflate `Runs`.

## Verification

```
CGO_ENABLED=1 go build -tags cg_skeleton ./...      # clean
CGO_ENABLED=1 go test  -tags cg_skeleton ./...      # all green
CGO_ENABLED=1 go test  -tags cg_skeleton -race ./apply/ ./components/... ./metrics/ ./proxy/   # green
make lint                                          # go vet + gofmt -l clean
```

Tests added: breakpoint reaches the wire with `tool_use.id`/`name`/`input` intact and nothing else altered; wire total within 4 on the real traffic shape; the 60-message synthetic case that produced 6; budget sees invisible breakpoints; four existing breakpoints exhaust the budget; the metadata-write allow/deny matrix; a discarded change increments the counter; and a capture-replay regression (gated on `CONTEXT_GURU_CAPTURE`, skipped in CI).

## Docs

- `docs/components/cacheinject.md` — a prominent correction at the top. Every placement figure on that page measured a suppressed component. The `acted=0` / "placement contributes $0" findings are true *as recorded* (nothing reached the provider, so nothing could contribute) but are **not** evidence that placement lacks headroom — that question had never been asked. The volatile-tail-split figures are unaffected; that path never went through the discard.
- `docs/design.md` — the writeback losslessness constraint, the metadata exception and its guards, host-side breakpoint budgeting, and discard attribution.
- `docs/results/improvement-plan.md` lives on the `docs/terminal-bench` branch, not `main`, so its A5 correction is not in this PR's diff. **A5 calls `cacheinject` "provably inert" because the 4th breakpoint lands inside the same 20-block window as BP3 — that reasoning describes a component that placed the breakpoint. It never did.** A5's conclusion may survive, but its stated reason is wrong and it should be restated on that branch before the sticky-anchor work builds on it.

## Benchmark — and what it does NOT show

`cacheonly` vs `off`, SWE-bench Verified, `aws/claude-sonnet-5`, n=1 per arm. `off`'s second trial died on a Docker-compose error, so only `astropy-12907` completed in both arms:

| metric | off | cacheonly | delta |
|---|--:|--:|--:|
| steps | 14 | 11 | −21.4% |
| billed cost | $0.17607 | $0.14931 | −15.2% |
| cache-read | 609,968 | 461,844 | −24.3% |
| cache-write | 8,695 | 11,062 | +27.2% |
| cache-hit | 98.59% | 96.66% | −1.93 pp |

The −15.2% is **not a saving** — the agent took 3 fewer steps and cost tracks steps at corr 0.95 here. Per step: cost +7.9%, cache-write +61.9%.

**No mechanism is established for the cache-write difference, and my original explanation is disproven.** I claimed the mark lands at `len−1`, above claude-code's own breakpoint. Measured across three captures: **0 of 106 marks land above** — ours consistently sits one message *below*, where the policy's own Rule 2 says an extra breakpoint costs exactly zero.

**`acted=0` does not isolate placement.** It rules out content compaction only. The `cacheonly` arm still runs `splitVolatileTail`, so the arm is "placement + split". A single trial per arm also cannot separate either from the step-count nondeterminism that produced the 3-step gap.

So: one task, once, contended box, degenerate control, no mechanism. **The cache-write direction should not block this PR**, and placement's value has still never been measured. That is the reason for the preset change below, not the table.

## Presets: `cacheinject` is no longer enabled by default

Its breakpoints only began reaching the provider with this PR, and placement has never been shown to help — so shipping it on by default would enable an unmeasured policy on every request. The issue anticipated this exactly: *"if the fix proves cacheinject harmful once live, the answer is to remove it from the default preset — a config change, not a new knob."*

This required separating two mechanisms that shared one config entry. `splitVolatileTail` was gated on `cacheinject` only because it needed somewhere to hang, but the split is **measured** (−34.1% cost, 0% → 96.7% hit in an isolated A/B) while placement is not. Dropping `cacheinject` alone would have silently disabled the split too — turning "disable an unproven component" into a real cost regression. So a `cachesplit` marker component now carries the split and the presets use it; it stays gated rather than unconditional so `off` remains a true passthrough A/B control. `TestNoPresetEnablesCacheinjectByDefault` locks both halves in.

---

## Benchmark verdict — placement moves cache-write the WRONG way (n=1, not a claim)

The `off` control arm came back **degenerate**: 1 of 2 trials died on a Docker-compose `RuntimeError` (infrastructure, not the proxy), so the aggregate table is **not comparable**. One task, `astropy-12907`, completed in both arms:

| metric | off | cacheonly | delta |
|---|--:|--:|--:|
| steps | 14 | 11 | −21.4% |
| billed cost | $0.17607 | $0.14931 | −15.2% |
| cache-read | 609,968 | 461,844 | −24.3% |
| **cache-write** | **8,695** | **11,062** | **+27.2%** |
| cache-hit | 98.59% | 96.66% | −1.93 pp |

**The headline −15.2% is not a win — it is the step count.** The agent took 3 fewer steps, and cost tracks steps at corr 0.95 on this traffic. Normalising that out inverts the sign:

| per step | off | cacheonly | delta |
|---|--:|--:|--:|
| cost | 0.01258 | 0.01357 | **+7.9%** |
| cache-read | 43,569 | 41,986 | −3.6% |
| **cache-write** | **621** | **1,006** | **+61.9%** |

Verified this is purely placement: **0 of 78 requests changed content tokens** (`acted=0` across 39 runs), which is exactly what the `cacheonly` arm exists to guarantee.

**Mechanism, and it is the one the issue told us to watch:** the component spends its one free slot at `len−1`, above claude-code's own breakpoint, shortening the readable prefix and converting reads into writes at 11.5× cost. That is v1's regression mechanism — now reachable for the first time, because before this fix the marks never left the process.

**This is n=1 with a degenerate control on a contended box. It is not a number, it is a sign** — and it is the sign that costs money.

## Recommendation

**Merge this as a correctness fix, on the wire evidence, which is independent of cost.** Then settle placement economics in a separate, properly-powered study before `cacheinject` stays in any default preset. The issue anticipated this outcome: *"If the fix proves cacheinject harmful once live, the answer is to remove it from the default preset — a config change, not a new knob."*

## Known gaps

- **Terminal-Bench never ran.** It was stopped to end a harness collision (see below) and the SWE re-run was prioritised. Harness ready at `/tmp/i32-runs/terminalbench_i32.py`.
- **A harness bug of mine invalidated the first run**, reported rather than hidden: the TB harness kills by binary name and shared `cg-proxy-i32` with the SWE run, so launching TB killed the SWE proxy mid-arm. Fixed with a distinct binary name; the stale logs are quarantined as `INVALID-killed-run-*`.
- The added breakpoint sits at `len−1`; if it lands above the agent's own it is the read→write conversion measured above. Placement policy is the follow-up, not this PR.
- `docs/results/improvement-plan.md` is not on `main` (it lives on the `docs/terminal-bench` branch, #23), so its §A5 correction is not in this diff. A5 says cacheinject is "provably inert… while consuming the slot A1 needs" — **pre-fix it consumed nothing, so that clause was false; post-fix it becomes true.** It should be restated there before any sticky-anchor work builds on it.

### DIFF
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -170,6 +170,10 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 		CtxWindow:    window,
 		CacheAware:   cacheAware,
 		MaxCachedIdx: maxCachedIdx,
+		// Every breakpoint already on the wire — including the ones no component can
+		// see (`system`, `tools`, and cache_control on blocks bifrost drops). The
+		// provider's cap of four counts them all (issue #32, defect 2).
+		ExistingBreakpoints: wireBreakpoints(body),
 	}
 
 	// Canonical form of each normalized message BEFORE the pipeline, so a
@@ -179,7 +183,7 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 		normPre[i], _ = json.Marshal(norm[i])
 	}
 
-	pipe.Run(chat, c)
+	rr := pipe.Run(chat, c)
 
 	// A component changed the message count (summarize restructures the transcript
 	// to [msg0, <summary>, last-K]). Rebuild the messages array preserving each
@@ -198,6 +202,9 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	// if no component changes a message.
 	changed := systemSplit
 	var changes []change
+	// Per-message count of changes this writeback threw away, attributed back to the
+	// components that made them once the loop is done.
+	discarded := map[int]int{}
 	for i := range chat.Input {
 		s := slots[i]
 		switch s.kind {
@@ -222,8 +229,20 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 			}
 			if !s.lossless {
 				// bifrost can't round-trip this message; splicing our re-marshal would
-				// drop provider fields it doesn't model. Discard the change, keep the
-				// original bytes. ponytail: correctness over the marginal saving here.
+				// drop provider fields it doesn't model. But if the ONLY change is added
+				// `cache_control` — metadata, not content — write those keys at their exact
+				// paths on the original raw bytes: nothing else is read or rewritten, so no
+				// provider field can be dropped. See metawrite.go (issue #32).
+				if w, ok := metadataOnlyWrites(s.pre, post); ok {
+					if nb, ok := applyMetaWrites(out, s.path, len(chat.Input[i].Content.ContentBlocks), w); ok {
+						out = nb
+						changed = true
+						continue
+					}
+				}
+				// Anything else: discard the change, keep the original bytes.
+				// ponytail: correctness over the marginal saving here.
+				discarded[i]++
 				continue
 			}
 			if out, err = sjson.SetRawBytes(out, s.path, post); err != nil {
@@ -235,8 +254,15 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 			changes = append(changes, mkChange(s.path, schema.MessageText(pm), schema.MessageText(chat.Input[i])))
 		}
 	}
+	pipe.RecordDiscards(rr, discarded)
 	if changed && dumpPath != "" {
 		dumpChanges(c.Session, changes)
+	}
+	if n := wireBreakpoints(out); n > maxWireBreakpoints {
+		// Should be unreachable: the component budgets against OuterBreakpoints. Loud
+		// rather than a 400 from the provider, and it names the request shape.
+		slog.Error("context-guru: cache breakpoint count exceeds the provider cap",
+			"breakpoints", n, "cap", maxWireBreakpoints, "session", c.Session)
 	}
 	return out, changed
 }
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -132,11 +132,18 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	// Volatile-tail split, before anything else touches the body. This is a
 	// body-level concern rather than a component one: the pipeline operates on
 	// `messages`, but the block that needs splitting lives in the top-level `system`
-	// array, which components never see. Gated on cacheinject being configured, so it
-	// is opt-in via the same pipeline entry and adds no new config surface. See
-	// prefixsplit.go for what it splits and why.
+	// array, which components never see. See prefixsplit.go for what it splits and why.
+	//
+	// Gated on `cachesplit` OR `cacheinject`. The two were coupled only because the split
+	// had to hang off some existing config entry, but they are independent mechanisms
+	// with very different evidence: the split is measured (−34.1% cost, 0% → 96.7% hit in
+	// an isolated A/B), while breakpoint PLACEMENT has never been measured — so #32 drops
+	// cacheinject from the default presets and puts `cachesplit` there instead. Without
+	// its own gate the split would have gone down with cacheinject, turning "disable an
+	// unproven component" into a real cost regression. It stays opt-in rather than
+	// unconditional so `off` remains a true passthrough control for A/B runs.
 	systemSplit := false
-	if !bypass && pipe != nil && pipe.Has("cacheinject") {
+	if !bypass && pipe != nil && (pipe.Has("cachesplit") || pipe.Has("cacheinject")) {
 		body, systemSplit = splitVolatileTail(body, provider)
 	}
 
@@ -171,7 +178,7 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 		CacheAware:   cacheAware,
 		MaxCachedIdx: maxCachedIdx,
 		// Every breakpoint already on the wire — including the ones no component can
-		// see (`system`, `tools`, and cache_control on blocks bifrost drops). The
+		// see (`system`, `tools`, and the marks our own normalize drops). The
 		// provider's cap of four counts them all (issue #32, defect 2).
 		ExistingBreakpoints: wireBreakpoints(body),
 	}
@@ -258,11 +265,14 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	if changed && dumpPath != "" {
 		dumpChanges(c.Session, changes)
 	}
-	if n := wireBreakpoints(out); n > maxWireBreakpoints {
-		// Should be unreachable: the component budgets against OuterBreakpoints. Loud
-		// rather than a 400 from the provider, and it names the request shape.
+	// A cap breach WE caused is a bug and must be loud. A request that arrived already
+	// over the cap is the client's to fix — we forward it as-is (fail open), and an ERROR
+	// blaming context-guru for it would be a false alarm. So compare against the inbound
+	// count and only shout when we added to an over-cap total.
+	if n := wireBreakpoints(out); n > maxWireBreakpoints && n > c.ExistingBreakpoints {
 		slog.Error("context-guru: cache breakpoint count exceeds the provider cap",
-			"breakpoints", n, "cap", maxWireBreakpoints, "session", c.Session)
+			"breakpoints", n, "inbound", c.ExistingBreakpoints,
+			"cap", maxWireBreakpoints, "session", c.Session)
 	}
 	return out, changed
 }
--- a/apply/apply_test.go
+++ b/apply/apply_test.go
@@ -3,6 +3,7 @@ package apply_test
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"reflect"
 	"strings"
 	"testing"
@@ -455,4 +456,105 @@ func TestDiscardedChangeIsCounted(t *testing.T) {
 	if r := snap.Components["testrewrite"].Runs; r != 1 {
 		t.Fatalf("Runs should stay 1, got %d", r)
 	}
+}
+
+// reverter always errors, so the pipeline rolls its change back. A reverted component
+// must NOT be charged a discard: its change never reached the writeback layer at all.
+type reverter struct{}
+
+func (reverter) Name() string                 { return "testrevert" }
+func (reverter) Enabled(*components.Ctx) bool { return true }
+func (reverter) Reformat(req *bschemas.BifrostChatRequest, _ *components.Report, _ *components.Ctx) error {
+	for i := range req.Input {
+		if req.Input[i].Role == bschemas.ChatMessageRoleAssistant && req.Input[i].Content != nil {
+			for b := range req.Input[i].Content.ContentBlocks {
+				if req.Input[i].Content.ContentBlocks[b].Text != nil {
+					s := "mutated then reverted"
+					req.Input[i].Content.ContentBlocks[b].Text = &s
+				}
+			}
+		}
+	}
+	return errors.New("deliberate failure so the pipeline reverts")
+}
+
+func init() {
+	components.Register("testrevert", func([]byte) (components.Component, error) { return reverter{}, nil })
+}
+
+// A rolled-back component must not appear as a discard. Otherwise the counter meant to
+// catch #32-class bugs becomes a false-positive generator.
+func TestRevertedComponentNotChargedDiscard(t *testing.T) {
+	// testrewrite runs AFTER and produces the discard, so the writeback layer really does
+	// throw a change away at that index. Pre-fix, testrevert's ChangedIdx was recorded
+	// before the rollback, so it got charged for a discard caused by the other component.
+	cfg := pipe(t, "pipeline: [testrewrite, testrevert]\n")
+	agg := metrics.NewAggregator()
+	p, _ := cfg.Build(agg)
+
+	body := []byte(`{"model":"claude-x","messages":[
+		{"role":"user","content":"go"},
+		{"role":"assistant","content":[{"type":"text","text":"narration"},{"type":"tool_use","id":"toolu_r","name":"Bash","input":{}}]}
+	]}`)
+	apply.Body(context.Background(), p, store.NewMemory(store.Options{}), bschemas.Anthropic, body, "", false)
+
+	cs := agg.Snapshot().Components["testrevert"]
+	if cs.Reverted != 1 {
+		t.Fatalf("expected the component to be reverted, got %+v", cs)
+	}
+	if cs.Discarded != 0 {
+		t.Fatalf("reverted component charged %d discards — its change never reached writeback", cs.Discarded)
+	}
+}
+
+// One discarded message is charged to exactly ONE component, not to every component
+// that touched it. Two components rewrite the same unmodellable message; only the last
+// one's change is what the writeback layer actually threw away.
+func TestDiscardChargedOnceNotPerToucher(t *testing.T) {
+	cfg := pipe(t, "pipeline: [testrewrite, testrewrite2]\n")
+	agg := metrics.NewAggregator()
+	p, _ := cfg.Build(agg)
+
+	body := []byte(`{"model":"claude-x","messages":[
+		{"role":"user","content":"go"},
+		{"role":"assistant","content":[{"type":"text","text":"a long narration both components rewrite"},{"type":"tool_use","id":"toolu_d","name":"Bash","input":{}}]}
+	]}`)
+	apply.Body(context.Background(), p, store.NewMemory(store.Options{}), bschemas.Anthropic, body, "", false)
+
+	snap := agg.Snapshot()
+	total := snap.Components["testrewrite"].Discarded + snap.Components["testrewrite2"].Discarded
+	if total != 1 {
+		t.Fatalf("one discarded message charged %d times (testrewrite=%d testrewrite2=%d)",
+			total, snap.Components["testrewrite"].Discarded, snap.Components["testrewrite2"].Discarded)
+	}
+	// It must land on the LAST toucher — its change is the discarded state.
+	if snap.Components["testrewrite2"].Discarded != 1 {
+		t.Fatalf("discard should be charged to the last component to change the message, got %v", snap.TopDiscarded)
+	}
+}
+
+// contentRewriter2 is a second rewriter so two components touch one message.
+type contentRewriter2 struct{}
+
+func (contentRewriter2) Name() string                 { return "testrewrite2" }
+func (contentRewriter2) Enabled(*components.Ctx) bool { return true }
+func (contentRewriter2) Reformat(req *bschemas.BifrostChatRequest, _ *components.Report, _ *components.Ctx) error {
+	for i := range req.Input {
+		if req.Input[i].Role != bschemas.ChatMessageRoleAssistant || req.Input[i].Content == nil {
+			continue
+		}
+		for b := range req.Input[i].Content.ContentBlocks {
+			if req.Input[i].Content.ContentBlocks[b].Text != nil {
+				s := "even shorter"
+				req.Input[i].Content.ContentBlocks[b].Text = &s
+			}
+		}
+	}
+	return nil
+}
+
+func init() {
+	components.Register("testrewrite2", func([]byte) (components.Component, error) {
+		return contentRewriter2{}, nil
+	})
 }
--- a/apply/apply_test.go
+++ b/apply/apply_test.go
@@ -3,6 +3,7 @@ package apply_test
 import (
 	"context"
 	"encoding/json"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -11,8 +12,10 @@ import (
 	"github.com/rossoctl/context-guru/components"
 	_ "github.com/rossoctl/context-guru/components/all"
 	"github.com/rossoctl/context-guru/config"
+	"github.com/rossoctl/context-guru/metrics"
 	"github.com/rossoctl/context-guru/store"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 )
 
 func pipe(t *testing.T, yaml string) *config.Config {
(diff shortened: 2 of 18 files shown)


OsherElhadad (commented):
## Independent review — **changes needed** (do not merge as-is)

I did not write this code. Reviewed at `db58dbf` in a clean scratch worktree.

**What I verified myself, and it holds up.** Both claimed failures reproduce on `origin/main` with the exact quoted messages (`expected cacheinject's breakpoint to change the body`; `6 breakpoints on the wire`). Full suite, `-race` on `./apply/ ./components/... ./metrics/ ./proxy/`, and `make lint` are all green on the branch. The replay test passes on real captures with the reported numbers (`capture-swe` 16/19, `capture-tb` 70/73). The core mechanism is genuinely byte-clean: I stripped every added `cache_control` from the output of 115 real captured requests across three captures and the `messages` array came back **byte-identical to the input in 115/115**, including 8 requests where the mark landed on a `tool_result` block. `tool_use.id/name/input` survive. The correctness thesis is real and the fix is the right shape.

But there are five things wrong, two of them blocking.

---

### 1. BLOCKING — `wireBreakpoints` misses Bedrock `cachePoint` in `system`/`tools`, so the cap is still breachable

`apply/metawrite.go:106-112` lists `cachePoint` only under `messages.#.content.#.cachePoint`. Bedrock puts `cachePoint` as its own **array entry in `system` and `tools`** — exactly the two places defect 2 is about. The count comes back 0 and the component budgets 4 free slots.

Constructed the shape and ran it: inbound `system` has one `cachePoint`, `tools` has one, 60 messages, provider `Bedrock`.

```
bedrock: pre-existing cachePoints in system+tools = 2; component added 4 message marks => wire total 6
```

Six on the wire, on this branch, and **the cap-breach `slog.Error` at `apply/apply.go:261` does not fire either** — it uses the same blind counter. This is defect 2, un-fixed, on a provider the PR's own `explicitBreakpointProvider` and `cacheAware()` both declare in scope. `docs/components/cacheinject.md:24` claims "the wire total is asserted never to exceed 4"; that assertion does not cover Bedrock.

Fix: add `system.#.cachePoint` and `tools.#.cachePoint` to `breakpointPaths`, and add the Bedrock shape to `TestBreakpointCounting`. Two lines and a test case. Note `metawrite_test.go:47` uses only the Anthropic spelling, which is why this got through.

### 2. BLOCKING — `discarded_changes` misattributes, and can charge components that changed nothing on the wire

Two independent bugs in `RecordDiscards`, both reproduced:

**(a) a REVERTED component is charged a discard.** `components/pipeline.go:98` sets `rep.ChangedIdx = changedIdx(before, req.Input)` *before* the switch below it that does `req.Input = before` on error / never-worse / offload-contract violation. A component that mutated and was then rolled back keeps a populated `ChangedIdx` and gets billed for someone else's discard:

```
testrevert:  reverted=1  discarded=1     <- it contributed nothing; the pipeline undid it
top_discarded=[testrevert testrewrite]
```

Fix: move the `ChangedIdx` assignment into the `default:` (success) branch, or clear it wherever `rep.Reverted = true` is set.

**(b) one discard is charged to every component that touched the message.** With `pipeline: [testrewrite, cacheinject]` on one unmodellable message, exactly one change is discarded, but:

```
discarded per component: map[cacheinject:1 testrewrite:1]
top_discarded=[cacheinject testrewrite]
```

`cacheinject` is named as suppressed on a request where the discard was caused by the other component. Given the whole point of this counter is to make #32-class bugs loud, a counter that fingers innocent components will get ignored the first time someone chases a false positive. At minimum document it as "components implicated" rather than "changes discarded", or attribute to the last writer of each index.

### 3. `slog.Error` on cap breach fires on traffic context-guru never touched — including under `bypass`

`apply/apply.go:261` counts the **output** unconditionally. Under `bypass=true` the pipeline is a no-op, yet a client that itself sends 5 breakpoints produces:

```
2026/08/10 07:20:45 ERROR context-guru: cache breakpoint count exceeds the provider cap breakpoints=5 cap=4 session=...
```

A hard `ERROR` blaming context-guru for the client's request. Same on the non-bypass path when inbound was already over 4. The comment says "should be unreachable: the component budgets against `OuterBreakpoints`" — also note that field is named `ExistingBreakpoints`, not `OuterBreakpoints`; fix the comment. Gate the log on `n > max && n > wireBreakpoints(body)` (i.e. we made it worse), or skip it entirely when `bypass`.

### 4. The stated root cause of "the third breakpoint is invisible" is wrong — bifrost is not the culprit

Every comment and doc paragraph in this PR says bifrost drops `cache_control` on `tool_result` blocks: `components/component.go:132`, `components/reformat/cacheinject.go:149`, `apply/metawrite.go:117`, `docs/design.md:150`, `docs/components/cacheinject.md:21` and `:157`. I tested it directly and it is **false** — bifrost round-trips `cache_control` fine on both `tool_result` and `tool_use`:

```
cc on tool_result(string): rt={"role":"user","content":[{"type":"tool_result","cache_control":{"type":"ephemeral"}}]}
cc on tool_use:            rt={"role":"assistant","content":[{"type":"tool_use","cache_control":{"type":"ephemeral"}}]}
```

The actual cause is **this repo's own `normalize()`**: a string-content `tool_result` is expanded into a synthetic `role=tool` message via `toolMessage()` (`apply/apply.go:~455`), which builds a fresh `ChatMessage` and never carries the block's `cache_control` forward. Confirmed:

```
norm[0] role=tool slotKind=1 blockCacheControlVisible=false     <- string-form tool_result: mark lost
array-form norm[0] role=user slotKind=0 blockCacheControlVisible=true   <- array-form: mark visible
```

The chosen fix still works, because it routes around visibility entirely. But five comments and two docs pages now assert a bifrost defect that does not exist, and the next person to touch this — or to consider the "upgrade bifrost" alternative the issue lists — will be reasoning from a false premise. Please correct the attribution to `normalize`/`toolMessage`. It is also a smaller latent bug worth its own line: `toolMessage` losing `cache_control` is the reason `hasBreakpoint` can double-mark.

### 5. ~+65% hot-path latency, and the expensive half is the observability counter

Benchmarked `apply.Body` with `preset: general` on a real 100 KB+ captured request, 100x, 3 counts, same box, this branch vs `origin/main`:

```
branch: 5164679 / 5161131 / 5133808 ns/op
main:   3063430 / 3407773 / 3184364 ns/op       => +2.0 ms/request, ~+65%
```

Isolating: stubbing out `changedIdx` alone recovers ~1.2 ms; stubbing the two `wireBreakpoints` calls recovers ~0.7 ms. So most of the regression is `changedIdx` — a **full `json.Marshal` of every message, twice, per component** (`components/pipeline.go:141`), i.e. O(components × transcript) marshals added to the hot path purely for a diagnostic counter. It is as expensive as `CloneMessages` (3.9 ms vs 4.2 ms at 120 messages) and 113× `tokensOf`.

The repo tracks `AddedLatencyMsAvg` in `/stats` precisely because this matters. Cheapest fixes, in order of laziness: swap the double marshal for `reflect.DeepEqual(before[i], after[i])` — I measured this at ~1.2 ms cheaper, recovering the whole `changedIdx` cost; and hoist `wireBreakpoints(body)` so it is computed once and reused for the breach check instead of twice.

---

### The benchmark verdict — I disagree with one link in the chain, and it matters

The self-criticism is admirably honest and the per-step normalisation is legitimate in principle: with n=1 and a 3-step difference on traffic where cost correlates with steps at 0.95, the headline −15.2% is not attributable to placement, and inverting to +7.9%/step is the right instinct.

But **`acted=0 / 0 of 78 requests changed content tokens` is not the proof it is presented as.** `acted` counts *content-token* savings, and `cacheonly` guarantees no content change by construction — so `acted=0` is a tautology of the arm, not evidence about placement. It rules out "content compaction confounded the result". It says nothing about the other live confounders: the arm still runs `splitVolatileTail`, which adds a `system` block and moves a breakpoint, and agent nondeterminism produced the 3-step difference itself. n=1 with a degenerate control cannot separate placement from either.

**And the stated mechanism does not reproduce.** The PR says the component "spends its one free slot at `len−1`, above claude-code's own breakpoint, shortening the readable prefix". I measured where the added mark actually lands on all three captures:

```
capture-swe:      addedMarkAboveAgentOwn=0  addedBelowOrEqual=16  noMarkAdded=3
capture-tb:       addedMarkAboveAgentOwn=0  addedBelowOrEqual=70  noMarkAdded=3
capture-swebench: addedMarkAboveAgentOwn=0  addedBelowOrEqual=20  noMarkAdded=3
```

**Zero of 106.** The agent's own mark is on the final message; ours consistently lands at `highestAdded = highestAgentOwn − 1` (the `mark()` walk-down from `len−1` hits the message below). By the component's own Rule 2 — "writes are billed as a SPAN, so an extra breakpoint BELOW the top costs exactly zero" — the +61.9% cache-write should not have this cause. So the paragraph naming v1's regression mechanism is unsupported, and the one number in the run that would cost money has **no established mechanism at all**. That is a reason to distrust the number as much as the mechanism.

My read: **the cache-write direction should not block the merge, but the merge should not ship it enabled either.** The correctness case stands on wire evidence alone, the honest cost signal is unresolved in *both* directions, and holding a correctness fix hostage to a properly-powered study is the wrong trade — the study needs this code to exist. Concretely: land the fix, and in the same PR drop `cacheinject` from `general`/`codesmart`/`codesafe`/`balanced` (`config/config.go:127,152,165`) pending the placement study. That is exactly the config-change remedy the issue anticipated, and it is one line each. Shipping it in the default presets on an n=1 signal pointing the wrong way, with no reproducible mechanism, is the one outcome I'd push back on.

### Docs

`docs/components/cacheinject.md:133` ships a literal `Placeholder — filled by the cacheonly vs off measurement in #32` under a heading (`## What placement is actually worth`) that **three other passages on the page link to as the authoritative answer** (`:31`, `:107`). A reader following those links lands on nothing. Either put the n=1 result there with its caveats, or drop the section and the links until there is a measurement. Repo rule is docs must not present plans as shipped; a cross-referenced empty section is that rule's failure mode.

Otherwise the docs corrections are good and appropriately loud — the `!!! danger` blocks on `cacheinject.md` and `results/components.md` correctly retract the placement claims rather than quietly restating them, and `docs/design.md:124-155` describes what the code does. Two fixes needed: the bifrost attribution above, and `:24` overstates the cap guarantee (not true on Bedrock).

### Metrics — backward compatible, correctly locked

Verified `/stats` is additive only: `discarded_changes` and `top_discarded` are new keys, nothing renamed or removed. `deploy/harbor/*.py` reads `runs`/`acted`/`saved_tokens`/`saved_tokens_unique` by `.get()`/subscript on unchanged names — `swebench.py:298`, `measure.py:142`, `deep_analysis.py:96`, `analyze.py:90` all keep parsing. New counters are inside `a.mu` (`metrics.go:115-116`). The `Discarded > 0 ⇒ early return` guard correctly keeps `Runs` from double-counting, and the test pins it. One gap: an early return means a discard-attribution report never reaches the `Mutated`/`Acted` bookkeeping, which is right — but it also means `Slog.Component` (`metrics.go:52`) emits `discarded_changes` on *every* component line as a constant 0, which is dead field width on a hot log. Minor.

### Security

Clean. No credential, token, key, or gateway URL anywhere in the diff or the PR body. Test fixtures use `claude-x` / `toolu_abc` placeholders.

---

### Summary, ranked

| # | Severity | Issue |
|---|---|---|
| 1 | **blocking** | `wireBreakpoints` blind to Bedrock `system`/`tools` `cachePoint` → 6 on the wire, breach log silent too. Reproduced. |
| 2 | **blocking** | `discarded_changes` charges reverted components, and charges one discard to every component that touched the message. Both reproduced. |
| 3 | high | cap-breach `ERROR` fires on client-caused breaches and under `bypass`. Reproduced. |
| 4 | high | root cause misattributed to bifrost across 5 comments + 2 docs; it is this repo's `toolMessage()`. Disproved bifrost claim directly. |
| 5 | medium | +2.0 ms / +65% on `apply.Body`, mostly `changedIdx`'s double marshal for a diagnostic. Measured on both branches. |
| 6 | medium | `docs/components/cacheinject.md:133` is a `Placeholder` that three passages link to as the answer. |
| 7 | medium | benchmark's stated cost mechanism does not reproduce (0/106 marks land above the agent's own); `acted=0` is a tautology of the arm, not proof about placement. |
| 8 | low | `slog` emits a constant-0 `discarded_changes` on every component line; `OuterBreakpoints` in the `apply.go:262` comment should be `ExistingBreakpoints`. |

**Verdict: changes needed.** 1 and 2 are code bugs in the two things this PR exists to fix, both reproducible in under a minute. 3–5 are one-to-few-line fixes. On the economics I would not hold the merge — but I would drop `cacheinject` from the default presets in this same PR rather than after the study.

**What I took on trust:** the historical measurement counts (46 applied / 0 forwarded over 40 requests; 1,771 of 1,794 requests at `(system=2, tools=0, messages=1)`; the 44-request live inbound/outbound distribution) and the `astropy-12907` billing figures — I did not re-run the benchmark or re-derive those captures. Everything else in this review I reproduced.

OsherElhadad:
## Retracting my own framing of the cache-write finding

I previously treated the **+61.9% cache-write per step** as the most consequential result here and said it "argues the honest follow-up is not ship this but measure placement properly." The independent review refuted the mechanism behind that reading, and it was right to.

Two problems with the analysis — both cutting *against* the alarming interpretation:

1. **`acted=0` is a tautology of the `cacheonly` arm, not proof the delta is purely placement.** That arm is *defined* as removing no content tokens, so `acted=0` is guaranteed by construction. It rules out content compaction; it says nothing about `splitVolatileTail`, which is **still live in that arm**, nor about whatever nondeterminism produced the 3-step gap between the two trials.

2. **The stated mechanism does not reproduce.** The claim was that the mark lands at `len−1`, *above* claude-code's own breakpoint, shortening the readable prefix. The reviewer measured all three captures: **0 of 106 marks land above the agent's own** — ours consistently sits one message *below*, which is precisely where the component's own Rule 2 says an extra breakpoint costs **exactly zero**.

So the number that supposedly costs money has no established mechanism behind it. n=1, degenerate control, contended box, and a mechanism that measurement contradicts. **The cache-write direction should not block this merge.**

I got this wrong in a specific and repeatable way: I accepted an unfavourable number without checking whether the mechanism it was attributed to had actually operated. That is the same error as treating a favourable number the same way — and it is now the fourth instance in this workstream (see the `F-1` rule added to `docs/results/improvement-plan.md`). Skepticism applied only to good news is not skepticism.

**What still blocks:** the Bedrock `cachePoint` gap in `system`/`tools` (6 on the wire, reproduced), the `discarded_changes` double-misattribution, the breach ERROR firing under `bypass`, and the root-cause misattribution across five comments and two doc pages — bifrost does **not** drop `cache_control` on `tool_result`; this repo's own `toolMessage()` in `normalize` does. That last one matters most for the long run: the original bug survived two full benchmark studies precisely because the mechanism was misunderstood, and shipping a second wrong explanation sets up the next one.

**What I'm adding as a merge requirement:** drop `cacheinject` from the `general`, `codesmart`, `codesafe` and `balanced` presets in this PR. The issue anticipated exactly this — *"if the fix proves cacheinject harmful once live, the answer is to remove it from the default preset — a config change, not a new knob."* We now have a live cost signal with no explanation and no demonstrated benefit. That is not grounds for calling it harmful, but it is ample grounds for not shipping it **on by default** while the properly-powered placement study is still outstanding. The component stays available for that study.

PR #37, opened 2026-08-10
feat(cmdfilter): port rtk's filter coverage (3 -> 24) and fix three DSL loss-typing bugs

### LINKED ISSUE #29: feat(cmdfilter): port rtk's filter coverage (3 -> ~25 filters) and fix three DSL loss-typing bugs
## Problem statement

context-guru's `cmdfilter` ships a **3-filter starter set** (`pytest`, `npm-install`, `make` — `components/offload/cmdfilter.go:149-179`) against rtk's **63** declarative filters. The DSL engine is already a faithful Go port of rtk's — same 8 stages, same field names — so the gap is coverage, not capability.

Three engine bugs found while diffing must be fixed as part of this work; two of them make a real loss invisible.

## Current behavior

`components/dsl/dsl.go` (339 lines) implements the full pipeline: `match` → `strip_ansi` → `replace` → `match_output` (with `unless`) → `strip_lines_matching` XOR `keep_lines_matching` → `truncate_lines_at` → `head`/`tail` → `max_lines` → `on_empty`, plus inline `tests` (`:71-80`) and a `Lossiness` enum. `cmdfilter.go:92` already implements rtk's `never_worse` at message level.

**The structural difference that governs this whole port:** rtk matches `match_command` against a **shell command string**; `cmdfilter` matches `match` against the **first non-empty line of the tool output** (`selectorKey`, `cmdfilter.go:126-136`). So rtk's `^terraform\s+plan` regexes are **not portable as written** — most would never fire. Every ported filter needs its selector rewritten to an output-shape signature (as the existing `pytest` filter already does: `"(pytest|=+ test session starts)"`). That translation is the real work; budget for it.

This is also context-guru's structural **advantage**: rtk's hook only sees Bash calls, so Claude Code's built-in `Read`/`Grep`/`Glob` are entirely uncovered (stated in rtk's own README). A proxy sees every tool result regardless of origin.

### Engine bugs to fix

1. **`truncate_lines_at` never records loss.** `dsl.go:192-200` truncates each line but the `loss` variable is only initialized at `:202`, *after*. Intra-line loss is therefore invisible to the caller, so no recovery hint is emitted for a real loss.
2. **`truncate_lines_at` truncates silently.** No ellipsis (rtk appends `...`). A mid-line cut with no marker reads as corrupted output to a model.
3. **`recoveryHint` collapses `LossTail` and `LossWhole`.** `cmdfilter.go:137-142` returns the same "call `context_guru_expand`" hint for both. rtk emits a cheaper partial-recovery hint for a tail cut (`[see remaining: tail -n +N ...]`). A tail cut is recoverable more cheaply than a whole-blob loss and should say so.

## Motivation

Deterministic filtering is free (no LLM, ~0 latency) and cache-safe — it acts on the newest tool output, in the mutable tail. That is the ideal shape for this workload: `extract_llm` costs money and is ~8× underwater (#28), while `cmdfilter` costs nothing. Broadening its coverage is the cheapest available savings. Its current contribution is tiny (34 acts, 886 unique tokens on Terminal-Bench) purely because there are three filters.

## Desired behavior

Broad deterministic coverage of common command output, using context-guru's DSL, with each filter tested and measured.

**Do not add filters to increase a count.** For each: preserve important diagnostics, avoid hiding failures, ship tests, measure compression, document behavior.

## PROVENANCE — settled, but the obligation is real

Investigated (see the analysis backing this issue):

- rtk is at **v0.42.4**, licensed **Apache-2.0**, `Copyright 2024 rtk-ai and rtk-ai Labs`, repo `github.com/rtk-ai/rtk`.
- **There is no upstream filter source.** No vendor dir, no submodules, no generator, no download step in `build.rs`/CI, no NOTICE/THIRD-PARTY file, no attribution comments. `build.rs` only concatenates and validates local `src/filters/*.toml`. Git history shows filters added incrementally by named contributors and sometimes removed on review — the signature of hand-authoring.
- rtk's own CLAUDE.md calls it "a fork", but history begins with its own initial commit and the remote is the canonical repo. That line is stale self-description, **not** evidence of an upstream.

**Conclusion: rtk is the canonical source; there is nothing better to source from.**

**Attribution obligation.** Apache-2.0 §4 applies — the filter TOMLs are copyrightable expression (curated regex sets, threshold choices, test corpora). Required: retain the Apache-2.0 notice for the derived portion, state the changes made, and preserve the `Copyright 2024 rtk-ai and rtk-ai Labs` notice. `components/dsl/dsl.go:1-4` already credits rtk in prose — **good practice but not sufficient** for a bulk port of 60+ filter definitions. Add a `THIRD-PARTY-NOTICES` entry naming rtk, its URL, Apache-2.0, and "filter definitions and DSL pipeline adapted from, with modifications". Do not strip the notice. Apache-2.0 is permissive; there is no copyleft obstacle.

### Tier 1 — port now (high value, distinctive output signature, low lossiness)

| filter | why | suggested output selector |
|---|---|---|
| gradle | `> Task :… UP-TO-DATE` is pure noise, huge volume | `^> Task :\|^> Configuring project` |
| xcodebuild | 31 patterns, best noise:signal ratio in the set | `^CompileC\|^CompileSwift\|^Ld ` |
| terraform-plan / tofu-plan | `Refreshing state` + `# … unchanged`; rtk asserts ≥60% savings | `^Refreshing state\|Terraform will perform` |
| pulumi preview/up/refresh/destroy | 20 patterns each, incl. JS stack-frame pruning; all lossless | `^Previewing update\|^Updating (\|^Destroying (` |
| make | replace the existing builtin — rtk's also handles `Nothing to be done`; asserts ≥60% | `^make\[\d+\]:` |
| liquibase | ASCII banner + jar lists | `^Starting Liquibase at\|^Liquibase ` |
| turbo / nx | monorepo cache noise | `cache (hit\|miss\|bypass)\|> NX Running target` |
| gcc | keeps errors, strips include-traces only | `^In file included from\|warnings? generated` |
| ssh | `debug1:` flood | `^debug1:\|^OpenSSH_` |
| ping | `tail_lines = 4` is the whole filter | `^\d+ bytes from ` |
| tofu-init | provider download spam | `^- Downloading\|^Initializing provider` |
| bundle/poetry/composer/uv/brew installs | supersedes the crude `npm-install` builtin | `^Using \|^  - Installing \|^==> Downloading` |

### Tier 2 — port, but add an `unless` guard first
All of rtk's `match_output` success-collapse rules (build → `ok`). **Only 2 of 11 carry an `unless` guard**; the other 9 will swallow warnings that co-occur with a success marker. rtk learned this itself — its `swift-build` test is literally named "warnings not swallowed when Build complete present". In a proxy the stakes are higher than in a CLI: the agent cannot re-run the command to check. Add `unless: "error|warning|failed|Exception"` to every collapse rule that lacks one.

### Tier 3 — skip, with reasons
- **The ~24 `truncate_lines_at`-only filters** (df, ps, du, jq, jira, markdownlint, yamllint, stat, gcloud, helm, iptables, skopeo, yadm, hadolint…). Whole-blob lossy for a modest unmeasured saving; their "filter" is mostly "strip blanks + cap width". If the effect is wanted, one generic width-cap beats 24 filters.
- **`filter_stderr`** — no proxy analogue; streams are already merged upstream.
- **The blank-line-only linter filters** (shellcheck, systemctl-status, sops, fail2ban, basedpyright, ty, oxlint, biome, mix-format, tofu-fmt, tofu-validate) — fold the useful `on_empty` collapses into a couple of generic linter filters instead of 11 files.
- **spring-boot** — `keep_lines_matching` allowlist over unbounded app logs; too easy to drop the one line that mattered.
- **rtk's 86 command-detection rules** — entirely about rewriting shell commands to `rtk <cmd>`; irrelevant to a proxy. Its per-tool `savings_pct` figures are hand-estimated, not measured — do not reuse them as expected-gain data.
- **rtk's ~50 native Rust filters** (43,850 lines: `--format json` → parse → re-render for cargo/rubocop/golangci/ruff/phpstan, TRX/binlog parsing, git diffstat compression). The highest-compression technique rtk has, but inexpressible in the DSL, and mostly unreachable from a proxy that cannot inject `--format json` into a command it never sees. Out of scope — except output-side: JSON/TRX blobs that *arrive* as tool results could be re-rendered later.

Note on ordering: with output-line matching, filter order matters more than in rtk (a generic pattern can shadow a specific one). The registry currently sorts by name (`dsl.go:294`) — deterministic but arbitrary. Consider an explicit `priority`, or order specific-before-generic.

## Relevant code locations

- `components/offload/cmdfilter.go` — `:32` `disable_builtins`, `:45` load, `:92` `never_worse`, `:126-136` `selectorKey`, `:137-142` `recoveryHint`, `:149-179` the 3 builtins.
- `components/dsl/dsl.go` — `:41-52` `Def` schema, `:57-66` `ReplaceRule`/`MatchRule` (`unless`), `:71-80` registry + inline tests, `:185-216` the truncate/head-tail/max-lines stages (bugs 1 and 2), `:294` registry ordering, `:306` linear `Match`, `:320` `RunTests`.
- rtk source: `src/filters/*.toml` (63 files, 154 inline tests), `src/core/toml_filter.rs` (the engine + its guardrail tests), `src/core/truncate.rs:4-11` (the `CAP_*` classes), `src/core/guard.rs:6` (`never_worse`).

## Proposed architecture

1. Fix the three engine bugs first, with tests.
2. Port the inline test cases alongside each filter — they are pure `(filter, input) → expected` triples in the same DSL `tests` shape `dsl.RunTests` already implements (`dsl.go:320`), so they convert mechanically from TOML to YAML. This is what makes the port verifiable rather than hopeful.
3. Adopt rtk's **`CAP_*` shared budget classes** by signal density (`CAP_ERRORS=20`, `CAP_WARNINGS=10`, `CAP_LIST=20`, `CAP_INVENTORY=50`) with a safe deviation helper, instead of 60 hand-picked `max_lines`. rtk's own TOML filters do **not** do this — it is a better idea than its implementation, and it makes a large set tunable from one place.
4. Add a size floor before doing filter work (rtk's `MIN_TEE_SIZE = 500`): below it, the marker often costs more than the saving.
5. Add compile/load-time guardrails mirroring rtk's `build.rs`: reject duplicate filter names, validate on load, and assert every filter has at least one test.
6. Log selector **misses** so the next filters to write become data rather than guesswork (rtk's `parse_failures` ledger does exactly this).

## Alternatives considered

- **Port all 63 verbatim.** Rejected — ~24 are low-value width-caps, 9 have unguarded success-collapse, and none of the `match_command` regexes work against an output selector.
- **Write filters from scratch.** Forfeits 154 tested cases and years of tuning for no benefit; the license permits the port.
- **Depend on rtk as a binary.** Reintroduces its Bash-only ceiling — the exact limitation context-guru does not have.

## Observability implications

`cmdfilter` metrics broken down **by command family** (git / tests / builds / package-managers / IaC / …): acts, gross and unique tokens saved, and `never_worse` rejections. Plus selector-miss logging. Surface in `/stats` and the dashboard (**#30**) so it is visible which families pay off and which filters never fire.

None beyond the existing marker + stash. The size floor reduces stash pressure.

## Backward compatibility

`disable_builtins` keeps working. Replacing the `make` and `npm-install` builtins with better versions is a behavior change on any config using builtins — measure it. User-supplied filters are unaffected. If the registry gains a `priority` field, absent means today's behavior.

## Testing plan

- Unit: the three engine-bug fixes (`truncate_lines_at` sets loss; emits an ellipsis; `recoveryHint` distinguishes `LossTail`).
- Every ported filter carries its converted inline tests; all run in CI via `dsl.RunTests`.
- Guardrail tests: duplicate names rejected; every filter has ≥1 test; all builtins compile.
- Explicit negative tests for the Tier-2 `unless` guards: a build with a warning **and** a success marker must not collapse.
- Compression measured per filter on its fixture; assert a floor where rtk does (`terraform-plan`, `make` ≥60%).
- `never_worse` still holds for every filter.

## Real-world benchmark plan

2–3 Terminal-Bench tasks whose output the new filters actually target (build/IaC-heavy — pick from the `codesmart` dump by selector frequency) and 2–3 SWE-bench tasks as a no-regression control. Record commit SHA, config, task ids, `cmdfilter` acts and unique tokens saved by family, all four token tiers, billed cost, steps, reward, latency, expands, reverts. **Watch reward specifically**: this is the failure mode that sank rtk on Terminal-Bench (94% Bash-output compression → the agent lost information → +27% steps → highest cost of any arm). If reward drops, the filters are too aggressive — report it and back them off.

## Acceptance criteria

- [ ] Three engine bugs fixed with tests.
- [ ] Tier-1 filters ported with rewritten output selectors and their converted inline tests, all green.
- [ ] Every ported success-collapse rule has an `unless` guard.
- [ ] Tier-3 exclusions documented with reasons (this issue is the record).
- [ ] `THIRD-PARTY-NOTICES` entry for rtk (Apache-2.0, URL, "adapted with modifications").
- [ ] Load-time guardrails: duplicate names, validation, every-filter-has-a-test.
- [ ] Per-family metrics in `/stats` + dashboard; selector misses logged.
- [ ] Measured on real tasks: unique tokens saved up, **reward not worse**, steps not up. Numbers in the PR.
- [ ] `go test -race` green.

## Documentation updates

`docs/components/cmdfilter.md` (the full filter table, what each preserves, the Tier-3 rationale), `docs/components/dsl.md` (the loss-typing fix and any `priority` field), `docs/how-to/custom-dsl-filter.md`, and `THIRD-PARTY-NOTICES`.

## Dependencies

Largely independent — good parallel candidate. Feeds **#30** (per-family attribution). Its reward-regression watch is informed by the rtk arm in `docs/results/terminal-bench-comparison.md`.

OsherElhadad:
Closes #29

`cmdfilter` shipped a 3-filter starter set against rtk's 63. The DSL engine was already a faithful
port, so the gap was coverage — but the port is not a copy, and the interesting work turned out to be
in the two things measurement contradicted.

## The three engine bugs, fixed with tests

1. **`truncate_lines_at` never recorded loss.** It ran before `loss` was initialized (`dsl.go:192-200`
   vs `:202`), so a real intra-line cut reported `LossNone` and `cmdfilter` emitted **no recovery
   hint for a real loss**. Now types as `LossWhole` — an intra-line cut is non-contiguous by nature,
   since every long line loses its own tail.
2. **`truncate_lines_at` cut silently.** A mid-line cut with no marker reads as corrupted output to a
   model. Now appends `...`, sized to fit *inside* the cap so the line never grows.
3. **`recoveryHint` collapsed `LossTail` and `LossWhole`.** Both got the same "call
   `context_guru_expand`" text, making a cheap partial recovery look like an expensive whole-blob
   one and pushing the agent to the costly path. Now distinct: a tail cut names the cut point.

## The port: 24 filters, every selector rewritten

rtk matches a shell **command** (`^terraform\s+plan`); a proxy only ever sees the **output**. Every
selector is rewritten to an output-shape signature. Copied verbatim, rtk's regexes compile fine and
**never fire** — so `TestEveryBuiltinFilterHasTestsAndRoutes` asserts each filter's own test input
actually *routes to that filter*. That test is what makes the rewrite verifiable instead of hopeful,
and it caught every shadowing regression below.

Ported (22): `pytest` `npm-install` `make` `gradle` `xcodebuild` `gcc` `swift-build` `dotnet-build`
`turbo` `nx` `terraform-plan` `terraform-init` `pulumi` `liquibase` `ssh` `ping` `rsync`
`bundle-install` `poetry-install` `composer-install` `uv-sync` `brew-install` `quarto-render`.
Filters whose output is indistinguishable are merged (terraform/tofu; the five pulumi subcommands).

**Tier 3 not ported**, with reasons recorded in `docs/components/cmdfilter.md`: the ~24
width-cap-only filters, the 11 blank-line-only linter filters, `spring-boot`, `filter_stderr`, rtk's
86 command-detection rules, and its ~50 native Rust filters.

**Every success-collapse rule carries an `unless` guard** (rtk ships 9 of 11 unguarded), each with an
explicit negative test: a build with a warning **and** a success marker must not collapse. In a proxy
the agent cannot re-run the command to find what a bare collapse swallowed.
`TestEveryMatchOutputRuleIsGuarded` fails the build if one is added without a guard. `dotnet-build`
needed care: dotnet prints `0 Error(s)` on success, so a guard on the word "error" would never let it
collapse — it guards on the diagnostic *form* (`error CS1002`).

Also: shared `cap` budget classes by signal density instead of 24 hand-picked `max_lines`; a
500-byte size floor (rtk's `MIN_TEE_SIZE`); load-time guardrails (duplicate names rejected, inline
tests **executed at load**); a bounded selector-miss ledger; per-family metrics in `/stats`.

## Two things measurement contradicted

I replayed the shipped selectors over recorded proxy traffic before trusting the plan. Both findings
changed the code.

**1. The issue's Tier-1 priorities did not survive contact with the data.** On a Terminal-Bench dump,
`pulumi`, `terraform-plan`, `xcodebuild` and `gradle` fired **zero** times. What dominated were two
shapes rtk has no filter for: apt/dpkg boilerplate (584 outputs, ~1.0 MB — the largest reachable
family) and gcc's `<file>: In function 'main':` header. I added both; `apt` gets 76% compression and
went on to carry ~73% of all savings in the live run. The IaC filters stay — correct, tested, free
when inert — but the docs now say plainly that a filter set's value comes from the workload, not its
size. That is what the miss ledger is for.

**2. The one-line selector was a real bug in my own port.** 112 pytest runs (311 KB) matched nothing
because the agent harness prepends its own preamble (`Exit code 1`, `Internet access disabled`) or
the report opens with a bare `ERROR path::test` — so pytest's session banner is never line 1. A
one-line selector ties a filter's reach to the agent's **output framing** rather than to the tool
that produced the output. `selectorKey` now spans the first 6 non-empty lines, with `(?m)` so `^`/`$`
anchor per line:

| workload | claimed before | claimed after |
|---|--:|--:|
| SWE-bench captured traffic (520 eligible outputs) | 13 (2.5%) | **124 (23.8%)** |
| Terminal-Bench dump (223 eligible outputs) | 140 (62.8%) | **172 (77.1%)** |

Widening it made generic filters shadow specific ones — exactly the ordering hazard #29 flagged — and
the routing guardrail caught every case. Resolved with explicit `priority`: tool banners at 20,
`gcc` at **-10** (its selector is the generic "some compiler said something" shape that also occurs
inside make/swift/dotnet output, so it is the last resort, not the first match).

**A false positive caught in a live run:** `swift-build`'s `^Compiling ` claimed **Cython** output and
stripped its `Compiling x.pyx because it changed` lines — it fired 80× on a C build task with almost
no unique saving, visible directly in the new per-filter ledger. A generic verb is not a tool
signature. Fixed to require Swift identity, with a regression test asserting Cython and cargo output
are not claimed.

## Benchmarks

Terminal-Bench, build-heavy tasks, `off` baseline vs `codesmart`, matched per task. **Reward and
steps are the numbers that matter here** — this is the failure mode that sank rtk on Terminal-Bench.

| task | reward off → cs | steps off → cs | norm cost off → cs |
|---|---|---|---|
| `build-pmars` | 1.0 → 1.0 | 32 → 34 | $0.461 → $0.503 |
| `build-cython-ext` | 0.0 → **1.0** | 70 → **59** | $1.177 → $1.213 |
| `compile-compcert` | 1.0 → **0.0** | 55 → 58 | $1.602 → **$1.128** |
| **aggregate** | **2/3 = 2/3** | **52.3 → 50.3 (−3.8%)** | **$3.240 → $2.843 (−12.3%)** |

Cache-hit 95.97% → **97.43%**; cache-write **393k → 239k (−39%)** — the lever that actually mattered
here. Total including context-guru's own LLM cost: $2.862 (**−11.7%**). cmdfilter: 128 acts / 153 runs,
0 reverts, 0 expand bounces, ~2 ms/request.

**Reward is net neutral (2 = 2), not improved — and I am claiming neither flip.** Both went in
opposite directions, and in each case **the losing trial was an `exception: true` with `agent_wall_s`
pinned exactly at its budget** (1800.0 s and 4800.0 s). Both flips are **wall-clock timeouts, not
information loss.** The cython "win" and the compcert "loss" are the same artifact.

**The one real concern, and it is *not* the rtk failure mode.** compcert's codesmart trial burned +17%
agent wall (4089 s → 4800 s) and hit its budget. Steps went *down* overall and aggregate reward held,
so this is not the "agent lost information → more steps" spiral that sank rtk. It is added latency:
`cg_added_ms_avg` 5 ms → **145 ms**, upstream 12.9 s → 15.6 s/req — dominated by **`extract_llm`
(17.8 s across 2 calls, 0 tokens saved)**, not cmdfilter. On a task already near its wall clock that
was enough to push it over. That belongs to #28, which independently concluded `extract_llm` is
net-negative on caching backends; it is not a reason to back off these filters.

Live per-family ledger (the new `/stats` fields):

```
cmdfilter: 128 acts, 0 reverted
  pkg    (apt)          — the largest reachable family, ~73% of live savings
  tests  (pytest)
  builds (swift-build ← the Cython false positive, now fixed)
```

**Honest caveats, stated plainly:**

- **The SWE-bench control arm is unusable and I am not reporting it as a result.** Another agent's
  proxy already held the port I had been assigned, so my first attempt was measuring through
  *someone else's config*; after moving to a verified-free port, 2 of 3 codesmart trials aborted with
  `API Error: ConnectionRefused` from proxy restarts. One clean trial is not a control. What it *did*
  give me is the capture file that exposed the selector bug — the more valuable outcome.
- **n=1 per task.** All three rewards carry single-trial noise; with two of the three decided by a
  timeout, the reward column should be read as "no regression detected", not as a measurement.
- **This ran the PRE-FIX binary.** The selector widening (2.5% → 23.8% match rate on SWE traffic) and
  the Cython false-positive fix landed *after* this run, so the merged version should produce more
  output than these numbers show.
- **Unique savings are small in absolute terms** (~3.6k tokens). On a ~96% cached agent this is a
  rounding error against the bill, consistent with the existing finding that cost tracks steps, not
  unique tokens. The case for cmdfilter is that it is **free** — no LLM, ~2 ms, cache-safe — not that
  it moves the bill on its own.
- **The TB arm ran the pre-fix binary**, so its cmdfilter numbers *understate* the merged version
  (the selector fix and the Cython fix both landed after it started).

## Licensing

`THIRD-PARTY-NOTICES` (new) names rtk, `github.com/rtk-ai/rtk`, Apache-2.0, the
`Copyright 2024 rtk-ai and rtk-ai Labs` notice, and states that filter definitions and the DSL
pipeline are "adapted from, with modifications" — enumerating them. Apache-2.0 §4 applies to the
filter TOMLs as copyrightable expression; the existing prose credit in `dsl.go` is good practice but
not sufficient for a bulk port.

## Verification

`CGO_ENABLED=1 go build -tags cg_skeleton ./...` · `go test -tags cg_skeleton ./...` ·
`go test -race` · `make lint` · `gofmt -l` — all clean. `/stats` additions are **additive only**, with
a test asserting the keys `deploy/harbor/*.py` parses still exist.

Docs: `docs/components/cmdfilter.md` (full filter table, what each preserves, the Tier-3 rationale,
the two measurement findings), `docs/components/dsl.md` (loss typing, `cap` classes, `priority`),
`docs/how-to/custom-dsl-filter.md` (both selector traps), `THIRD-PARTY-NOTICES`.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -1,0 +1,54 @@
+THIRD-PARTY NOTICES
+===================
+
+context-guru includes third-party material. This file records the notices those
+works require. It covers material adapted INTO this source tree; Go module
+dependencies carry their own licenses in the module cache and in go.sum.
+
+-------------------------------------------------------------------------------
+rtk (Rust Token Killer)
+-------------------------------------------------------------------------------
+
+Project:   rtk
+Homepage:  https://github.com/rtk-ai/rtk
+License:   Apache License, Version 2.0
+Notice:    Copyright 2024 rtk-ai and rtk-ai Labs
+
+Portions of this software are adapted from rtk, with modifications:
+
+  - components/dsl (DSL filter pipeline). The declarative filter engine — the
+    eight-stage pipeline (strip_ansi, replace, match_output+unless, strip/keep
+    lines, truncate_lines_at, head/tail, max_lines, on_empty), the field names,
+    the Lossiness typing, and the shared truncation cap classes (CAP_ERRORS,
+    CAP_WARNINGS, CAP_LIST, CAP_INVENTORY) — is adapted from rtk's TOML filter
+    DSL (src/core/toml_filter.rs, src/core/truncate.rs) and re-implemented in Go.
+
+  - components/offload (cmdfilter filter definitions and test corpora). The
+    shipped filter set is adapted from rtk's src/filters/*.toml, including its
+    inline test cases.
+
+Modifications made:
+
+  - Re-implemented in Go, from Rust.
+  - Filter documents are YAML with a schema_version, not TOML.
+  - Every filter's selector is rewritten. rtk matches a shell COMMAND string
+    (match_command); context-guru is a proxy and never sees the command, so each
+    filter matches an output-shape signature against the first non-empty line of
+    the tool output instead. rtk's command regexes are not portable as written.
+  - Filters whose outputs are indistinguishable from the output alone are merged
+    (terraform/tofu plan; terraform/tofu init; the pulumi subcommands).
+  - Every success-collapse (match_output) rule carries an `unless` guard; most of
+    rtk's do not. A proxy's agent cannot re-run a command to discover a warning
+    that a bare collapse swallowed.
+  - Per-filter line budgets are replaced by shared `cap` budget classes, applying
+    rtk's truncate.rs cap idea to the filter definitions (rtk's own TOML filters
+    hand-pick a max_lines each).
+  - truncate_lines_at records intra-line loss and marks the cut with an ellipsis.
+  - A `buildlog` cap class, a `family` field (per-family metrics), a `priority`
+    field (selector ordering), load-time guardrails (duplicate names rejected,
+    inline tests executed at load), and a selector-miss ledger were added.
+  - rtk's filter_stderr, its command-detection rules, and its native Rust
+    (non-DSL) filters are not ported — a proxy has no analogue for them.
+
+The Apache-2.0 license text is available at http://www.apache.org/licenses/LICENSE-2.0
+and, since context-guru is itself Apache-2.0 licensed, in this repository's LICENSE.
--- a/components/component.go
+++ b/components/component.go
@@ -123,6 +123,22 @@ type Ctx struct {
 	// -1 = unknown/first turn/cache off ⇒ no tail restriction. Only meaningful when
 	// CacheAware is true.
 	MaxCachedIdx int
+	// FilterStats receives cmdfilter's per-filter ledger (which command families pay
+	// off, and which output shapes matched nothing). nil = not recording.
+	FilterStats FilterStatsSink
+}
+
+// FilterStatsSink records cmdfilter's per-filter/per-family ledger. metrics.Aggregator
+// implements it; the pipeline depends only on this interface. Implementations must be
+// safe for concurrent use.
+type FilterStatsSink interface {
+	// FilterAct notes one applied filter: its family (builds/tests/iac/pkg/net/...),
+	// its name, the content key (so a compaction re-sent verbatim next turn is counted
+	// once), and the tokens saved.
+	FilterAct(family, filter, contentKey string, saved int)
+	// FilterMiss notes a selector that matched no filter — the ledger that says which
+	// filter is worth writing next (after rtk's parse_failures table).
+	FilterMiss(selector string)
 }
 
 // TailOnly reports whether a supersession/age-based offloader may mutate the message
--- a/components/dsl/dsl.go
+++ b/components/dsl/dsl.go
@@ -155,7 +155,11 @@ func Compile(name string, d Def) (*Compiled, error) {
 	} else if d.CapReduce != 0 {
 		return nil, fmt.Errorf("dsl: filter %q sets cap_reduce without cap", name)
 	}
-	m, err := regexp.Compile(d.Match)
+	// The selector spans a few leading lines, not one, so `^`/`$` in a match regex must
+	// mean "start/end of A line" rather than "of the whole selector". Without (?m) a
+	// filter anchored at ^ only matches output whose very FIRST line is its signature,
+	// which is exactly the output-framing dependence a multi-line selector removes.
+	m, err := regexp.Compile("(?m)" + d.Match)
 	if err != nil {
 		return nil, fmt.Errorf("dsl: filter %q match: %w", name, err)
 	}
--- a/components/dsl/dsl.go
+++ b/components/dsl/dsl.go
@@ -252,7 +252,9 @@ func Apply(c *Compiled, input string) (string, Lossiness) {
 	if c.def.HeadLines != nil || c.def.TailLines != nil {
 		var htLoss Lossiness
 		lines, htLoss = headTail(lines, c.def.HeadLines, c.def.TailLines)
-		loss = worseLoss(loss, htLoss)
+		if htLoss > loss { // keep the more severe of an intra-line cut and a line drop
+			loss = htLoss
+		}
 	}
 	// 7. max_lines (absolute cap, counts the omission marker)
 	if c.def.MaxLines != nil && len(lines) > *c.def.MaxLines {
@@ -285,14 +287,6 @@ func truncateRunes(s string, n int) string {
 		return "..."
 	}
 	return string(r[:n-3]) + "..."
-}
-
-// worseLoss returns the more severe of two loss kinds.
-func worseLoss(a, b Lossiness) Lossiness {
-	if b > a {
-		return b
-	}
-	return a
 }
 
 func filterLines(lines []string, res []*regexp.Regexp, keep bool) []string {
--- a/components/dsl/dsl.go
+++ b/components/dsl/dsl.go
@@ -36,10 +36,42 @@ const (
 	LossWhole                  // non-contiguous / whole-blob loss; only full retrieval recovers
 )
 
+// Caps are the shared line budgets a filter picks by SIGNAL DENSITY (`cap: errors`)
+// instead of hand-picking a max_lines per filter. The first four names and values
+// are rtk's (src/core/truncate.rs); `buildlog` is ours — a build/plan transcript is
+// mostly noise but the line that matters can sit anywhere in it, so it gets a
+// deliberately generous budget. One map tunes the whole filter set.
+var Caps = map[string]int{
+	"errors":    20, // most actionable, shown the most
+	"warnings":  10, // lower signal density than errors
+	"list":      20, // flat lists (packages, services): one line per item
+	"inventory": 50, // exhaustive lookups (installed packages, file listings)
+	"buildlog":  80, // full build/plan transcripts: verbose, signal is positional
+}
+
+// ReducedCap is rtk's `reduced` deviation helper: a cap lowered for a more verbose
+// data class, underflow-safe — a deviation can never empty the budget.
+func ReducedCap(cap, by int) int {
+	if by > 0 && by < cap {
+		return cap - by
+	}
+	return cap
+}
+
 // Def is a raw filter definition (from YAML). All fields except Match are optional.
 type Def struct {
-	Description        string        `yaml:"description"`
-	Match              string        `yaml:"match"` // regex against the selector key
+	Description string `yaml:"description"`
+	Match       string `yaml:"match"` // regex against the selector key
+	// Family groups filters for per-family metrics (builds, tests, iac, pkg, net, ...).
+	Family string `yaml:"family"`
+	// Priority orders matching: higher first, then by name. Absent (0) = today's
+	// behavior (name order). Use it to put a specific filter ahead of a generic one,
+	// which matters more here than in rtk because we match on output shape.
+	Priority int `yaml:"priority"`
+	// Cap selects a shared budget class from Caps instead of a literal MaxLines;
+	// CapReduce lowers it for an extra-verbose variant. MaxLines wins if both are set.
+	Cap                string        `yaml:"cap"`
+	CapReduce          int           `yaml:"cap_reduce"`
 	StripANSI          bool          `yaml:"strip_ansi"`
 	Replace            []ReplaceRule `yaml:"replace"`
 	MatchOutput        []MatchRule   `yaml:"match_output"`
@@ -110,6 +142,18 @@ func Compile(name string, d Def) (*Compiled, error) {
 	}
 	if len(d.StripLinesMatching) > 0 && len(d.KeepLinesMatching) > 0 {
 		return nil, fmt.Errorf("dsl: filter %q sets both strip_lines_matching and keep_lines_matching", name)
+	}
+	if d.Cap != "" {
+		base, ok := Caps[d.Cap]
+		if !ok {
+			return nil, fmt.Errorf("dsl: filter %q unknown cap %q", name, d.Cap)
+		}
+		if d.MaxLines == nil { // an explicit max_lines still wins
+			n := ReducedCap(base, d.CapReduce)
+			d.MaxLines = &n
+		}
+	} else if d.CapReduce != 0 {
+		return nil, fmt.Errorf("dsl: filter %q sets cap_reduce without cap", name)
 	}
 	m, err := regexp.Compile(d.Match)
 	if err != nil {
@@ -189,20 +233,26 @@ func Apply(c *Compiled, input string) (string, Lossiness) {
 	} else if len(c.keepLines) > 0 {
 		lines = filterLines(lines, c.keepLines, true)
 	}
-	// 5. truncate_lines_at (unicode-safe per-line cap)
+	// 5. truncate_lines_at (unicode-safe per-line cap). An intra-line cut is a REAL
+	// loss and is non-contiguous by nature (every long line loses its own tail), so
+	// it types as LossWhole; and it appends an ellipsis, because a silent mid-line
+	// cut reads as corrupted output to a model (both after rtk).
+	loss := LossNone
 	if c.def.TruncateLinesAt != nil {
 		n := *c.def.TruncateLinesAt
 		for i, l := range lines {
-			if r := []rune(l); len(r) > n {
-				lines[i] = string(r[:n])
+			if t := truncateRunes(l, n); t != l {
+				lines[i] = t
+				loss = LossWhole
 			}
 		}
 	}
 
-	loss := LossNone
 	// 6. head/tail
 	if c.def.HeadLines != nil || c.def.TailLines != nil {
-		lines, loss = headTail(lines, c.def.HeadLines, c.def.TailLines)
+		var htLoss Lossiness
+		lines, htLoss = headTail(lines, c.def.HeadLines, c.def.TailLines)
+		loss = worseLoss(loss, htLoss)
 	}
 	// 7. max_lines (absolute cap, counts the omission marker)
 	if c.def.MaxLines != nil && len(lines) > *c.def.MaxLines {
@@ -212,7 +262,7 @@ func Apply(c *Compiled, input string) (string, Lossiness) {
 		if loss == LossNone {
 			loss = LossTail
 		} else {
-			loss = LossWhole
+			loss = LossWhole // already lossy above: the cap is no longer a clean tail cut
 		}
 	}
 	out := strings.Join(lines, "\n")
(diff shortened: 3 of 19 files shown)


OsherElhadad:
## Orchestrator verification — the benchmark reading is correct, and the caveats are real

I re-derived everything from `/tmp/i29-runs/` row files rather than taking the table on trust.

### Both reward flips confirmed as wall-clock timeouts, not information loss

This is the load-bearing claim, and it checks out exactly as stated:

| task | reward | steps | exception | agent_wall_s |
|---|---|---|---|---|
| `build-cython-ext` | 0.0 → **1.0** | 70 → 59 | **True → False** | **1800.0** → 1386.1 |
| `compile-compcert` | 1.0 → **0.0** | 55 → 58 | **False → True** | 4089.4 → **4800.0** |
| `build-pmars` | 1.0 → 1.0 | 32 → 34 | False → False | 776.7 → 747.4 |

Each losing trial is an `exception: true` with `agent_wall_s` pinned *exactly* at its budget (1800.0 and 4800.0). Declining to claim either flip is the right call — they are the same artifact pointing opposite directions, and treating the cython "win" as a gain while explaining away the compcert "loss" would have been the tempting and wrong move.

### Aggregate figures reproduce

| | off | codesmart |
|---|--:|--:|
| billed cost | $3.240 | **$2.843** (−12.3%) |
| cache-read | 9,379,052 | 9,183,987 |
| **cache-write** | **393,402** | **239,210** (−39.2%) |
| cache-hit | 95.97% | **97.43%** |

Cache-write −39% is the result that matters on this benchmark, and it is the lever the two-benchmark study identified as decisive on long-horizon contexts. Worth noting it lands in the same direction and similar magnitude as #35's async arms (−45% SWE / −39% TB), from a completely different mechanism.

### The SWE control is worse than "unusable" — confirm it is excluded

```
astropy-12907  reward 0.0 steps 2  exception True   →  None (exception True)
astropy-14365  reward 0.0 steps 2  exception True   →  1.0 (34 steps)
astropy-8707   reward 0.0 steps 2  exception True   →  None (exception True)
```

The `off` arm is **$0.000 total, zero cache-read, all three trials 2-step aborts**. There is nothing there. The one `codesmart` trial that ran has no counterpart. Correctly reported as not-a-result; I confirmed no delta is quoted from it anywhere in the PR body.

### On the `extract_llm` attribution

The claim that compcert's +17% wall came from `extract_llm` (17.8 s across 2 calls, 0 tokens saved) rather than cmdfilter is consistent with two independent findings: **#34** concluded the component is net-negative on caching backends and recommends disabling it there, and **#35** measured it making *zero* calls on Terminal-Bench. Three separate workstreams converging on the same component is worth more than any one of them.

That said, this does mean **the merged `codesmart` preset would still carry the latency that pushed compcert over its budget.** That is a preset-composition decision for the merge, not a defect in this PR — but it should not be lost: shipping #37 and #34 together, and acting on #34's recommendation, is what actually removes the observed harm.

### What to fix before merge

The caveats are all stated in the PR body, which I've verified. The one thing I'd add: make it explicit that **this run used the pre-fix binary**, so the selector widening (2.5% → 23.8% SWE match rate) and the Cython false-positive fix are *not* reflected — the merged version should produce more output than these numbers, and a future reader comparing against a re-run needs to know that.

Verdict from my side: reward is net neutral with no regression detected, the cost and cache-write improvements are real, and the methodology disclosure is honest. The `apt` filter carrying ~73% of savings makes it the highest-stakes one — I've asked the independent reviewer to hunt for more cases of the `^debconf: ` class of over-broad pattern that your own safety test already caught once.

OsherElhadad:
Superseded by #42 — same 7 commits rebased onto main after #33/#36/#40 merged. Could not rebase in place because branch protection blocks force-push to this branch.

The `components/component.go` conflict needed a **manual** merge, not the keep-both resolution that works for `metrics/metrics.go`: #36 adds an `ExistingBreakpoints` field to `Ctx`, while this branch adds a `FilterStats` field *and* a `FilterStatsSink` interface immediately after the struct. Concatenating the two sides drops the struct field into the interface body and yields invalid Go. Resolved with both fields inside `Ctx` and the interface separate; full gates re-run green.

PR #38, opened 2026-08-10
feat(dash): first-class persistent observability dashboard

### LINKED ISSUE #30: feat(dashboard): first-class persistent observability UI — what value is context-guru providing?
## Problem statement

context-guru has no dashboard. Its entire observability surface is `GET /stats` (`proxy/proxy.go:538-549`) returning a single JSON snapshot of an **in-memory** `Aggregator` (`metrics/metrics.go:70` — one mutex-guarded struct, ~267 lines). Nothing is persisted; restart the proxy and every number is gone. There is no HTML, no JS, no embedded assets anywhere in the repo (verified: zero `go:embed` directives).

A user therefore cannot answer the question the product exists to answer: **what value is context-guru providing?**

## Motivation

Two benchmarks (`docs/results/comparison.md`, `docs/results/terminal-bench-comparison.md`) established that this workload's economics are counterintuitive: the request is ~99.95% cached, cache-write costs **11.5×** a cache-read, unique token removal is **0.02–0.13%** of the billed total, and cost tracks agent **steps** at `r = 0.95`. A tool whose savings are reported as "tokens removed" is therefore reporting the least important number — and, per `docs/results/improvement-plan.md`, we are currently **blind on our largest cost line**. Every other issue in this batch needs somewhere to show its numbers.

Both reference implementations were studied. Their lessons are concrete and are folded into the requirements below:

- **headroom** has a genuinely strong dashboard (2,755-line single HTML, Alpine + Tailwind + htmx from CDN, 5 s polling, three view modes). Its best ideas: four *different* savings denominators each with a stated justification; cache-miss attribution bucketed `ttl_expiry | prefix_change | unknown | cold_start` with TTL winning ties; **reporting the cost of its own safety mechanisms alongside their benefit** (`compression_vs_cache`, `prefix_freeze`) — which is what makes its numbers credible; `token_accounting_status: complete|partial|missing` per row; model-aware bust detection (first request per model is a cold start, never a bust); `uncompressed_requests` reason buckets answering "why didn't you compress this?"; a `?cached=1` 5 s snapshot with the request tail refreshed on top; prose `description` strings inside the payload stating each layer's scope; bounded persistent state with percentages derived at read time; cost stored at write time so history does not reprice.
  Its weaknesses to beat: CDN assets (breaks offline, in a product with an air-gapped install story); the richest attribution data (`compressions_by_strategy`, `tokens_saved_by_strategy`) computed, serialized, and **rendered nowhere**; no histograms in `/metrics` so p95 is uncomputable; **no search, filter, or pagination anywhere**; no session list at all despite sessions existing internally; per-request history is a 10k in-memory deque; three tabs with three different definitions of "saved"; and its headline metric contradicts its own design note.
- **gateway** is *not* a reference for features — it has no database, no charts, no filters, no session concept, and its 100-request in-memory ring is lost on restart. It **is** a near-perfect reference for the *delivery pattern*: one static HTML file read at boot and returned from one route, plus SSE for live updates, with no frontend build step at all. Also worth porting: allowlist-based redaction (blanket-redact all headers by key, allowlist config keys), out-of-band capture so logging can never add latency or fail a request, per-attempt rows, and a write timeout + dead-client eviction on the SSE fan-out. Its one genuinely missing view is the one we most need: it carries `finalUntransformedRequest.body` and `transformedRequest.body` side by side and never diffs them.

## Desired behavior

A production-quality dashboard that makes the optimization **understandable**, not just numeric. Within seconds a user should see: tokens saved, dollars saved, gross vs honest/net, cost without context-guru vs with, which components produced the savings, what overhead we added, whether compaction caused restoration/expand events, whether cache behavior improved or regressed, and — where benchmark data exists — whether quality/reward changed.

### Overall metrics
total requests · total sessions · tokens before · tokens after · gross saved · unique saved · adjusted/net saved · savings % · estimated dollars saved · baseline estimated cost · actual estimated cost · context-guru's own LLM cost · expand/restoration tokens · bounce/expand count · expand rate · reverts · passthroughs · average context-guru latency · upstream latency · cache-read / cache-write / fresh-input / output tokens.

Several of these already exist in `metrics/metrics.go:198-221` (`tokens_before`, `saved_tokens`, `saved_tokens_unique`, `overcount_ratio`, `wasted_tokens`, `bounces`, `adjusted_saved`, `cg_added_ms_avg`, `upstream_ms_avg`, `llm_*`); the rest need adding.

### Time-series graphs
token savings · dollar savings · baseline vs context-guru cost · context size over conversation turns · cache reads · cache writes · context-guru latency · upstream latency · expand/restoration rate · request volume.

**The important one:** cumulative cost **without** context-guru vs **with**, the area between them being money saved.

### Honest savings waterfall
baseline cost → deterministic savings → LLM-based savings → cache savings → restoration penalty → context-guru's own LLM cost → final net cost → net savings.

### Per-component observability
executions · acted · mutated · skipped · reverted · gross tokens saved · unique tokens saved · adjusted savings · average and total latency · effectiveness · restoration behavior where attributable · economic value · cost of running the component. It must be easy to see **which components earn their place** — on current evidence `cacheinject` is inert, `extract_llm` is ~8× underwater, and `failed_run` never fires on the benchmark config. The dashboard should make that obvious without reading a doc.

### Sessions
Searchable/filterable session list, which today does not exist in any form. Per session: id · model/provider · application/agent · preset/config · start/end · turns · tokens before/after/saved · dollars saved · cache reads/writes · expands · context-guru latency. Clicking a session shows its requests/turns.

### Request/turn details
original size · compacted size · savings · which components ran · exactly when each activated · component timings · restores/expands · configuration in effect · cache behavior · before/after content where appropriate.

### Diffs — the headline feature
Useful diffs for: original tool output/context, the transformed/compacted version, and restored content when expansion happens. Git-style rendering for code-oriented transformations. It must be easy to inspect: **what did context-guru actually remove or rewrite?** This is the view both reference implementations have the data for and neither built.

### Configuration visibility
active preset · full **effective** configuration (resolved, not as typed) · enabled components · thresholds · model used by LLM components · cache policy · operating mode · changes over time where possible. Ideally compare configurations or sessions.

### Benchmark views
Ingest and display benchmark results: baseline · context-guru · headroom · rtk, with reward · billed cost · token categories · steps · latency · tool cost, including cost-vs-reward visualizations. The `rows-*.json` + `summary.json` produced by `deploy/harbor/*.py` are the natural input format.

### Search and filtering
By time range · session · model · provider · agent/application · preset · component · mode · benchmark · task · repository/project where available.

### UX quality
Not a Prometheus/Grafana clone — a polished product. Study both references and improve on them.

## Relevant code locations

- `proxy/proxy.go:110-142` — the route table (`/healthz`, `/stats`, `/expand`, `/compact`, the chat handlers); where new routes attach.
- `proxy/proxy.go:538-549` — the `/stats` handler; `:547` merges `cheapmodel.Usage()` at serve time.
- `proxy/proxy.go:355-383` — where compaction runs and `RecordAddedLatency` fires; the natural capture point.
- `proxy/proxy.go:421-500` — upstream round-trip, `RecordUpstreamLatency`, the expand loop and `RecordExpand`.
- `metrics/metrics.go:70` (single mutex), `:88-100` (per-component stats incl. `saved_tokens_unique`, `overcount_ratio`, `duration_ms`), `:196-221` (`Snapshot` shape).
- `components/pipeline.go` — per-component run/revert accounting; the source of "which components ran and when".
- `apply/apply.go:115-126` (fail-open backstop), `:152-173` (`Ctx` construction, session id, `CacheAware`, `MaxCachedIdx`) — where per-request facts are known.
- `config/config.go` — presets, for the effective-config view.
- `deploy/harbor/*.py` + `docs/results/*.md` — benchmark data to ingest.
- `session/` — session resolution.

## Proposed architecture

Grounded in what the references actually do, and deliberately boring.

1. **Storage: SQLite via a pure-Go driver** (`modernc.org/sqlite`, no CGO beyond what tree-sitter already forces). headroom's total lack of durable per-request history is its defining limitation; do not copy it. Two tables:
   - `requests`: id, session_id, ts (epoch ms — **not** a locale string, which is gateway's mistake), model, provider, agent, preset, mode, status, tokens_before, tokens_after, fresh/cache_read/cache_write/output, cost_usd, baseline_cost_usd, cg_llm_cost_usd, cg_latency_ms, upstream_ms, expands, reverts, token_accounting_status.
   - `request_components`: request_id, component, ran, acted, mutated, reverted, saved_gross, saved_unique, duration_ms.
   - Optional `request_content`: request_id, before, after (compressed, size-capped, opt-out flag) — required for the diff view.
   Indexes on `(ts DESC)`, `(session_id, ts)`. Retention by age **and** size. **No rollup tables** — bucket at query time (`ts/:bucket*:bucket GROUP BY 1`); SQLite handles millions of rows and pre-aggregation is the speculative complexity to skip until a query is measurably slow.
2. **Capture off the hot path.** The handler hands a struct to a buffered channel (drop + count when full); one writer goroutine batches inserts in a transaction and fans the summary to SSE clients. This preserves gateway's best property — logging can never add latency or fail a request — while adding durability it lacks. The existing `Aggregator` stays as the fast in-process counter.
3. **Frontend: one embedded HTML file** via `go:embed`, no npm, no build step (gateway's delivery model), but with **CSS custom properties and a dark-mode block from line one** (headroom had to bolt on ~25 light-mode overrides because it hardcoded hex). **Vendor every asset into the embed FS** — no CDN, since context-guru ships into VPC/air-gapped contexts. A small no-dependency chart library embedded as bytes (uPlot-class, ~45 KB) covers the time series.
4. **API**: history **and** live, not live-only. `GET /api/requests` with server-side filters + keyset pagination; `GET /api/requests/{id}` returning the request plus its component rows and before/after; `GET /api/sessions`; `GET /api/stats?bucket=&since=`; `GET /api/config` (effective, redacted); `GET /api/benchmarks`; `GET /api/events` (SSE, summary rows only, honoring `Last-Event-ID` so a reconnect backfills instead of silently losing the gap).
5. **Redaction before the DB, not on read.** Allowlist config keys, blanket-redact headers by key. Secrets must never reach disk. Size-cap content capture and make it opt-out.
6. **Access gating** (headroom's `_request_can_view_dashboard_metadata`): per-request content and effective config for loopback or an explicit trusted-CIDR allowlist; aggregates open. Cheap, and the right default for a proxy people bind to `0.0.0.0`.

## Metric-semantics requirements (non-negotiable)

- **Report multiple denominators, each labeled with what it divides by.** A whole-request ratio recounts the transcript every turn, so a 200-turn session reads ~0%. Ship at least: saved / attempted-to-compress, and saved / provider-billed-new-input (headroom's `new_input_savings_percent` — the cleverest number in either tool), guarded so a provider with no cache usage data cannot divide savings by themselves and report ~100%.
- **Report the cost of our own safety mechanisms next to their benefit** — tokens saved by compaction vs tokens lost to cache busts, and freeze/tail-only's foregone compaction. A compaction proxy that only shows tokens removed is unfalsifiable.
- **Never count fake savings.** Gross, unique, and adjusted must be visibly distinct; `overcount_ratio` already exists and should be surfaced, not hidden.
- **Cache-miss attribution** with a cold-start bucket that is not a failure.
- **`token_accounting_status` per row** — never render a partially-instrumented request as exact.
- **"Why didn't you compact this?"** as a first-class reason bucket.

## Alternatives considered

- **Prometheus + Grafana only.** Explicitly rejected by the requirement. Also cannot show diffs or per-request detail.
- **Keep in-memory, add an HTML view.** Cheapest, but a restart erases everything and no historical analysis is possible — the same limitation that makes headroom's Session tab a footnote.
- **A real SPA with a build step.** Rejected: toolchain rot in a Go binary's release artifact, and gateway proves a single embedded file is sufficient at this scale.
- **Write to JSONL instead of SQLite.** Simpler, but filtering and pagination then move into the client, which is exactly the gap in headroom.

New dependency (`modernc.org/sqlite`) and a new on-disk artifact. Needs: a documented default path, a schema version with a discard-and-preserve path for a mismatch (headroom's approach), bounded growth, and a no-op/in-memory mode so the proxy still runs where the path is unwritable. Every percentage derived at read time; **cost stored at write time** so history does not reprice when a model's rate changes.

## Backward compatibility

`/stats` must keep its current shape — the benchmark harnesses (`deploy/harbor/*.py`) parse it, and breaking it invalidates the reproduction path. Add fields; do not rename or remove. The dashboard is additive; persistence must be disableable.

## Configuration design

`--dashboard` / env to enable, a DB path, a retention setting, a content-capture toggle, and a trusted-CIDR list. Sensible defaults; **do not** copy headroom's "disable observability in production" gate — that is backwards for a tool whose value *is* observability.

## Testing plan

- Unit: schema migration, retention pruning, keyset pagination, every filter dimension, query-time bucketing.
- Unit: redaction — a known secret in a header/config never appears in a stored row.
- Unit: the capture channel drops rather than blocks when full, and the drop is counted.
- Unit: SSE fan-out with a hung client (write timeout + eviction).
- Unit: `/stats` shape unchanged (golden test against the current payload).
- `go test -race` on the capture path and the SSE hub — mandatory.
- Playwright: see below.

## Real-world benchmark plan

This is dashboard work, so the requirement is **real runtime data**, not fixtures. Run 2–3 SWE-bench and 2–3 Terminal-Bench tasks plus a real Claude Code session through the proxy with the dashboard on, then validate against that populated data. Also ingest a full existing benchmark run (`rows-*.json` + `summary.json`) and verify per-task drill-down. Record: commit SHA, config, model, task ids, all four token tiers, per-component cost, total billed cost, steps, reward, latency, expands, reverts, savings. Measure and report the **capture overhead** — if the dashboard costs measurable request latency, that is a regression in a tool that sells latency awareness.

## UI verification (Phase 5)

Playwright/browser automation against the actual rendered app, with screenshots of: Overview · savings/cost graph · component metrics · sessions · request detail · content/Git diff · configuration · observe-mode visualization · benchmark comparison · filters/search. Check desktop and a smaller viewport, empty states, loading states, large datasets, long session ids and long content, charts, tooltips, filtering, navigation. Follow headroom's one clearly right practice here: `data-testid` on every stat tile, asserted in tests, so the visual layer is regression-tested.

## Acceptance criteria

- [ ] Persistent storage with schema versioning, bounded retention, and a working no-persistence mode.
- [ ] Capture is off the hot path; measured added latency reported; drops counted.
- [ ] Secrets never reach disk; redaction unit-tested.
- [ ] Dashboard served from embedded assets with **no network fetches** (verifiable offline).
- [ ] Every metric group above present; multiple labeled savings denominators; safety-mechanism costs shown beside benefits; `overcount_ratio` and `token_accounting_status` surfaced.
- [ ] Baseline-vs-actual cumulative cost chart with the saved area.
- [ ] Honest savings waterfall.
- [ ] Per-component economics view making it obvious which components earn their place.
- [ ] Session list + drill-down to requests; request detail with component timings and before/after diff.
- [ ] Effective-configuration view.
- [ ] Benchmark ingestion + cost-vs-reward view for all four arms.
- [ ] Server-side search/filter across every listed dimension, with pagination.
- [ ] Dark mode from the start; empty/loading/error states; small-viewport behavior.
- [ ] Playwright screenshots in the PR and in the docs.
- [ ] `/stats` shape unchanged (golden test).
- [ ] `go test -race` green.

## Documentation updates

New `docs/dashboard.md` (with screenshots), `docs/how-to/measure-savings.md` rewritten around the dashboard, `docs/reference/routes.md` (new API routes), `docs/reference/config.md`, `docs/design.md` (the observability/event/store architecture), `mkdocs.yml` nav, and a README section. The GitHub Pages site must build and render.

## Dependencies

Consumes metrics from **#25** (frozen hit/miss/flip), **#26** (restoration activity), **#27** (`xdedup` attribution), **#28** (extraction economics), **#29** (per-family `cmdfilter` savings), and must render **#31**'s three modes with unmistakably distinct semantics. Land the storage/event architecture early so the others can emit into it; the per-issue panels can follow. This is the **largest** issue in the batch — consider splitting the PR into storage+API, then UI, then benchmark ingestion.

OsherElhadad:
Closes #30

context-guru had no dashboard: the entire observability surface was `GET /stats`, one
in-memory `Aggregator` snapshot that vanished on restart. A user could not answer the
question the product exists to answer — **what value is context-guru providing?**

This adds a persistent dashboard: durable per-request storage, an off-hot-path capture
pipeline, a filterable JSON/SSE API, and an embedded single-page UI. `--dashboard` turns it
on; without the flag the route table is byte-identical to before and `/stats` is unchanged.

Overview

## What is here

**Storage** — SQLite via `modernc.org/sqlite` (pure Go, no extra C toolchain), WAL mode.
`requests` · `request_components` · `request_content` (gzip, size-capped, skippable) ·
`bench_runs`/`bench_tasks`. Epoch-**ms** timestamps throughout; a formatted locale string
cannot be range-queried, sorted portably, or bucketed. Retention bounded by age **and**
size. Schema versioned, with a mismatch renaming the old file aside rather than refusing to
boot. `--dashboard-db :memory:` for no-persistence, which is also the automatic fallback
when the path is unwritable.

**Capture off the hot path** — the handler builds one struct from values the request path
already computed and does a channel send with a `default:` branch. A full queue **drops and
counts**; the drop count is surfaced in the UI, because a dashboard that hides its own
coverage gaps cannot be trusted about anything else. One writer goroutine batches inserts
and fans summaries to SSE.

**API** — `/api/stats`, `/api/series?bucket=`, `/api/requests` (server-side filters +
**keyset** pagination), `/api/requests/{id}`, `/api/sessions`, `/api/components`,
`/api/facets`, `/api/config`, `/api/benchmarks[/{id}/tasks]`, `/api/capture`, `/api/events`
(SSE honoring `Last-Event-ID`). No rollup tables — series are bucketed in SQL at query time.

**UI** — three files via `go:embed`. No npm, no bundler, no framework, and **no CDN**:
context-guru ships into VPC and air-gapped contexts, so a test fails on any external origin
and the assets are served under `default-src 'self'`. Charts are hand-drawn SVG. Design
tokens are CSS custom properties from line one, so dark mode redefines only tokens.

## Metric semantics — the actual point

| Requirement | How |
|---|---|
| Multiple labelled denominators | Four ratios, each shipping its divisor **and a prose description in the payload**: of-attempted · of-new-provider-billed-input · of-whole-request (diluted) · unique-of-whole |
| Never divide savings by themselves | The new-input ratio is guarded on the **billed** figure; with no usage data it reports **n/a**, not ~100% |
| Never count fake savings | gross · unique · adjusted are separate columns; `overcount_ratio` surfaced |
| Report the cost of our own safety | Frozen-for-cache-safety tokens, restorations, reverts, and context-guru's own latency and LLM spend, each beside its benefit |
| `token_accounting` per row | `complete\|partial\|missing`. An unpriceable request reads **unknown** — never zero, never exact |
| Cache-miss attribution | `hit · cold_start · ttl_expiry · prefix_change · unknown`; cold start is **not** a failure, TTL wins ties |
| "Why didn't you compact this?" | First-class reason bucket: `bypassed · below_trigger · cache_frozen · found_nothing · reverted · no_messages` |

Cost is computed at **write** time so history does not reprice; percentages at **read** time
so a filter change needs no rebuild.

## The headline feature: the before/after diff

Both reference implementations carry this data and neither renders it.

Git-style content diff

LCS diff with common head/tail trimmed first, Git-style hunks with line numbers and
collapsed unchanged runs, plus side-by-side and after-only modes. Blocks are ordered
biggest-saving first and the largest opens — leading with an unchanged block buries the
answer.

## Per-component economics

Per-component economics

On the real traffic above the verdicts do their job: `extract` is *earning its place*,
`cacheinject` *mutates, saves no content* (its win is provider-side), and a component that
burned wall time for nothing reads *costly and inert*.

## Benchmarks

Benchmark comparison

Ingested straight from the harness artifacts (`summary.json` + `rows-<arm>.json`) — no new
export format, so **42 historical runs from `/tmp/tb-runs` and `/tmp/cg-runs` ingested on
first boot** (652 task rows). Per arm: solve rate, mean reward, total/mean cost, **cost per
solve**, cache hit rate, exceptions, with a cost-vs-reward scatter and per-task drill-down.
Re-ingesting replaces a run rather than duplicating it.

## Sessions, requests, config, dark mode

Sessions
Request detail
Effective configuration
Dark mode

## Measured capture overhead

The issue asks for this explicitly, since a dashboard that costs request latency is a
regression in a tool that sells latency awareness.

```
BenchmarkRecord-16              200000        7567 ns/op   1141 B/op   7 allocs/op
BenchmarkRecordFullQueue-16     200000         388 ns/op    399 B/op   1 allocs/op
BenchmarkObserve-16             200000          64.8 ns/op
TestCaptureOverheadIsNegligible: capture overhead 175 ns per request (20,000 events)
```

**~175 ns per request** on the request goroutine — against a measured 8.4 s mean upstream
round trip on this workload, that is ~0.000002%. `BenchmarkRecord`'s 7.5 µs includes the
benchmark's own `&Event{}` allocation, which the handler does regardless. The full-queue
(drop) path is 388 ns, so an overloaded dashboard degrades instead of becoming a latency
incident. A regression test fails above 50 µs, so moving I/O onto the capture path breaks CI.

## Security

Redaction happens **before the database, never on read** — a secret on disk is a secret
forever, and a redact-on-read filter is one forgotten code path from leaking it. Headers are
blanket-redacted by key against a short allowlist (a denylist fails the moment a gateway
invents a new auth header); config keys are allowlisted with credential-named keys always
withheld; content is pattern-scrubbed **then** capped, so a secret at the truncation
boundary cannot survive by being cut into an unmatchable prefix.

Per-request content and the effective config are loopback/trusted-CIDR only; aggregates stay
open, because a proxy bound to `0.0.0.0` should still report its own numbers. There is
deliberately **no** "disable observability in production" switch.

## Real-data validation (not fixtures)

Everything above is rendered from real traffic through this proxy on port 4107:

- **40 real requests** replayed from captured SWE-bench and Terminal-Bench `claude-code`
  transcripts, against the live provider — so all four billed token tiers, the prompt-cache
  structure, cold starts and TTL expiries are real, not synthesised.
- A real `claude` CLI session through the proxy.
- **42 benchmark runs / 652 task rows** ingested from the existing jobs roots, with per-task
  drill-down verified.
- Data survived a proxy restart (persistence verified end to end).

Observed: 4 sessions · 120k → 106k content tokens · 14k gross / 2.0k unique saved
(overcount **7.0×**) · 328k cache reads vs 144k writes · $0.46 baseline vs $0.44 actual ·
accounting 32 `complete` / 9 `partial` · cache misses attributed 26 hit / 8 unknown /
4 cold_start / 3 ttl_expiry.

## Tests

- `dash`: schema migration + version-mismatch preservation, retention by age **and** size
  (with cascade), keyset pagination covering every row exactly once, **every** filter
  dimension, query-time bucketing, redaction (incl. a canary written through the real
  capture path and read back from every column), the capture channel dropping rather than
  blocking with the drop counted, SSE fan-out with a **hung client** evicted, `Last-Event-ID`
  backfill, cache attribution, benchmark ingest from real harness-shaped artifacts.
- `proxy`: **golden `/stats` shape** in both directions (a lost field fails; a new field
  fails until recorded), usage parsing in both dialects, the streaming sniffer, and
  end-to-end capture through a real pipeline.
- `go test -race ./...` — **20/20 packages green**, including the capture path and SSE hub.
- `make lint` clean; `mkdocs build --strict` clean.

**Browser verification: 124 checks, 0 failures** against the live app — every stat tile's
`data-testid`, chart rendering and tooltips, filters, pagination, the drawer, all three diff
modes, benchmark drill-down, dark/light, a 390 px viewport, and empty states. Every
screenshot was scanned for credentials byte-wise and by OCR before being committed.

390px viewport
Empty state

## Bugs found and fixed along the way

Four real defects, each with a regression test:

1. **`internal/modelinfo` never resolved anything in production.** The LiteLLM prices map
   was decoded into one typed `map[string]struct{…}`, which fails on the document's
   prose-filled `sample_spec` entry and on models that spell an integer as a float —
   yielding **nothing**, so every context-window lookup silently returned "unknown" and
   fraction-based triggers have never fired. Per-entry decode keeps the ~2,900 good rows;
   the resolved rates now match the harness's published price vector exactly. *(Pre-existing,
   unrelated to the dashboard.)*
2. **Streaming output tokens were under-reported.** An output-only usage block — Anthropic's
   `message_delta`, which carries the *final* completion count — parsed as "no usage".
3. **The streaming sniffer dropped its tail window** whenever both halves saturated, losing
   usage for every large streamed response.
4. **A dropped paren in `app.js` rendered a blank dashboard.** Go compiled, every Go test
   passed, the HTML served 200. There is now a test that parses `app.js` (node when present,
   a bracket-balance scan otherwise).

Plus two redaction over-reaches caught by the browser checks: redacting the `components`
subtree wholesale made the config view show nothing, and a substring match on `token` also
swallowed `max_tokens`/`min_tokens`. Safety that destroys the feature is not safety.

## Notes for review

- Split into three commits (storage+API → UI+wiring → docs) for reading; one PR.
- `apply.BodyFull` now delegates to `apply.BodyTrace`, so the rewrite is byte-identical
  whether or not anyone is looking. No behaviour change on the compaction path.
- `metrics.Snapshot` gains fields only — never renames or removals — guarded by the golden
  test, because `deploy/harbor/*.py` parses that payload by name.
- One new dependency: `modernc.org/sqlite` (pure Go).

## Open concerns

- **Per-request attribution of context-guru's own LLM spend** uses the delta of a
  process-global cheap-model counter, so under concurrency a request can be charged a
  neighbour's call. Correct in aggregate, approximate per row. Fixing it properly needs the
  `Model` interface to carry a per-request label.
- **`Recorder.Observe`/`MarkUnique` clear their maps on overflow** rather than evicting LRU
  (marked with `ponytail:` comments). A reset re-reports a cold start, which is honest but
  loses attribution precision; it needs a real LRU only if session churn gets large.
- **The diff falls back to a whole-block replace above 1,500 changed lines** — the LCS table
  is O(n·m). A real Myers O(nd) would remove the ceiling.
- **Benchmark ingestion keys on the run directory name**, so two runs with the same basename
  under different roots collide.
- The issue lists #31's three modes; the `mode` column and filter are in place, but only
  `active`/`bypass` occur today — `observe` needs #31.


OsherElhadad (commented):
## Independent review — verdict: **changes needed** (2 blockers), and the `modelinfo` fix should be **split out**

I reviewed this from my own worktree at `origin/feat/i30-dashboard` (e650817) and ran everything below myself rather than taking the PR body on trust. A lot of this holds up extremely well — see "what I verified and confirmed" at the end. Two findings are blockers.

---

### 🔴 BLOCKER 1 — `net_dollars_saved` is inflated by the overcount ratio (13.1× on my run). This is the exact defect the issue calls "the worst possible defect in an observability tool".

`dash/event.go:235`:

```go
e.BaselineCostUSD = e.CostUSD + float64(e.Saved())*p.CacheWrite
```

`e.Saved()` is **gross** — `tokens_before - tokens_after`, which recounts the same compaction every turn the agent re-sends the transcript. The PR is careful and correct about this everywhere in *token* space (`saved_gross` / `saved_unique` / `overcount_ratio` are separate columns, `denominators()` uses `SavedUnique` for the honest ratios). Then the dollar figure — the single number a user actually reads — silently uses gross.

Measured on 63 real replayed requests through my own build:

```
saved_gross  3,718,890    saved_unique 283,479    overcount_ratio 13.12x
reported compaction saving   $7.0244   (= gross x cache_write)
unique-based saving          $0.5354   (= unique x cache_write)
reported net_saved_usd       $7.0027
DOLLAR SAVINGS OVERSTATED 13.1x
```

The overview renders **"NET DOLLARS SAVED $7.00"** in the tile row *immediately next to* **"OVERCOUNT RATIO 13.1×"**. The dashboard is displaying the correction factor for its own headline number and not applying it. Same bug propagates to `SessionRow.SavedUSD` (`dash/query.go`), the waterfall's `compaction` step (`dash/overview.go:waterfall`), and the cumulative baseline-vs-actual chart — i.e. the issue's "the important one".

There is a second, smaller error inside the same line: even the *unique* tokens are all priced at `CacheWrite`. Re-sent (non-unique) removed tokens would have been served from cache, so at most they'd have been billed at `CacheRead` (~1/12th). Pricing them at the write rate is what produces the 11.5x-amplified inflation.

**Fix:** price unique savings at `CacheWrite`, and if you want to keep a gross line at all, price `(gross − unique)` at `CacheRead` and label it separately. Concretely:

```go
// unique content genuinely never reached the provider as new input
e.BaselineCostUSD = e.CostUSD + float64(uniqueSaved)*p.CacheWrite +
                    float64(e.Saved()-uniqueSaved)*p.CacheRead
```

Then add the regression test that would have caught this: build a window where `overcount_ratio == 10`, assert `net_saved_usd` is ~1/10th of the gross-priced figure. Right now no test asserts a dollar figure against a known-overcounted token figure, which is why an otherwise very well-tested PR shipped this.

---

### 🔴 BLOCKER 2 — content redaction runs on the request goroutine and costs **53 ms/request**, not 175 ns. The headline overhead claim measures the wrong thing.

`proxy/dashcapture.go:152-161` runs `dash.RedactContent` (9 regexes) over up to `ContentMaxPerRequest` (24) blobs × before+after = **48 × 16 KiB** — inside `capture.finish`, which is called from `serve`'s `defer` (`proxy/proxy.go:472`). A deferred call in the handler runs **before the handler returns**, so the connection is not released and (for keep-alive clients, i.e. every real agent) the next request on that connection waits behind it.

I measured this against three of my own builds on unique ports, same fake upstream, same 380 KB realistic transcript, median of 10:

```
dashboard OFF          median 214 ms
dash ON, content OFF   median 215 ms   <- +1 ms  (the 175 ns claim is fine HERE)
dash ON, content ON    median 268 ms   <- +53 ms
```

So: the capture *channel send* really is ~175 ns as claimed and `BenchmarkRecord` is honest about what it measures — but it is not the dashboard's per-request cost, because redaction was placed on the caller's goroutine. `docs/dashboard.md` and the PR body both state the dashboard costs ~0.000002% of a request; with the default `--dashboard-content=true` it costs ~25%.

The dominant cost is one regex, `dash/redact.go:161`:

```
#7  3.449 ms/blob   (?i)\b([A-Z0-9_]*(?:API_?KEY|AUTH_?TOKEN|SECRET|...)[A-Z0-9_]*)\s*[:=]\s*["']?[^\s"',}]{8,}
```

(full per-regex timings over a 17 KB log-shaped blob: #7 3.4 ms, #8 930 µs, #6 612 µs, #3 552 µs, #2 551 µs, #5 520 µs, #4 457 µs, #0 90 µs, #9 50 µs → ~7 ms per blob, ~310 ms for a full 48-blob request in the worst case.)

**Fix (either is fine, first is smaller):** move redaction to the writer goroutine. It is already the right place — `capture.go:run()` owns the event, redaction is idempotent, and nothing downstream needs it earlier. The comment at `dashcapture.go:156` says redaction "runs on the writer's caller — this goroutine, after the response is out", which is the bug in one sentence: after the response *body* is out, but still on the request goroutine and still before the handler returns. Additionally, cheap wins: gate the whole regex pass behind a `strings.ContainsAny(s, ":=")`-style prefilter, and merge #0/#2/#3/#5 into one alternation.

Also note `TestCaptureOverheadIsNegligible` cannot catch this, because it calls `Record` directly and never goes through `finish`. A regression test at the `finish` level (or an end-to-end handler latency assertion with content capture on) is what's missing.

---

### 🟠 Should the `modelinfo` fix be in this PR? **No — split it out.** But it is correct.

I reproduced both sides independently against the live LiteLLM document:

```
origin/main:                 claude-sonnet-4-5 -> window=0  ok=false   (every lookup, all four models)
this branch:  byKey=3664 priceBy=3378
              claude-sonnet-4-5   window=200000 price={3e-06 1.5e-05 3e-07 3.75e-06} ok=true
              aws/claude-sonnet-5 window=1000000 price={2e-06 1e-05 2e-07 2.5e-06}   ok=true
```

The fix is correct and genuinely per-entry (`json.RawMessage` map, `continue` on a bad entry), and the `float64`-then-`int()` handling of float-spelled integers is right. It's a real, severe, pre-existing bug — matching #39.

But it does not belong here:
1. It changes **compaction behaviour**, not observability. `components/trigger.go` resolves fraction thresholds against `Ctx.CtxWindow`; today that is always 0 so fractions are ignored and only absolutes apply. After this fix, every fraction-configured deployment starts compacting differently — on the first request, silently, with no flag. That is a behaviour change hidden inside a +9,254-line dashboard PR, and it is un-bisectable if it regresses a benchmark.
2. It is a prerequisite for the dashboard's pricing, not a consequence of it. Land it first as its own small PR (fixing #39), with the before/after resolution evidence in *that* PR's description, then rebase this one on top. The `Price`/`Pricer` additions can come with the dashboard; the per-entry decode fix should not.
3. The stated blast radius deserves its own review: no preset currently ships a fraction (`config.go` presetConfigs use absolute `min_tokens`/`min_request_tokens` only), so the practical risk today is low — but that is an argument for it being an easy separate PR, not for burying it.

The other three bug fixes are appropriately in scope (all three are on the usage-parsing path this PR introduces) and each has a test. `usage.go`'s output-only-block branch and the sniffer head+tail fix both look right, and the `app.js` bracket-balance test is a reasonable answer to "no Go test could catch this".

---

### 🟡 Medium

**`Open(":memory:")` uses `cache=shared`, so every in-memory DB in the process is the same database.** `dash/store.go:37`:

```go
return openDSN("file::memory:?cache=shared", "")
```

My test:
```go
a, _ := Open(":memory:"); b, _ := Open(":memory:")
a.insertBatch([]*Event{{TS:1, SessionID:"only-in-a"}})
// b sees 1 row  -> LEAK
```
Production only opens one, so this is not a live bug today — but it makes `:memory:` tests leak state into each other (a latent source of the flakiest possible failure), and it means the unwritable-path fallback of two proxies in one process would silently merge. Use `file:dashN?mode=memory&cache=private` with a per-instance name, or just drop `cache=shared` (a single `*sql.DB` connection pool needs it only if you rely on multiple conns seeing the same memory DB — which, with one writer goroutine, you don't... verify against `sql.DB`'s pooling before dropping, since a private in-memory DB is per-connection).

**Benchmark ingestion counted 42 runs but only 17 have data — and the UI shows 25 empty rows.** I pointed a real proxy at `/tmp/cg-runs,/tmp/tb-runs`:
```
proxy log: "ingested benchmark runs runs=17 tasks=652"
/api/benchmarks: 42 rows, 25 with NO arms  (e.g. smoke-hd, rerun2, dbg1, final50-v5)
```
`IngestBenchDir` inserts a `bench_runs` row *before* it knows whether any `rows-*.json` parsed, then returns `tasks=0` — so `IngestBenchRoots` doesn't count the run, but the row is already committed. The result is a Benchmarks tab padded with 25 contentless entries and a log line that disagrees with the UI. The PR body's "42 historical runs ingested" is really "17 ingested, 25 empty shells". Fix: only commit the run row if `tasks > 0` (or `len(rowFiles) > 0`), and make the counter and the table agree.

On the flagged basename-collision concern: I checked, and there are **no** basename collisions across those two roots today, so it's a latent issue rather than a live one. Keying on the full path (or root+basename) is a two-line fix; not a blocker.

**Redaction is a denylist for content, and it rots — as predicted.** Header and config redaction are properly allowlist-based (correct, and `RedactHeaders` is genuinely blanket-by-key). But `contentSecrets` is a pattern denylist, and it is the one place secrets from arbitrary agent output land. I threw 22 realistic credential shapes at `RedactContent`; **11 got through**:

```
LEAK  Authorization: Bearer SECRET…      -> "Authorization: «redacted» SECRET…"   (!! the value survives)
LEAK  {"api_key": "SECRET…"}             -> unchanged   (JSON form: name is quoted, #7 needs \b[A-Z0-9_]*)
LEAK  {"apiKey":"SECRET…"}               -> unchanged
LEAK  https://user:SECRET@host/path      -> unchanged   (basic-auth URL)
LEAK  postgres://admin:SECRET@db/x       -> unchanged
LEAK  {"private_key_id":"SECRET…"}       -> unchanged   (GCP service-account JSON)
LEAK  DefaultEndpointsProtocol=…;AccountKey=SECRET…  -> unchanged (Azure)
LEAK  glpat-SECRET… / sk_live_… / hf_…   -> unchanged   (GitLab PAT, Stripe, HuggingFace)
```

The `Authorization: Bearer` one is the most alarming: regex #8 matches `authorization\s*:\s*\S+` and `\S+` stops at the space before the token, so the replacement redacts the word "Bearer" and leaves the credential. Please add a test for that exact string.

The JSON-form misses are one character each — `#7`'s leading `\b[A-Z0-9_]*` won't cross the `"`. Suggest: also match `["']?[A-Za-z0-9_]*(api_?key|secret|token|password|private_key)[A-Za-z0-9_]*["']?\s*[:=]`, add a `://user:pass@` URL rule, and add the well-known prefixes (`glpat-`, `sk_live_`, `hf_`, `AIza`, `ya29.`) to the prefix alternation. Note the *config* path correctly refuses `anthropic_upstream: https://user:SECRET@…` — no wait, it doesn't: my nested-config test showed that one leaking too, because the key is allowlisted and only the *key name* is checked, never the value. A credential-in-URL value check on allowlisted string values would close both.

To be fair on what redaction does get right: the DB canary test (`redact_test.go:262`) is real, end-to-end, and asserts on every stored column; I confirmed my `sk-ant-…`/`Bearer …` request headers do **not** reach the database at all (`strings dash.db | grep SECRETCANARY` → 0 hits), because headers are never captured in the first place. Redaction is on the write path, as required. The gap is specifically arbitrary-content patterns.

**`Options.Mode` and `Options.Preset` are set in `main.go:98` and never read.** `proxy/dashcapture.go:55` takes the preset from `h.opts.Preset` instead, and mode comes from `Event.FromTrace`. Dead config surface — delete both fields, or read them.

---

### 🟢 Minor / non-blocking

- `GET /api/benchmarks?refresh=1` performs a filesystem scan and DB writes on an **unauthenticated GET**. Aggregates being open is the documented design, but a *mutating* GET behind no gate is worth putting behind `a.trusted(r)` (and arguably POST).
- `Prune`'s size rule runs `VACUUM` inside the writer goroutine's `select` loop, up to 8 times. `VACUUM` rewrites the whole file and blocks; on a 512 MiB DB that stalls capture for seconds and starts dropping events. Capture won't add request latency (good), but the drop counter will spike for reasons the UI attributes to load. Consider `incremental_vacuum` or accepting a looser size bound.
- The overview's time-series charts degenerate to a single plotted point when all data falls in one bucket (my whole replay was inside one minute). Not wrong, but a 1-point line chart with no fallback message reads as broken; consider a "not enough time range to chart" empty state.
- `percentile()` interpolates nothing and filters `col > 0`, so p95 latency silently excludes zero-latency requests. Fine, but it means p95 is over a different population than the mean beside it.
- `cg_latency_ms` recorded **876,759 ms** for my 400 KB-blob request. That's a real (pre-existing) pipeline cost, not a dashboard bug — but the dashboard is now the thing that surfaces it, and the request-detail view renders it as "876759 ms" rather than "14.6 min". `dur()` handles hours; the drawer isn't using it for this field.
- On the flagged per-request LLM-cost attribution via a process-global counter delta: **not a blocker**. It is documented, it's approximate only under concurrency, and `cg_llm_cost_usd` is ~0.3% of the figures involved. Same for clear-on-overflow vs LRU (marked `ponytail:`, honest failure mode) and the 1,500-line diff fallback (marked, with a stated ceiling). `observe` mode being plumbed-but-unused is correctly attributed to #31.

---

### Scope

54 files is large but mostly *coherent* — `dash/` is 16 files of one new package and the UI is 3. The genuinely separable pieces are: (a) the `modelinfo` per-entry decode fix → its own PR, per above; (b) benchmark ingestion (`dash/bench.go` + the Benchmarks tab, ~600 lines) is independent of per-request capture and would have been a clean third PR, as the issue itself suggested. I would not ask you to re-split the storage/API/UI at this point — the commits are already readable in that order and re-splitting now costs more than it buys.

---

### What I verified myself (vs took on trust)

**Verified by running it:**
- **`BodyFull` byte-identity — the highest-risk change: PASSES.** I compiled the same test into both `origin/main` and this branch, ran 96 configurations (6 presets × cache auto/on/off × bypass on/off × 4 conversation turns with a stateful store), and diffed SHA-256 of every output: **byte-identical, zero differences**. Also checked every caller (`proxy.chat`, `proxy.compact`, `adapters/bifrost`) — all route through `BodyFull`→`BodyTrace`→`bodyFull` and the trace is genuinely write-only. `attemptedTokens` does run unconditionally even with the dashboard off, but it's `tokens.Count` on already-memoized strings; I benchmarked `BodyFull` on a 1.4 MB body across both branches (30 iters × 3): 891/891/902 ms vs 929/897/900 ms — no regression.
- `go test -race ./...`: 20/20 green. `go test -race -count=5 ./dash/`: green, 135 s, no races.
- Writer lifecycle, my own tests: `Close()` **does** flush the tail batch (7 events with a 1-hour flush interval all landed); `Record` after `Close` neither panics nor blocks (5,000 late calls, instant); an unwritable path **does** degrade to `:memory:` instead of failing the proxy. Drop counter is `atomic.Int64`. SSE `Publish` is non-blocking with per-client buffer + evict, so a hung client cannot stall the writer. All correct.
- `/stats` backward compat: golden test is genuinely bidirectional (loses-a-field fails AND gains-a-field fails). I independently grepped every `deploy/harbor/*.py` reader — `measure.py`, `replay.py`, `replay2.py`, `swebench.py`, `sweep.py` read 16 distinct keys, all present, none renamed. Snapshot changes are additive only. **Verified safe.**
- `mkdocs build --strict`: clean. All 13 screenshots referenced, none orphaned, every `--dashboard*` flag and `DASHBOARD_*` env var documented, all 12 documented API routes return 200. Docs describe implemented behaviour — **except** the overhead claim (Blocker 2).
- All 18 committed screenshots scanned byte-wise (`strings`) for `sk-ant|sk-…|ghp_|AKIA|Bearer|api_key`: **0 hits**, and no EXIF/comment metadata. Clean.
- Model resolution before/after, live, against the real LiteLLM doc (see above).

**UI, driven with Playwright against my own build on port 4190** (40 real requests replayed from `/tmp/cg-runs/capture-swebench.jsonl`, read-only, + 42 bench dirs ingested):
- All six views render with real data; **0 console errors, 0 page errors, 0 off-origin requests** across every probe. `data-testid` on every tile as the issue required.
- CSP verified live: `default-src 'self'; img-src 'self' data:; style-src 'self'; script-src 'self'; connect-src 'self'` — and nothing violates it (the `setStyle`-via-CSSOM trick is a nice touch).
- **XSS attempt fully blocked.** I sent a request with session id `"><script>window.__XSS2=1</script>` and a `<script>` in tool_result content, then opened the row in the drawer: `__XSS`/`__XSS2`/`__XSS3` all `false`, 1 script tag (the app's own), 0 injected imgs, and the markup renders as literal text. `innerHTML` appears exactly once in `app.js` — in a comment saying not to use it. `el()` throws on an `html` prop. Genuinely good.
- Request detail + before/after diff is the best thing in this PR — real Git-style hunks with line numbers over actual replayed content, all three modes (Git diff / side-by-side / after-only) working, per-component table with outcomes (`skipped`/`acted`/`mutated only`).
- Dark mode: `rgb(15,18,22)`, correct. **390 px viewport: `scrollWidth == clientWidth == 390`, no horizontal overflow.** Long session ids (300 chars) don't break layout. Empty result set renders "0 / — / unknown" with "no priced requests", not zeros pretending to be data.

**Took on trust:** the "124 browser checks" count (I ran my own ~30 instead), the exact `BenchmarkRecord` numbers (I reproduced the conclusion for the channel send, not the digits), and that `make lint` is clean.

---

This is genuinely high-quality work — the metric semantics are thought through far past what the issue asked for, the honesty machinery (`token_accounting`, `overcount_ratio`, `uncompressed_reason`, cold-start-is-not-a-failure, four labelled denominators each carrying its own prose description, `SafetyCost` beside its benefit) is all present and correct, and the byte-identity of the rewrite path holds under 96 configurations. Fix the dollar denominator and get redaction off the request goroutine, and I'd merge it.

OsherElhadad:
Superseded by #57.

Same work, rebased onto `main` (9 PRs landed since this branch forked, HEAD `5768b00`) with the three blockers fixed. This PR could not be updated in place — the rebase rewrote history and force-push is blocked by branch protection — so the branch is `feat/i30-dashboard-rebased`.

Blockers, each with the regression test that would have caught it:

- **Dollar denominator.** Baseline now prices UNIQUE savings at the cache-write rate and the re-sent remainder at the cache-read rate. On the same 63 replayed requests: **$0.002235 → $0.000241**, 9.3x inflation removed. `TestDollarsDeriveFromUniqueNotGrossSavings` pins the aggregate against a deliberately-overcounted fixture and fails on the old formula.
- **Redaction on the request goroutine.** Moved to the writer goroutine, still before the INSERT. With content capture ON, medians of 40 paired requests: **+53 ms → −3 ms** (within noise). The new handler-level guard measures +87 ms and fails if it moves back.
- **Content redaction denylist.** 22/22 shapes now redact (`Authorization: Bearer` included), pinned by a table test, plus a credential-in-URL check on allowlisted config *values*. Since a denylist over arbitrary agent output cannot be proved complete, `--dashboard-content` now defaults **off**.

Also: unique `:memory:` DSN, atomic bench-run commit (log 17 / API 17 / 0 empty rows, was 42 / 25), `dash.Options.Mode` wired and `.Preset` deleted, and the observe banner + `observe_queue` counters now that #43 and #50 are merged.

Thanks for the review — the verification work in it (96-configuration byte-identity, the per-regex timings, the 22 shapes, the Playwright pass) is what made these findable, and every one of them reproduced exactly as described.

PR #43, opened 2026-08-10
feat(proxy): add observe mode — measure what compaction would save, without enforcing it

### LINKED ISSUE #31: feat(proxy): three operating modes — sync, async (cache-safe deferred compaction), and observe (measure without enforcing)

OsherElhadad:
Partially closes #31: **observe mode lands here; async is deferred to #35** (kept as a
draft, all its review fixes intact).

## Why split

#31 specified three modes. Observe is the smaller, lower-risk half and delivers the
adoption story on its own — the issue itself suggested landing it first. Async is held
because its measured benefit has since evaporated:

- its headline latency win (1,599 ms → 25 ms) came almost entirely from deferring
  `extract_llm`, and **#34 hard-declines `extract_llm` on caching backends** — so on the
  primary workload there is nothing expensive left to defer. Terminal-Bench already showed
  exactly this: zero LLM calls, async 26.8 ms vs sync 26.9 ms;
- its cache-write evidence was withdrawn under review (the tail protection turned out to be
  inert on claude-code, which sets its own breakpoints), so re-establishing it needs a fresh
  50-task paired arm;
- its `dropped` / `stale_discarded` paths — the ones guarding the correctness invariant —
  are exercised by tests only, never by production load.

Carrying a bounded worker pool, per-session generations, a copy-on-write store overlay and a
cache policy that must decline on the default agent is a lot of concurrency surface for an
unproven benefit. The primitives held up under a hostile review, which is why #35 stays on a
branch rather than being discarded: it is ready to measure, not ready to merge.

## What observe does

The agent receives its request **untouched, byte for byte**.

Byte-identity is **structural**, not a property of careful copying: the request path does
not run the pipeline at all, and skips `expand.Inject` too — injecting a tool declaration
would be a modification. There is no code path in observe mode that could alter a forwarded
body. A test asserts it anyway, including on a tool-carrying request where `expand.Inject`
would otherwise append.

A copy runs off-path against observe's own store, purely to record what compaction *would*
have achieved.

```yaml
mode: observe     # sync (default) | observe
```

**Cost to the enforced path: 0.062 ms/req**, against `sync`'s 1,599 ms on the same
benchmark. Confirmed independently in a live Claude Code session (0.209 ms vs 28.964 ms) and
on Terminal-Bench (0.076 ms vs 26.9 ms).

headroom has no observe/shadow/dry-run mode at all — its `token` and `cache` modes are both
enforcing, and its only control arm is a 10% output-shaper holdout. This is a genuine
differentiator, not a port.

## Why the projection is trustworthy

Both requirements were found by comparing the projection against `sync`'s actuals, not by
reading code — and each was a real bug:

- **It shares the cached-prefix boundary the enforced path uses.** Without that,
  cache-awareness gates nothing, every message looks compactable, and the projection
  overstates by exactly what cache-awareness costs: **9.5% projected against 0.8% actually
  achieved** on the same tasks. A projection that ignores cache-awareness is a projection of
  what a *cache-blind* proxy would do.
- **Its store is persistent.** Offloaders *freeze* a decision and replay it on every later
  turn, which is where most of the sustained saving lives. Against a discarded buffer observe
  sees only the current tail and **under**-projects by 3x.

It is nevertheless **disjoint** from the live store: a decision observe made must never be
replayable by a real request — that would be a request modification arriving by the back
door. A test asserts zero writes to the live store.

Three lines of evidence, strongest first:

| Evidence | Result |
|---|---|
| Controlled same-traffic comparison | **exact**: 10,020 tokens / 23.06% both sides |
| Terminal-Bench (negative control) | sync 1.02% actual → observe **0% projected** |
| SWE-bench arms | 6.40% vs 0.82% — consistent but noisy, reasons given, not explained away |

The negative control is the more convincing shape: on traffic with almost nothing to save,
observe reports almost nothing rather than inventing a headline. A mode that only ever
agreed on high-savings traffic would be far weaker proof.

## Metric namespace separation

Observe results live in **physically separate** accumulators serialized under `potential_*`
/ `projected_*`, sharing no key with an enforced metric. In observe mode every enforced
savings aggregate is zero by construction — `requests`, `saved_tokens`, `sync_enforced`,
`components: {}` — which is the machine-readable form of "context-guru did not modify
requests". A mislabelled hypothetical silently inflates the headline claim, so this is a
correctness boundary, and a test asserts no enforced aggregate can reach an observe result.

Two enforced keys are deliberately **real** there rather than zeroed:

- `cg_added_ms_avg` — the actual enforced-path latency. Reading ~0 **is** the headline;
  zeroing it would hide the result.
- `llm_calls` / `llm_input_tokens` / `llm_output_tokens` — context-guru's own model spend.
  Observe measures off-path and that costs real money, so relabelling it `potential_*` would
  be a worse lie than leaving it. It stays where cost tooling reads it, labelled by
  `observe_llm_notice` as the cost of measuring rather than enforcing.

## Also included: the `prevLen` race fix

Independent of modes, and worth flagging for #40's author. The per-session cached-prefix
boundary was read from the store and written back in a `defer`, so two concurrent turns of
one session both read the same length and the second's write-back could land first — leaving
a boundary that describes neither turn. A boundary that is too high lets an offloader mutate
content the provider has already cached, costing a full cache-write of the suffix.

`modes.Tracker.Turn` now reads and records it in one locked call. Callers without a tracker
(library users, `/compact`) keep the legacy path unchanged.

## `store.Buffer` / `FrozenLoser` — async-only, flagged deliberately

I raised that a component asserting `c.Store.(store.FrozenLoser)` would see a *wrapper* and
silently take the degraded path. **That concern does not apply to this branch:** observe's
shadow store is a plain `*store.Memory`, which implements `FrozenLost` natively. The
copy-on-write `store.Buffer` exists only on #35, where the forwarding fix lives with it.
Nothing is lost by #35 going to draft.

## Compatibility

`mode: sync` remains the default and is byte-identical to today, asserted by a golden test
comparing the two entry points' output. `/stats` gains `mode`, `sync_enforced` and the
observe hypotheticals; every pre-existing field keeps its name and shape (a test pins the
full set, since `deploy/harbor/*.py` parses this payload). `/compact` is mode-independent by
contract, also tested.

Rebased onto current `main`, so it includes #36 and #40; the projection-agreement result was
re-verified after that rebase.

## Tests

`go test -race` throughout:

- `sync` byte-identical to the legacy entry point, and unset mode identical to explicit
  `sync` (golden);
- observe leaves the forwarded body byte-identical, including a tool-carrying request that
  `expand.Inject` would otherwise modify;
- observe's projection agrees with `sync`'s actual on identical traffic (fails at ratio 0.33
  without the shadow store);
- observe never writes the live store;
- observe metrics unreachable from every enforced savings aggregate, and the two
  vocabularies disjoint in the serialized payload;
- concurrent turns of one session cannot corrupt the boundary, and no concurrent write is
  lost;
- enqueue dedup atomic against 32 concurrent enqueues of one key; a full queue drops, counts
  and never blocks; a panicking observation is contained; `Stop` does not wait on an
  uncancellable job; no goroutine leak on `Stop` or `Handler.Close`, both idempotent;
- `/stats` backward compatibility; `/compact` mode-independence; unknown mode rejected
  (including `async`, which is not valid on this branch).

Gates: `build` · `test -tags cg_skeleton` · **`-race -count=5`** · `gofmt -l` · `go vet` ·
`mkdocs build --strict` — all clean.

## Docs

New `docs/how-to/operating-modes.md` (when to use each, how to read the projection, and
what observe *cannot* tell you — cache effects are projected not measured, no expand bounce
is exercised, and measuring still spends cheap-model tokens). New
`docs/results/observe-mode.md` with the three evidence lines and a "what is not established"
section. `docs/design.md`, `docs/reference/config.md`, `docs/how-to/measure-savings.md`,
`mkdocs.yml`, README.

## Remaining for #31

Async mode, on #35 (draft): benefit conditional and currently unmeasured now that
`extract_llm` no longer runs on caching backends. What would establish it: a 50-task paired
arm with `async.strip_caller_breakpoints: true`, showing cache-write no higher than `sync`'s
and a real latency reduction on traffic that still makes model calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,34 @@ are in **[docs/components.md](docs/components.md)** and **[docs/results/componen
 | `smartcrush` | Offload | keeps anchor items of a long JSON array, drops the middle |
 | `summarize` | Offload (LLM) | compresses the middle of the trajectory into one summary (run alone) |
 
+## Operating modes
+
+`sync` (the default) compacts inline and the caller waits. One other mode changes that:
+
+| Mode | The request path | Use it when |
+|---|---|---|
+| **`sync`** *(default)* | Compacts inline; the caller waits. | You want the savings. |
+| **`observe`** | Forwards the request **untouched, byte for byte**, and reports what compaction *would* have saved. | You want to evaluate context-guru on your own traffic without enforcing it. |
+
+```yaml
+mode: observe
+```
+
+Byte-identity is **structural**: in observe mode the request path never runs the pipeline
+at all (and never injects the expand tool), so no code path could alter a forwarded body.
+Measured cost to the enforced path: **0.062 ms/req**, against 1,599 ms for `sync` on the
+same benchmark.
+
+Observe-mode numbers are reported under their own `potential_*` / `projected_*` keys that
+share no name with an enforced metric, so a hypothetical can never be read as a realized
+saving. On identical traffic its projection matches what `sync` actually achieved exactly
+(23.06% both sides), and on traffic with nothing to save it correctly projects 0%.
+
+This is a genuine differentiator, not a port: headroom has no observe/shadow/dry-run mode
+at all — its `token` and `cache` modes are both enforcing.
+
+Details in [docs/how-to/operating-modes.md](docs/how-to/operating-modes.md).
+
 ## Integrate
 
 | Option | What | Where |
@@ -182,7 +210,8 @@ Details in [docs/integrations.md](docs/integrations.md).
 
 ## Docs
 
-- [docs/design.md](docs/design.md) — architecture: component model, fail-open pipeline, store, session, expand loop, metrics.
+- [docs/design.md](docs/design.md) — architecture: component model, fail-open pipeline, store, session, expand loop, metrics, operating modes.
+- [docs/how-to/operating-modes.md](docs/how-to/operating-modes.md) — sync vs observe: when to use each, and how to read observe's projections.
 - [docs/components.md](docs/components.md) — every registered component: how it works, live before→after, lossiness, config, best use.
 - [docs/integrations.md](docs/integrations.md) — proxy gateway vs AuthBridge plugin, with request paths.
 - [docs/setup.md](docs/setup.md) — setup + a concrete SWE-bench run through the eval-containers gateway.
--- a/apply/apply.go
+++ b/apply/apply.go
@@ -112,7 +112,19 @@ func BodyWithModelWindow(ctx context.Context, pipe *components.Pipeline, st stor
 // cache-awareness when the backend is a prompt-caching provider or the request
 // already carries cache_control breakpoints; "on" forces it; "off" restores the
 // legacy compact-everything behavior (correct for confirmed non-caching backends).
-func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, provider bschemas.ModelProvider, body []byte, explicitSession string, bypass bool, models components.ModelSpec, window int, cacheMode string) (result []byte, changedBody bool) {
+func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, provider bschemas.ModelProvider, body []byte, explicitSession string, bypass bool, models components.ModelSpec, window int, cacheMode string) ([]byte, bool) {
+	r := BodyOpts(ctx, pipe, st, Opts{
+		Provider: provider, Body: body, Session: explicitSession, Bypass: bypass,
+		Models: models, Window: window, CacheMode: cacheMode,
+	})
+	return r.Body, r.Changed
+}
+
+// BodyOpts is the full entry point: everything BodyFull takes plus the operating mode
+// (#31) and the per-session boundary tracker. Hosts that support modes call this;
+// BodyFull is the positional shim every other caller keeps using.
+func BodyOpts(ctx context.Context, pipe *components.Pipeline, st store.Store, o Opts) (res Result) {
+	body, provider, bypass := o.Body, o.Provider, o.Bypass
 	// Top-level fail-open backstop: the per-component recover in pipeline.runOne only
 	// covers component code. A panic anywhere else on the rewrite path (normalize, the
 	// sjson splice, rebuildCountChanged, a marshal) must NOT 500 the client — forward
@@ -120,13 +132,18 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	// the whole entry point, not just inside components.
 	defer func() {
 		if r := recover(); r != nil {
-			slog.Error("context-guru: recovered from panic in BodyFull; forwarding original request", "panic", r)
-			result, changedBody = body, false
+			slog.Error("context-guru: recovered from panic in BodyOpts; forwarding original request", "panic", r)
+			res = Result{Body: body}
 		}
 	}()
+	mode := o.Mode
+	if mode == "" {
+		mode = components.ModeSync
+	}
+	models := o.Models
 	msgsRaw := gjson.GetBytes(body, "messages")
 	if !msgsRaw.Exists() || !msgsRaw.IsArray() {
-		return body, false
+		return Result{Body: body}
 	}
 
 	// Volatile-tail split, before anything else touches the body. This is a
@@ -149,7 +166,7 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 
 	norm, slots := normalize(provider, msgsRaw.Array())
 	if len(norm) == 0 {
-		return body, systemSplit // keep the split even with nothing to compact
+		return Result{Body: body, Changed: systemSplit} // keep the split even with nothing to compact
 	}
 
 	if debugTraffic {
@@ -157,16 +174,27 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	}
 	chat := &bschemas.BifrostChatRequest{Provider: provider, Input: norm}
 	sys, firstUser := systemAndFirstUser(norm)
-	sessionID := session.Resolve(explicitSession, sys, firstUser)
-	cacheAware := resolveCacheAware(cacheMode, provider, body)
+	sessionID := session.Resolve(o.Session, sys, firstUser)
+	res.Session = sessionID
+	cacheAware := resolveCacheAware(o.CacheMode, provider, body)
 	maxCachedIdx := -1
 	if cacheAware && !bypass {
 		// Messages present on the previous turn of this session are already committed
 		// to the provider cache; only the new tail is being cache-written this turn.
 		// Restrict supersession/age offloaders to that tail so they never mutate the
 		// cached prefix. Growth-based (dialect-agnostic; needs no cache_control mapping).
-		maxCachedIdx = prevLen(st, sessionID) - 1
-		defer putLen(st, sessionID, len(norm))
+		//
+		// With a Tracker the boundary is read and this turn's recorded in ONE locked call.
+		// The legacy path below read it from the store and wrote it back in a `defer`, so
+		// two concurrent turns of one session raced on it — see the modes package comment.
+		// Callers without a tracker (library users, /compact) keep the legacy path: same
+		// numbers, same race, no behavior change for them.
+		if o.Tracker != nil {
+			maxCachedIdx = o.Tracker.Turn(sessionID, len(norm)) - 1
+		} else {
+			maxCachedIdx = prevLen(st, sessionID) - 1
+			defer putLen(st, sessionID, len(norm))
+		}
 	}
 	c := &components.Ctx{
 		Ctx:          ctx,
@@ -174,13 +202,14 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 		Store:        st,
 		Model:        models,
 		Bypass:       bypass,
-		CtxWindow:    window,
+		CtxWindow:    o.Window,
 		CacheAware:   cacheAware,
 		MaxCachedIdx: maxCachedIdx,
 		// Every breakpoint already on the wire — including the ones no component can
 		// see (`system`, `tools`, and the marks our own normalize drops). The
 		// provider's cap of four counts them all (issue #32, defect 2).
 		ExistingBreakpoints: wireBreakpoints(body),
+		Mode:                mode,
 	}
 
 	// Canonical form of each normalized message BEFORE the pipeline, so a
@@ -191,6 +220,7 @@ func BodyFull(ctx context.Context, pipe *components.Pipeline, st store.Store, pr
 	}
 
 	rr := pipe.Run(chat, c)
+	res.Run = rr
 
 	// A component changed the message count (summarize restructures the transcript
 	// to [msg0, <summary>, last-K]). Rebuild the messages array preserving each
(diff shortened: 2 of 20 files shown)


OsherElhadad:
Rebased onto current `main` (now includes #36, #40 and #42). Three conflicts resolved, all pure unions — #42's cmdfilter ledger fields and my mode fields are additive in the same `Aggregator`/`Snapshot` regions, and #40's `FrozenLoser` work needed no adjustment here.

Re-verified after the rebase: the projection-agreement result is unchanged (`sync` saved 10,020 / 23.06%, `observe` projected 10,020 / 23.06% — exact), and all gates are green including `-race -count=5`.

Worth restating since #40 has now landed: observe's shadow store is a plain `*store.Memory`, so it picks up `FrozenLost` natively and there is no wrapper to forward through. The `store.Buffer` → `FrozenLost` forwarding I flagged belongs only to the async branch (#35, draft), where `store.Buffer` lives.

PR #57, opened 2026-08-10
feat(dash): first-class persistent observability dashboard

### LINKED ISSUE #30: feat(dashboard): first-class persistent observability UI — what value is context-guru providing?

OsherElhadad:
Closes #30. **Supersedes #38** — same work, rebased onto `main` (9 PRs landed since that
branch forked) with the three review blockers fixed. #38 could not be updated in place: the
rebase rewrote history and force-push is blocked by branch protection.

### 🔴 BLOCKER 1 — `net_dollars_saved` was inflated (gross vs unique)

`Event.Price` built the baseline from `Saved()` (tokens_before − tokens_after for **this**
turn). The agent re-sends its whole transcript every turn, so one compaction was paid for
once per remaining turn — and the tile rendered the correction factor, `overcount_ratio`,
immediately beside the figure without applying it.

Two errors on one line, the second multiplying the first: non-unique removed tokens would
have been served as cache **reads**, not writes, so pricing them at `CacheWrite` (11.5× a
read) inflated on top of the overcount. Now unique savings are priced at the write rate and
the re-sent remainder at the read rate, clamped so a per-component unique figure exceeding a
request's own gross saving cannot drive the remainder term negative.

Every downstream figure — session rows, the savings waterfall, the cumulative baseline chart
— sums the per-row `baseline_cost_usd`, so all of them follow from the one fix.

Measured on the same 63 replayed SWE-bench requests the review used:

| | old (gross × cache-write) | new (unique-based) |
|---|---|---|
| net savings | **$0.002235** | **$0.000241** |

**9.3× inflation removed**, verified two ways: recomputing from the per-row API against the
served aggregate (exact match to 9 decimal places), and `TestDollarsDeriveFromUniqueNotGrossSavings`,
which builds a deliberately-overcounted fixture (one 1,000-token compaction re-sent over ten
turns, `overcount_ratio` 10) and pins the aggregate dollar figure. Reverting `Price` to the
old formula makes it fail: `$0.0250` where `$0.0043` is correct.

### 🔴 BLOCKER 2 — redaction ran on the request goroutine

`capture.finish` is called from `serve`'s `defer`, which runs **before the handler returns**,
so nine regexes over up to 48 blobs were paid by the next request on a keep-alive connection
— every real agent. `BenchmarkRecord`'s ~175 ns was honest about the channel send and blind
to the rest, because it calls `Record` directly and never `finish`.

Redaction moves to the writer goroutine, which already owns the event. Secrets still never
reach disk: it now runs immediately before the INSERT rather than at the capture site, so
what changed is which goroutine pays. The regex pass also got cheaper (one merged prefix
alternation instead of seven, and a `ContainsAny(":=")` prefilter in front of the expensive
key-assignment rules).

Measured with content capture **ON**, 24-tool-result transcript, one keep-alive connection,
medians of 40 **paired and interleaved** requests against a dashboard-off handler:

| | before | after |
|---|---|---|
| added per request | **+53 ms** (~25%) | **−3 ms** (within noise) |

The guard is now `TestDashboardAddsNoRequestLatencyWithContentCapture` at the handler level.
Restoring redaction to the request path measures **+87 ms** and fails it. The pairing matters:
measuring one config fully and then the other attributes machine drift to the dashboard, and
consecutive runs of that shape disagreed by 4×.

### 🔴 BLOCKER 3 — content redaction leaked 11 of 22 credential shapes

Reproduced (12 with a lowercase variant), now **22/22 redact**, pinned by a table-driven
test. The worst was `Authorization: Bearer <token>`: the pattern matched `\S+` after the
colon, so it redacted the word "Bearer" and left the credential rendered in the diff view.
Also fixed: JSON-form `{"api_key": …}` (the leading `\b[A-Z0-9_]*` could not cross the
quote, which also let GCP service-account keys through), basic-auth URLs, Azure connection
strings, and the `glpat-` / `sk_live_` / `hf_` / `AIza` / `ya29.` prefixes. Auth headers now
match to end-of-line; URL userinfo has its own rule that replaces only the password so the
host stays readable. The config path gets the other half of the same gap — an allowlisted key
like `anthropic_upstream` had its **value** passed through verbatim, and an upstream URL is
exactly where a `user:password@` credential lives.

But 22/22 passing does not make a denylist complete, and content is arbitrary agent output
that cannot be allowlisted the way headers and config keys are. So **`--dashboard-content`
now defaults OFF**. The diff view is still the best thing here; it is opt-in.

### Mediums

- `Open(":memory:")` used `file::memory:?cache=shared`, where the **name** identifies the
  database — so every in-memory dashboard in the process was the same one. Per-instance name,
  with the leak test. `cache=shared` stays: a private in-memory DB is per-connection and
  `database/sql` pools connections.
- `IngestBenchDir` committed the run row before knowing whether any `rows-*.json` parsed, so
  the log said 17 runs while the API returned 42 with 25 empty. Committed only when
  `tasks > 0`. Verified live against `/tmp/cg-runs,/tmp/tb-runs`: **log 17, API 17, 0
  empty-arm rows** (was 42 / 25).
- `dash.Options.Mode` was set to a hardcoded `"active"` and never read; it now carries the
  real mode. `dash.Options.Preset` was a second copy of a value the capture site already
  reads from `proxy.Options` — deleted rather than wired.

### Observe mode (#43) and `observe_queue` (#50)

Both merged while #38 was open, so the dashboard now renders them. `/api/capture` serves the
mode and the off-path pool's counters, and the UI shows an unmissable banner: *"You are
currently in OBSERVE mode. context-guru did not modify any request…"* plus
`processed`/`pending`/`dropped`. `dropped` is called out specifically, because a dropped
observation means the projection **understates** what compaction would have saved. Verified
live on a `--mode observe` proxy.

`applyMode` returns the zero `Trace` in observe mode deliberately: the enforced path ran
nothing, so crediting the off-path projection to a request that was forwarded untouched is
exactly the confusion the `potential_*` namespace exists to prevent.

### Rebase notes

`Trace` is now **embedded in** `apply.Result` rather than living beside it, so `Session` and
`Run` exist once instead of twice — #43's `BodyOpts`/`Result` refactor and this branch's
`BodyTrace` had introduced two copies of both. `BodyTrace` is gone; `BodyOpts` is the single
implementation and `BodyFull` still delegates to it, so the rewrite stays byte-identical
whether or not anyone reads the trace. Two early returns that built a fresh `Result` were
changed to assign, or a bypassed request with no `messages` array would have reported itself
as `no_messages`.

`metrics.Snapshot` conflicts were additive on both sides and all fields are kept — the naive
keep-both would have folded the dashboard's token-tier fields **into `QueueStats`**, which is
the interleaving hazard flagged for `components/go`. The eight observe/mode keys from #43/#50
are added to the golden test's reviewed list; `/stats` remains append-only.

### Gates

`CGO_ENABLED=1 go build -tags cg_skeleton ./...` · `go test -tags cg_skeleton ./...` ·
`go test -race ./... -count=2` · `make lint` · `mkdocs build --strict` — all clean.

UI re-verified with Playwright against a live build on port **4193** replaying the 63 captured
requests: 0 console errors, 0 page errors, 0 off-origin requests, all six views render,
request detail and diff work, XSS flags still `false` with one script tag (the app's own), no
horizontal overflow at 390 px. The **overview screenshot is re-shot** because the dollar
figures changed, and scanned for credential shapes and EXIF/XMP/GPS before committing.

### Not fixed, deliberately

Global-counter LLM attribution, clear-on-overflow vs LRU, and the 1,500-line diff fallback
stay as documented `ponytail:` ceilings, per the review. `?refresh=1` mutating on an
unauthenticated GET, `VACUUM` in the writer loop, the single-bucket chart degenerating to one
point, `percentile()`'s `> 0` filter, and `cg_latency_ms` rendering as raw ms in the drawer
are all still open and non-blocking.

---

<details>
<summary>Original #38 description</summary>

Closes #30

context-guru had no dashboard: the entire observability surface was `GET /stats`, one
in-memory `Aggregator` snapshot that vanished on restart. A user could not answer the
question the product exists to answer — **what value is context-guru providing?**

This adds a persistent dashboard: durable per-request storage, an off-hot-path capture
pipeline, a filterable JSON/SSE API, and an embedded single-page UI. `--dashboard` turns it
on; without the flag the route table is byte-identical to before and `/stats` is unchanged.

Overview

## What is here

**Storage** — SQLite via `modernc.org/sqlite` (pure Go, no extra C toolchain), WAL mode.
`requests` · `request_components` · `request_content` (gzip, size-capped, skippable) ·
`bench_runs`/`bench_tasks`. Epoch-**ms** timestamps throughout; a formatted locale string
cannot be range-queried, sorted portably, or bucketed. Retention bounded by age **and**
size. Schema versioned, with a mismatch renaming the old file aside rather than refusing to
boot. `--dashboard-db :memory:` for no-persistence, which is also the automatic fallback
when the path is unwritable.

**Capture off the hot path** — the handler builds one struct from values the request path
already computed and does a channel send with a `default:` branch. A full queue **drops and
counts**; the drop count is surfaced in the UI, because a dashboard that hides its own
coverage gaps cannot be trusted about anything else. One writer goroutine batches inserts
and fans summaries to SSE.

**API** — `/api/stats`, `/api/series?bucket=`, `/api/requests` (server-side filters +
**keyset** pagination), `/api/requests/{id}`, `/api/sessions`, `/api/components`,
`/api/facets`, `/api/config`, `/api/benchmarks[/{id}/tasks]`, `/api/capture`, `/api/events`
(SSE honoring `Last-Event-ID`). No rollup tables — series are bucketed in SQL at query time.

**UI** — three files via `go:embed`. No npm, no bundler, no framework, and **no CDN**:
context-guru ships into VPC and air-gapped contexts, so a test fails on any external origin
and the assets are served under `default-src 'self'`. Charts are hand-drawn SVG. Design
tokens are CSS custom properties from line one, so dark mode redefines only tokens.

## Metric semantics — the actual point

| Requirement | How |
|---|---|
| Multiple labelled denominators | Four ratios, each shipping its divisor **and a prose description in the payload**: of-attempted · of-new-provider-billed-input · of-whole-request (diluted) · unique-of-whole |
| Never divide savings by themselves | The new-input ratio is guarded on the **billed** figure; with no usage data it reports **n/a**, not ~100% |
| Never count fake savings | gross · unique · adjusted are separate columns; `overcount_ratio` surfaced |
| Report the cost of our own safety | Frozen-for-cache-safety tokens, restorations, reverts, and context-guru's own latency and LLM spend, each beside its benefit |
| `token_accounting` per row | `complete\|partial\|missing`. An unpriceable request reads **unknown** — never zero, never exact |
| Cache-miss attribution | `hit · cold_start · ttl_expiry · prefix_change · unknown`; cold start is **not** a failure, TTL wins ties |
| "Why didn't you compact this?" | First-class reason bucket: `bypassed · below_trigger · cache_frozen · found_nothing · reverted · no_messages` |

Cost is computed at **write** time so history does not reprice; percentages at **read** time
so a filter change needs no rebuild.

## The headline feature: the before/after diff

Both reference implementations carry this data and neither renders it.

Git-style content diff

LCS diff with common head/tail trimmed first, Git-style hunks with line numbers and
collapsed unchanged runs, plus side-by-side and after-only modes. Blocks are ordered
biggest-saving first and the largest opens — leading with an unchanged block buries the
answer.

## Per-component economics

Per-component economics

On the real traffic above the verdicts do their job: `extract` is *earning its place*,
`cacheinject` *mutates, saves no content* (its win is provider-side), and a component that
burned wall time for nothing reads *costly and inert*.

## Benchmarks

Benchmark comparison

Ingested straight from the harness artifacts (`summary.json` + `rows-<arm>.json`) — no new
export format, so **42 historical runs from `/tmp/tb-runs` and `/tmp/cg-runs` ingested on
first boot** (652 task rows). Per arm: solve rate, mean reward, total/mean cost, **cost per
solve**, cache hit rate, exceptions, with a cost-vs-reward scatter and per-task drill-down.
Re-ingesting replaces a run rather than duplicating it.

## Sessions, requests, config, dark mode

Sessions
Request detail
Effective configuration
Dark mode

## Measured capture overhead

The issue asks for this explicitly, since a dashboard that costs request latency is a
regression in a tool that sells latency awareness.

```
BenchmarkRecord-16              200000        7567 ns/op   1141 B/op   7 allocs/op
BenchmarkRecordFullQueue-16     200000         388 ns/op    399 B/op   1 allocs/op
BenchmarkObserve-16             200000          64.8 ns/op
TestCaptureOverheadIsNegligible: capture overhead 175 ns per request (20,000 events)
```

**~175 ns per request** on the request goroutine — against a measured 8.4 s mean upstream
round trip on this workload, that is ~0.000002%. `BenchmarkRecord`'s 7.5 µs includes the
benchmark's own `&Event{}` allocation, which the handler does regardless. The full-queue
(drop) path is 388 ns, so an overloaded dashboard degrades instead of becoming a latency
incident. A regression test fails above 50 µs, so moving I/O onto the capture path breaks CI.

## Security

Redaction happens **before the database, never on read** — a secret on disk is a secret
forever, and a redact-on-read filter is one forgotten code path from leaking it. Headers are
blanket-redacted by key against a short allowlist (a denylist fails the moment a gateway
invents a new auth header); config keys are allowlisted with credential-named keys always
withheld; content is pattern-scrubbed **then** capped, so a secret at the truncation
boundary cannot survive by being cut into an unmatchable prefix.

Per-request content and the effective config are loopback/trusted-CIDR only; aggregates stay
open, because a proxy bound to `0.0.0.0` should still report its own numbers. There is
deliberately **no** "disable observability in production" switch.

## Real-data validation (not fixtures)

Everything above is rendered from real traffic through this proxy on port 4107:

- **40 real requests** replayed from captured SWE-bench and Terminal-Bench `claude-code`
  transcripts, against the live provider — so all four billed token tiers, the prompt-cache
  structure, cold starts and TTL expiries are real, not synthesised.
- A real `claude` CLI session through the proxy.
- **42 benchmark runs / 652 task rows** ingested from the existing jobs roots, with per-task
  drill-down verified.
- Data survived a proxy restart (persistence verified end to end).

Observed: 4 sessions · 120k → 106k content tokens · 14k gross / 2.0k unique saved
(overcount **7.0×**) · 328k cache reads vs 144k writes · $0.46 baseline vs $0.44 actual ·
accounting 32 `complete` / 9 `partial` · cache misses attributed 26 hit / 8 unknown /
4 cold_start / 3 ttl_expiry.

## Tests

- `dash`: schema migration + version-mismatch preservation, retention by age **and** size
  (with cascade), keyset pagination covering every row exactly once, **every** filter
  dimension, query-time bucketing, redaction (incl. a canary written through the real
  capture path and read back from every column), the capture channel dropping rather than
  blocking with the drop counted, SSE fan-out with a **hung client** evicted, `Last-Event-ID`
  backfill, cache attribution, benchmark ingest from real harness-shaped artifacts.
- `proxy`: **golden `/stats` shape** in both directions (a lost field fails; a new field
  fails until recorded), usage parsing in both dialects, the streaming sniffer, and
  end-to-end capture through a real pipeline.
- `go test -race ./...` — **20/20 packages green**, including the capture path and SSE hub.
- `make lint` clean; `mkdocs build --strict` clean.

**Browser verification: 124 checks, 0 failures** against the live app — every stat tile's
`data-testid`, chart rendering and tooltips, filters, pagination, the drawer, all three diff
modes, benchmark drill-down, dark/light, a 390 px viewport, and empty states. Every
screenshot was scanned for credentials byte-wise and by OCR before being committed.

390px viewport
Empty state

## Bugs found and fixed along the way

Four real defects, each with a regression test:

1. **`internal/modelinfo` never resolved anything in production.** The LiteLLM prices map
   was decoded into one typed `map[string]struct{…}`, which fails on the document's
   prose-filled `sample_spec` entry and on models that spell an integer as a float —
   yielding **nothing**, so every context-window lookup silently returned "unknown" and
   fraction-based triggers have never fired. Per-entry decode keeps the ~2,900 good rows;
   the resolved rates now match the harness's published price vector exactly. *(Pre-existing,
   unrelated to the dashboard.)*
2. **Streaming output tokens were under-reported.** An output-only usage block — Anthropic's
   `message_delta`, which carries the *final* completion count — parsed as "no usage".
3. **The streaming sniffer dropped its tail window** whenever both halves saturated, losing
   usage for every large streamed response.
4. **A dropped paren in `app.js` rendered a blank dashboard.** Go compiled, every Go test
   passed, the HTML served 200. There is now a test that parses `app.js` (node when present,
   a bracket-balance scan otherwise).

Plus two redaction over-reaches caught by the browser checks: redacting the `components`
subtree wholesale made the config view show nothing, and a substring match on `token` also
swallowed `max_tokens`/`min_tokens`. Safety that destroys the feature is not safety.

## Notes for review

- Split into three commits (storage+API → UI+wiring → docs) for reading; one PR.
- `apply.BodyFull` now delegates to `apply.BodyTrace`, so the rewrite is byte-identical
  whether or not anyone is looking. No behaviour change on the compaction path.
- `metrics.Snapshot` gains fields only — never renames or removals — guarded by the golden
  test, because `deploy/harbor/*.py` parses that payload by name.
- One new dependency: `modernc.org/sqlite` (pure Go).

## Open concerns

- **Per-request attribution of context-guru's own LLM spend** uses the delta of a
  process-global cheap-model counter, so under concurrency a request can be charged a
  neighbour's call. Correct in aggregate, approximate per row. Fixing it properly needs the
  `Model` interface to carry a per-request label.
- **`Recorder.Observe`/`MarkUnique` clear their maps on overflow** rather than evicting LRU
  (marked with `ponytail:` comments). A reset re-reports a cold start, which is honest but
  loses attribution precision; it needs a real LRU only if session churn gets large.
- **The diff falls back to a whole-block replace above 1,500 changed lines** — the LCS table
  is O(n·m). A real Myers O(nd) would remove the ceiling.
- **Benchmark ingestion keys on the run directory name**, so two runs with the same basename
  under different roots collide.
- The issue lists #31's three modes; the `mode` column and filter are in place, but only
  `active`/`bypass` occur today — `observe` needs #31.

</details>


OsherElhadad:
Superseded by #58 — same five commits rebased onto `main` after #54, #55 and #56 landed. Branch protection blocks force-push, so the rebase had to go to a new branch.

Three conflicts resolved, one of them consequential: both this branch and #56 changed the same `apply.BodyOpts` call in `/compact`. #56 added context-window resolution and the shared boundary `Tracker`; this branch added capture instrumentation. Kept both, with **#56's `Window: window`** — the dashboard side passed `Window: 0`, which is exactly the bug #56 fixed, so taking that side would have silently reintroduced the eval-vs-production divergence.

Also worth noting: this PR's own `/stats` golden test caught all 15 fields the intervening merges added (`frozen_*`, `mode`, `sync_enforced`, `potential_*`, `projected_*`, `actual_baseline_tokens`, `extract`, `top_discarded`) and refused to pass until each was consciously added to the reviewed list. That is the test doing precisely what it was written for.

PR #80, opened 2026-08-17
feat(coref): co-reference-aware compaction

amiddavid:
Implements [co-reference-aware compaction](docs/proposals/coref-compaction.md) — picking *what* to drop
at a threshold crossing by looking at **back-references** rather than at content or age — and measures
it end to end on LOCA.

> **Status: awaiting rebase, not review-ready as a diff.** Much of what this branch built has since
> landed on `main` independently as **`extract_llm_sweep`** (#118) and the general-fixes PR (#110). This
> branch will be **re-cut from `main`** once three follow-up PRs land (#136, #137, and a `collapse` fix),
> leaving only the parts main does not have. **Read this for the result and the argument, not the diff.**

---

## The claim

Selective, reference-aware removal **defers the blunt summary**, and that deferral pays in latency and
operational cost **even when the provider's cache has not expired** — the case a purely
expiry-triggered sweep structurally cannot reach.

## The evidence — iteration 021, paired, pre-registered

Two arms differing by **exactly one inserted component**, same frozen binary (`cg-proxy-v19`, sha256
`ecc02f28…`, code at `fcf78cd`), same 15 tasks × 5 seeds, run sequentially.

```
A  [format, toon, dedup, failed_run, cmdfilter,              extract, cachesplit, summarize]
B  [format, toon, dedup, failed_run, cmdfilter, extract_llm, extract, cachesplit, summarize]
```

| | A (shipped pipeline) | B (+ merged) | |
|---|---|---|---|
| **wall clock** | 8,640 s | **7,062 s** | **−18%** |
| per task | 115.2 s | **94.2 s** | −18% |
| **total cost** | $224.06 | $228.66 | **+2% (parity)** |
| errored runs | 17 / 75 | **10 / 75** | −7 |
| `summarize` fired on | 71% | **56%** | −15 pts |
| steps / run | 45.7 | **32.7** | −28% |
| tool calls / run | 83.5 | **61.4** | −26% |
| **exact repeats / run** | 38.6 (46.2% of calls) | **20.7 (33.7%)** | **−46%** |
| requests | 3,446 | 2,466 | −28% |
| unique tokens removed | — | 6,545,956 | |
| expand restored / unresolved | 361 / 1 | 717 / 0 | |
| CG added latency / request | 3,817 ms | 6,170 ms | +62% |

**Merged costs ~2.4 s more proxy time per request and the agent still finishes 18% faster**, because it
makes 28% fewer requests and `summarize` runs 58% less.

The repeat result is the one that most directly supports the thesis, and it **reverses** a worry held
earlier in this work: removing content was expected to make the model re-fetch. It re-fetches **less**.
The likely mechanism is that blunt summarization on 71% of requests destroys detail the model then goes
back for, while selective removal preserves what is still live.

## What iteration 021 did NOT establish

**Reward.** ITT over all 75 pairs: **14.00 → 15.00 solves, +1.00, p = 1.0000** (task-clustered over 15
clusters, which governs: 2 better, 2 worse, 11 tied; per-pair sensitivity: 7 gained, 6 harmed, 62
unchanged; harm 95% upper bound 15.2%, which does not block).

That is a **null, not a refutation**. **Eleven of fifteen tasks score 0.00 in both arms**, so the whole
comparison rests on four tasks with two moving each way — LOCA had almost no room to show a reward
difference for any configuration. Establishing reward needs benchmarks with headroom, which is the next
step rather than a claim made here.

Two further limits, both recorded in the pre-registration before the numbers existed:

- **Verdict coverage ended at 61%** (2,114 answered of 3,461 offered), and arm B ran below its own
  pre-registered 50% checkpoint for the first third of the run (amendment 2). Arm B's numbers are a
  **floor**.
- **Neither arm carried `collapse` or `mask`**, so neither is a shipped preset, and `summarize` ran
  alongside in-place offloaders against the advice in `config.go`.

## Why this branch must keep `prefix_econ`, given `extract_llm_sweep` on main

Main's sweep fires only inside a **pre-expiry window** — `remaining = CacheTTLMs − IdleMs`, positive and
under `preExpiry` — so the removal is free: the cache entry was about to be discarded anyway. That is a
better idea than pricing the break-even, because it replaces an unknowable estimate (turns remaining)
with an observation.

But it cannot fire in the case this branch is about. Arm B fired on **context pressure** — `high context
pressure` on 1,747 requests, `moderate pressure with fast growth` on 109 — with `prefix_econ` pricing
each prefix mutation and **refusing 385 of them**. `CacheTTLMs`/`IdleMs` were not in the picture at all,
so `sweeping()` would have fired on **none** of iteration 021.

So the two triggers are complementary, and the rebased branch implements both:

| trigger | the removal is | accuracy matters |
|---|---|---|
| pre-expiry window (on `main`) | **free** — the entry was expiring | barely |
| context pressure + `S·T > 11.5·W` (this branch) | **paid** — it forces a suffix re-write | **a great deal** |

That sharpens what coref is for: on the free path a wrong removal costs an entry that was expiring
regardless; on the **paid** path you have spent a real suffix re-write, so the co-reference evidence is
what makes the paid trigger worth having at all.

## What remains unique to this branch

- **`internal/coref`** — the tier-1 reference index. `main` has no index; its sweep carries an
  explicitly empty `Evidence` **seam** documented as waiting for one.
- **`components/offload/coref.go`, `corefstub.go`** — the component, and the residue reasoning.
- **`components/offload/prefix_econ.go`** — the second trigger's economics (see above).
- **The experiment record** — `docs/experiments/loca/iter007`–`iter021` and `docs/results/*`, which is
  the evidence behind every claim here.

Superseded by `main` and being dropped: `extract_llm_merged.go`, `internal/extract/bulk.go`,
`expand/restore.go` (main's `RepairToolResults` is better — it has `Unavailable()` for aged-out stashes),
`proxy/expandoffered.go`.

## The substrate measurement that started this

Reference density is a **property of the workload**, not a constant — three corpora disagree by a factor
of three, which is why the index is evidence for a decision rather than the decision itself:

| | Claude Code (interactive) | UltraHorizon | LOCA-bench |
|---|---|---|---|
| `unreferenced` | 23% | 78% | 95% |
| `closed` | 15% | 8% | **0%** |
| `open` | 60% | 13% | 4% |
| …restricted to ≥20 later turns | 21% | 70% | 70% |

## Measurement defects found along the way

Eight, in one component's measurement path — recorded because **every one produced the identical
misleading signal, "the model declines to act"**, and four of them predate this branch:

| defect | effect |
|---|---|
| coref prefix pre-filter | removed 149,681 candidates, leaving ~1 per call |
| `llm_max_per_request` | caps candidates in a one-call design |
| economic gate | prices each candidate against a whole call; suppressed 1,497 vs 224 through |
| any pinned floor disables the pressure trigger | fired on 72% of requests regardless of context ([#134](https://github.com/rossoctl/context-guru/issues/134)) |
| `verdicts ÷ calls` used as batch size | counts what was ANSWERED, not OFFERED |
| 2048 output-token ceiling | reply cut mid-array on ~70% of calls |
| `tool_choice: none` | drove the model into **prose** — 0 of 6 verdicts (#137) |
| rig: marker check used the unescaped spelling | `has_marker` read 0.0% on every arm ever run |

The judgement machinery was never the problem. A sampled reply shows the model reasoning correctly under
the criterion and simply saying so in sentences — which the contract itself calls a valid answer.

## Next

1. Land #136, #137 and the `collapse` fix on `main`.
2. Re-cut this branch from `main`; port coref, fill the `Evidence` seam, implement the second trigger.
3. Re-run on LOCA **and at least one benchmark with reward headroom** — the axis iteration 021 could not
   speak to.


docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+
+## 1. Why today's relevance signal cannot do this
+
+Every LLM component gets its relevance signal from `conversationGoal`
+([`components/offload/common.go`](https://github.com/rossoctl/context-guru/blob/main/components/offload/common.go)): the first user turn
+(the task), plus the most recent assistant and user turns (current intent). Tool outputs are
+deliberately excluded — they are the mass being reduced, not the goal.
+
+That signal is **forward-looking and position-free**. It answers "what is the agent trying to
+do", never "which earlier span does this turn point back at". Co-reference is therefore not a
```
amiddavid: pls explain this sentence, perhaps add an example

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+tuning change to an existing input; it is a new input, and it is the only input that can
+justify dropping a *large*, *early* span rather than projecting a recent one.
+
+The deterministic projector (`internal/extract/deterministic.go`) has the adjacent primitives
```
amiddavid: pls explain this paragraph in more details, I find it hard to follow, especially for a proposal document

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+extraction is *contained in* its original. Point containment the other way (is this span of the
+tool result contained in a **later** message?) and you have the beginning of a reference index.
+
+## 2. What a reference is, in three tiers
+
+Ordered by how deterministically it can be detected:
+
+| Tier | Signal | Detectable |
+|---|---|---|
+| **1** | `tool_use_id` ↔ `tool_result` pairing; and **literal carry-over** — a span introduced by tool result *i* reappearing verbatim in a later `tool_use` argument or assistant text (paths, symbols, line numbers, IDs, hashes, error strings) | exact, zero LLM |
```
amiddavid: add an example column

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+surviving copy. Score on `(reference count, how long ago the last reference was, witness
+present)` — with recency measured **from the head of the transcript**, not from the output's own
+position, since "recent messages vs early messages" is a statement about now.
+
+And the witness turns out to be free. `coref` only ever cuts **tool outputs**; references live
+in **assistant** turns, which are never cut. So any reference at all *is* a surviving copy —
+"the model referred back to this" and "the value it took still exists in the request" are the
+same fact. That is what makes the closed case cheap to establish rather than a second search.
+
+Framed this way, `coref` is `dedup` generalized: from "this tool output is byte-identical to
```
amiddavid, line 212: I'm not sure I fully agree. consider this scenario:
a tool output returned: { "name": "david", "id": 123, "address": "foobarbaz"}
, { "name": "osher", "id": 235, "address": "banana"} the agent said, I need to remember david 123 address.
the address itself wasn't coref, but the tool output is needed and cannot be removed.

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+another" to "this tool output's useful content survives elsewhere in the request".
+
+## 4. The economics, and why they reshape the design
+
+This is where the proposal has to survive contact with what the repo already measured
+([improvement plan §0 and §C](../results/improvement-plan.md)).
```
amiddavid, line 218: I dont see improvement-plan in the docs

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+taken still exists in the request". The witness needs no separate search.
+
+### Validated against a known-answer fixture
+
+[`deploy/harbor/coref_fixture.py`](https://github.com/rossoctl/context-guru/blob/main/deploy/harbor/coref_fixture.py) builds four tool
+outputs whose correct classification is fixed by construction — one closed, two unreferenced,
+one open — including the echo confound from §2 (a `Read(src/config.py)` whose only later
+overlap is the path that arrived as the `tool_use` argument). All four classify correctly.
+
+The **negative control** is the part worth keeping: with the echo-exclusion guard disabled, the
```
amiddavid: what is the echo-exclusion guard, pls elaborate on it

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+compactor cuts something load-bearing.
+
+**Decision rules this pass feeds:**
+
+- If unreferenced mass is large and concentrated, the first version is deterministic and needs
+  no model call.
+- If referenced mass is dominated by low reference counts with distant last-references, the
+  closed-case predicate is worth building.
+- If the break-even table says no realistic `T` clears `11.5 × W` even when batched, the
+  component is a step-reduction and agent-compaction-deferral play only, and must be evaluated
```
amiddavid, line 576: overall while we measure cost, any expermient we do should measure the reward as well.

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+| **Terminal-Bench 2.0** (already wired) | A **different cost regime** — output tokens are 47% of the bill, larger than cache-read. Must be tuned separately, and it is where a step-reduction claim is won or lost |
+| **LOCA-bench** (MIT, native `anthropic` SDK + `LOCA_ANTHROPIC_BASE_URL` → direct attach) | The **controlled instrument**: context length is a dial (8K→256K) with fixed task semantics, deterministic binary scoring, and built-in `memory_tool` / `ptc` / context-editing arms — the naive-compaction baselines to beat. But its BigQuery/Sheets/Snowflake domains *aggregate and compute over* tool results, so references arrive transformed: it is a **Tier-2/3 stress test**, not a showcase for exact matching. Note its native trimmer orphans `tool_use`/`tool_result` pairs at 64K and provokes provider 400s — a bug class our byte-lossless splice and reversibility are designed to avoid, and worth claiming |
```
amiddavid: we should probably copy the fixes we introduced in forever

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+| **SlopCodeBench** | Resets per checkpoint, so sessions never approach the threshold. Structurally cannot test this |
+
+**Acceptance criteria** (in priority order, from §4): reward parity or better; **steps** down;
+**cache-write** within a stated budget rather than "unchanged"; `expand` rate below a
+pre-registered ceiling; billed cost down. And two methodology guards the corpus has already
+paid for: do not stop at first significance (`p = 0.036` at `n ≈ 22` regressed to `p = 0.22` at
+`n = 30`), and prevent-and-measure rather than filter-after-the-fact (dropping anomalous runs
+introduced survivorship bias when the failure rate was arm-imbalanced).
+
+## 9. Implementation status
```
amiddavid: this should be in its own document, not part of the proposal, unless this is how it is done in other proposals in this repo

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+component against numbers nobody has — so the implementation is scoped to the part that has no
+calibration in it. `cut_unreferenced` needs no threshold: "no later turn used anything this output
+introduced" is a fact about the transcript. `cut_closed` needs two (`closed_dist`, `open_reps`), which
+are precisely what §7 produces, so it defaults to **off** and the shipped values are placeholders
+carried over from `coref.py`'s defaults for comparability, not recommendations. `coref` is in **no
+preset** for the same reason.
+
+Also not built: the Tier-2 LLM escalation (§10), and the incremental per-session reference index. The
+index is currently recomputed per firing turn — acceptable because the trigger makes firings rare, but
+it is the latency question in §10 and it is unmeasured.
```
amiddavid: pardon my ignorance, when u refer to things like 10 here, where is that referring to?

docs/proposals/coref-compaction.md
```diff
@@ -0,0 +1,430 @@
...
+then left alone.
+
+Four things it settles, and one it overturns:
+
+- **`cut_unreferenced` is justified as the default** — 21% of mass on interactive traffic and ~70% on
+  benchmark traffic, with no calibrated threshold and no model call. Decision rule one from §7 is
+  answered yes on every corpus.
+- **A reference consumes a median 18.7% of what its output introduced** (11.5% on UltraHorizon).
+  Hypothesis A — "took one value, does not need the rest" — is confirmed rather than assumed.
+- **Tier-2 leakage is 2%** of model turns (a stated numeric absent from all prior context) — real, and
```
amiddavid: can u explain this more, I'm not following

docs/proposals/coref-compaction.md
amiddavid: ahh its in the glossary... but this document should be self-contained as u start reading from here :) , right ?

docs/proposals/coref-compaction.md
amiddavid: i.e. doesnt this contradicts case B?

docs/reference/coref-glossary.md
```diff
@@ -0,0 +1,150 @@
...
+The original framing separated A from B by **distance** (B = recent, A = early). The measurement says
+distance barely works — see `closed_dist` below.
+
+## 3. The three verdicts (what the classifier outputs)
+
+For each tool output, exactly one of:
+
+| Verdict | Means | Cut it? |
+|---|---|---|
+| **`unreferenced`** | No later turn ever used anything this output introduced. | **Yes — the free cut.** No threshold needed, no model call. This is the shipped default (`cut_unreferenced`). |
```
amiddavid: if it a recent turn, it might not had the chance to be referenced, dont we need to guard from this ?

docs/reference/coref-glossary.md
```diff
@@ -0,0 +1,150 @@
...
+
+**Why "closed" is cheap to establish:** `coref` only ever cuts *tool outputs*, and references live in
+*model turns*, which it never cuts. So "a later turn referred back to this" and "the value it took still
+exists in the request" are the same fact — the surviving copy (the **witness**) needs no separate search.
+
+## 4. The two thresholds that decide `closed` vs `open`
+
+| Knob | Default | Means | Verdict from the data |
+|---|---|---|---|
+| **`closed_dist`** | 12 | How many messages **ago** the last reference must be before the output counts as `closed`. Newer than this ⇒ `open`. | **Nearly inert.** A 10× sweep (4→40) moves the answer 2–3 points. Don't tune it. |
```
amiddavid: dont tune it, but still matters?

docs/reference/coref-glossary.md
```diff
@@ -0,0 +1,150 @@
...
+
+| Term | Means |
+|---|---|
+| **batching** | One rewrite must serve **every** cut in the pass, so `S` is the sum of all of them. That's what makes break-even reachable (60k of a 150k transcript needs `T` > 23). Hence a rare, threshold-triggered pass — never a per-output, per-turn decision. |
+| **`min_batch_frac`** | The operational form of that: the pass must cut at least this fraction of the request (default 0.15) or it declines and leaves the request byte-identical. |
+| **`rewrite_budget`** | Prefix-rewrite passes allowed per session (default 3). `coref` is the only component that spends cache-writes **on purpose**, so the spend is capped and reported. |
+| **step reduction** | The real prize. `corr(Δsteps, Δcost) = +0.95`; unique token removal is ~0.02% of the bill. The objective is **steps and reward, not bytes**. |
+| **deferring agent compaction** | Claude Code compacts itself at ~167k on a 200k model. Staying under that avoids a full-transcript summarization — a large cache event *and* a quality loss. Plausibly the biggest win. |
+
+**The counter-intuitive consequence:** firing at 90% of the context window means `T` ≈ 0 — paying a
```
amiddavid, line 150: love this. 💌 
though it depends how much is being cut isn't it? and for the determinsitic one, its cheap to calculate and can tell u how much deferring is happening.

I would also appreciate some thought on what it means for larger context windows that are now more and more frequent.... up to 1M

docs/reference/coref-glossary.md
```diff
@@ -0,0 +1,150 @@
...
+
+| Term | Means |
+|---|---|
+| **latching** | The decision is stored per session and **replayed byte-for-byte** thereafter, never re-derived. A co-reference decision depends on *history*, so re-deriving it against a longer transcript could emit different bytes — which is exactly the prefix flip that costs a second cache-write. |
+| **one-way / monotonic** | Keep → cut only. New evidence can never un-cut, because un-cutting is another rewrite. Monotonicity is a cost requirement, not tidiness. |
+| **`freeze` / `reapplyFrozen`** | The mechanism that does it: record the replacement text against the original's content hash, and replay it on every later turn at any depth. |
+| **`TailOnly`** | The rule every *other* age-based offloader follows: never touch the already-cached prefix. `coref` deliberately violates it — that's its purpose — which is why the spend is budgeted. |
+| **`repairLostFreeze`** | A repair `mask`/`failed_run` may use: re-derive a lost decision at depth, safe because their output is a pure function of `(content, config)`. **`coref` must never use it** — its decision is history-dependent, so re-deriving is the very byte-flip the repair exists to prevent. |
```
amiddavid: since I'm not familiar with this repo yet, I would appreciate if you can in a comment explain this more

docs/proposals/coref-compaction.md
amiddavid: Rewritten with a worked example rather than the assertion. It now walks turn 4 reads `src/auth.py` / turn 5 says "the bug is `TOKEN_GRACE_SECONDS`" / thirty turns later the agent is on tests — and shows that asked "is the turn-4 output still needed?", `conversationGoal` can only answer "the task is still about auth", which is true of *every* output and so decides nothing. The fact that settles it (the one value taken sits in turn 5, and turn 5 isn't going anywhere) is positional and backward-looking, which that signal cannot represent at all.

docs/proposals/coref-compaction.md
amiddavid: Expanded into a bulleted walk-through of the two existing pieces and what each contributes: `deterministic.go`'s important-key list is already an answer to "which parts of an output would a model carry forward?", and `contain.go` today checks a shrunken output is a subset of its original. The reusable idea is the second one run **backwards** — today it asks "is this compacted text contained in the original?", inverted it asks "is this span of the original contained in a *later message*?", and the same primitive becomes a reference detector. Same test, opposite direction: one validates a rewrite, the other measures reuse.

docs/proposals/coref-compaction.md
amiddavid: Added an EXAMPLE column. Same reference at each tier: Tier 1 `TOKEN_GRACE_SECONDS = 0` reappearing verbatim in an `Edit` argument; Tier 2 `[{"ms":1200},{"ms":1800}]` → "total latency is 3 seconds" (the `3` appears nowhere — it was computed); Tier 3 a directory listing → "as I saw earlier, the tests live beside the source", which is unmistakable to a reader and shares **no token at all**.

docs/proposals/coref-compaction.md
amiddavid: Fixed — the guard is now named and explained inline at that point rather than assuming the glossary: only tokens the output *introduced* are eligible, so anything already present at or before the producing tool call is discarded as an echo of context that predates the output.

And yes, you're right about the principle: the proposal should be self-contained from the top. The glossary is meant as a lookup for the vocabulary once it's spread across four documents, not as a prerequisite for reading the argument. I've gone through the other places review flagged the same way.

docs/proposals/coref-compaction.md
amiddavid: Done, and made concrete rather than abstract. §8 now points at the actual fix to port: `repair_tool_pairing()` in forever's `forever/benchmarks/_anthropic_auth_hop.py`, which solves it rig-side in two phases — drop `tool_result` blocks whose `tool_use` was trimmed away (plus any message left empty by that), then synthesize placeholder results for `tool_use` blocks left unanswered — and counts the repairs so the rate is visible instead of silent.

Also worth separating: `coref` cannot *cause* that bug. It rewrites a tool message's text in place and never removes a message, so pairing is structurally preserved. `summarize` is the one that restructures the list.

docs/proposals/coref-compaction.md
amiddavid: You're right, and there's no existing convention to defer to — `docs/proposals/` is new in this PR, so this is the first one. Split out to **`docs/proposals/coref-implementation.md`**, linked from the proposal with a short pointer.

The reason is the one you're implying: implementation status goes stale on every commit while the argument doesn't, and a proposal that doubles as a changelog stops being reviewable as a proposal.

docs/proposals/coref-compaction.md
amiddavid: Not ignorance — that was genuinely opaque. `§10` meant "section 10 of this same document" (Open questions), which is both unhelpful and fragile, as splitting the doc immediately proved by making the number wrong. All bare section-number cross-references are named links now ("the latency question in [open questions]"). Where a number appears in prose it should now always be a §-plus-title or a link.

docs/proposals/coref-compaction.md
amiddavid: Expanded. The short version: Tier 2 is a reference that arrived *transformed*, so by definition no substring match can find it — what's countable is a **symptom**. If a model turn states a numeric value appearing nowhere in any earlier message, it computed that number from something, and that something was almost certainly a tool output. 2% of turns look like that on interactive traffic, which is why a zero-LLM first version is viable.

Two caveats now stated, and the second is a self-inflicted one worth knowing: it's a **lower** bound (only numeric transformations leave this trace — reworded prose is invisible), and tightening the identifier rules also **blinded the proxy**, since bare numbers now need 5+ digits and most computed values are small. So its 0% on LOCA means "none among tokens the tokenizer still accepts", not "none" — on a corpus with 0% `closed` and 40% `opaque`, the honest reading is that Tier-2 references there are common and simply unmeasured.

docs/reference/coref-glossary.md
amiddavid: Good catch, and no — there was no guard, which was a real gap. An output near the tail has had no chance to be referenced, so scoring it as unused would make a batched pass preferentially cut the **most recent** context, which is the worst possible choice. `mask` avoids this with `keep_recent`; `coref` had nothing.

Added `min_later_turns` (default 8): an output with fewer model turns after it is treated as `open` regardless of everything else. Worth noting what the state was before — the *measurement* had bounded this bias (LOCA's raw 95% fell to 70% when restricted to outputs with 20+ later turns) but nothing in the *component* guarded against it. Bounding a bias in a report is not the same as not having it in the code.

docs/reference/coref-glossary.md
amiddavid: Fair — that phrasing was self-contradictory. Rewritten to what's actually true: `closed_dist` is **load-bearing but flat**. Set it to 0 and the `closed` class stops existing, so it certainly matters; but anywhere in 4–40 gives the same answer within 2–3 points, so there's no *return* on tuning it. Leave it at the default and spend the effort on `open_reps`, which moves the answer 18 points across the same kind of range.

docs/reference/coref-glossary.md
amiddavid: Expanded both from first principles rather than by name.

**`TailOnly`** is a helper on `Ctx` answering "may I safely modify the message at index *i*?" It returns false for anything the provider has already cached, because editing cached content breaks the prefix hash and forces a cache-write of everything after it. Every other age-based offloader (`mask`, `failed_run`, `collapse`) consults it and declines. `coref` deliberately ignores it — reaching into the cached prefix *is* the point, since by the time a session crosses the threshold all the mass is back there — which is exactly why its spend is budgeted rather than forbidden.

**`repairLostFreeze`** needs the background first: an offloader `freeze`s its replacement text against the original's content hash and replays it every turn so the bytes stay stable. If the store *drops* that record (TTL, eviction), it would normally decline to act at depth — but then the message reverts to full text, which is itself a prefix change. So `mask` and `failed_run` may re-derive even deep in the prefix: their replacement is a pure function of `(content, config)`, so re-deriving reproduces byte-for-byte what the provider already cached. `coref` must never do this, because its decision depends on the whole transcript — re-deriving against a longer one can yield a different class and different bytes, the precise flip the repair exists to prevent.

docs/proposals/coref-compaction.md
amiddavid: You're right, and this was the most valuable comment on the PR — it found a real bug, not just a wording problem.

I ran your exact example through the index rather than reasoning about it, and it's worse than you flagged. `david`, `123`, `foobarbaz` are short lowercase words and a 3-digit number — precisely what the precision rules in §2 exclude — so the output yields **zero** trackable tokens. Zero novel tokens means zero references, which scored `unreferenced`, **which is the class the default config cuts.** So the shipped default would have deleted that output while the agent was still asking for the address.

Two separate defects, both now fixed in e7a2623:

1. **Your conceptual point.** "Any reference is a surviving copy" is too strong. The model referenced an **anchor** (`david`, `123`) in order to point at a payload (`foobarbaz`) it never restated. An exact matcher can't distinguish an anchor reference from a payload reference — so `closed` can't rest on "referenced once, long ago" alone. That is now stated as *the* reason `cut_closed` ships off, rather than mere caution. It also inverts my §7 reading of `used_frac`: a **low** value is ambiguous, not evidence for case A, because "took the value, rest is chaff" and "took an anchor, still needs the payload" look identical.
2. **The concrete one.** `refs == 0` conflated two opposite states — "introduced 200 identifiers, nobody touched one" (evidence of deadness) and "introduced nothing I can see" (absence of evidence). There's now an `opaque` class that is **never cut at any setting**. It is not a corner case: **8%** of tool-output mass on interactive traffic, **20%** on UltraHorizon, **40%** on LOCA-bench — the last being 11 outputs averaging 22k tokens of exactly the record-dump shape you described.

Re-measuring dropped the headline `unreferenced` figures from 23/78/95% to **13/51/22%**, and break-even from 15/30 to 9/30 sessions. Your counter-example is now a test case on both sides of the implementation.

docs/proposals/coref-compaction.md
amiddavid: `docs/results/improvement-plan.md` does exist on `main` (verified with `git cat-file -e main:docs/results/improvement-plan.md`) and is in the mkdocs nav, so the link resolves on the published site. It's just not in this PR's diff, so GitHub can't render it as a clickable target here.

docs/proposals/coref-compaction.md
amiddavid: Agreed, and this is a real gap in how §7 was framed. Added explicitly: **every decision rule in that section is about cost, and cost alone can never authorize shipping this.** A compactor that saves tokens and loses one task is a regression, because the value of the context it removed was never denominated in tokens. So reward is a **gate, not a metric**.

One honest limitation now stated there too: this particular measurement *cannot* speak to reward by construction — it reads traffic that already happened and asks what a cut would have removed. Nothing about outcomes. §8's acceptance criteria were already ordered with reward parity first; the fix is that §7 no longer reads as though cost were sufficient.

docs/reference/coref-glossary.md
amiddavid: Thanks — and both of your points landed in the doc.

**On "it depends how much is being cut":** yes, and more sharply than I'd written it. The agent-compaction prize is a **step function**, not a slope — you either drop below the threshold or you don't, and cutting 90% of what was needed to get there is worth nothing. Which argues for sizing the batch against the *threshold distance*, something `min_batch_frac` cannot currently express. Noted as a limitation.

**On deterministic measurement:** agreed, and it's the cheapest real metric available here — compare the API-reported usage against the documented compaction threshold and count the turns of headroom the cut bought. No benchmark scoring, no seeds, no LLM judge. It isn't in the metrics yet; it should be.

**On 1M windows** — I worked this through and the answer surprised me. **Break-even is scale-invariant.** Rearranged, `S × T > 11.5 × W` is `T > 11.5 × (W/S)` — it depends only on the *ratio* of rewritten suffix to cut mass, never on absolute size. A 1M transcript with the same density of cuttable mass needs the same `T`. So a bigger window neither rescues nor damns the token economics; it only moves *when* the trigger fires. What improves the ratio is cutting a larger share of what lies *after* the shallowest cut — an argument for cutting deep and rarely, not for cutting more.

Three things do genuinely change, now a table in §7 of the cheat sheet: cache-read becomes the entire bill (so `coref` is a **cost** play at 1M rather than a fit play — the strongest argument for it there); the agent's own compaction recedes to ~967k, making that prize **rarer but much larger**; and the index cost scales linearly, so an incremental per-session index stops being an optimization and becomes a requirement.

amiddavid:
Replaced by #186 (feat(offload): extract_llm_sweep — co-reference index, economic trigger, and the fixes that let it act), which was merged into main. Closing this in favor of that PR.

PR #86, opened 2026-08-22
feat(cache): keep an idle prompt cache alive, price the 1h write tier, and say what cachesplit did

OsherElhadad:
## The problem, in one table

Measured on the hosted service: 19,805 requests, **$3,139.97**, 2026-08-16 → 21.

| miss reason | n | share | $ | share of spend | $/req |
|---|---:|---:|---:|---:|---:|
| hit | 16,368 | 82.6% | 1,928.65 | 61.4% | 0.1178 |
| **ttl_expiry** | **742** | **3.7%** | **741.07** | **23.6%** | **0.9987** |
| cold_start | 1,626 | 8.2% | 253.82 | 8.1% | 0.1561 |
| prefix_change | 292 | 1.5% | 205.05 | 6.5% | 0.7022 |
| unknown | 777 | 3.9% | 11.38 | 0.4% | 0.0146 |

A TTL-expired request costs **8.5x** a request that hit. Against each row's own
counterfactual — the same tokens billed as a read — the penalty is **11.35x**, of which
91.2% is pure re-write. The addressable population is 357 claude-cli requests / $734.75.

**$584.83 — 18.6% of ALL spend — sits in misses whose idle gap was under an hour.** For
scale: the entire transformation pipeline's measured net over the same window is **$9.72**.

## The caching math, with the doc quotes it rests on

> By default, the cache has a 5-minute lifetime. **The cache is refreshed for no
> additional cost each time the cached content is used.**

> **The lifetime is measured from the start of the request that writes or reads the cache
> entry, not from the end of its response.** Time spent generating a response counts
> against the lifetime.

> **Cache writes happen only at your breakpoint.** … Because the hash is cumulative,
> covering everything up to and including the breakpoint, changing any block at or before
> the breakpoint produces a different hash on the next request.

> **Exact matching:** Cache hits require 100% identical prompt segments.

Multipliers: 5m write **1.25x** base input, 1h write **2.0x**, read **0.1x**. Backed out of
production billing independently and confirmed to the cent — read **0.1000** at every prefix
size, write **1.250**, so write/read is **12.5x**.

Converting one miss recovers `(1.25 − 0.1) = 1.15x` on the prefix; one ping costs `0.1x`.
**Saving : ping = 11.5 : 1**, and both sides scale with the model's input rate, so the ratio
survives the whole price list.

---

## 1. The idle keep-alive (`proxy/keepalive.go`)

Host-level, not a component, because the expensive event happens when **no request is in
[...]

PR #93, opened 2026-08-22
feat(dash): show the prompt behind every number, and make the Inventory tab readable

OsherElhadad:
## What the owner reported, and which of the three kinds each was

The owner used the shipped Inventory tab and said four things. Serving the dashboard against
a scrubbed copy of the production snapshot and opening it settled which was which — none of
the four turned out to be the asset-cache bug.

| Reported | Actually |
|---|---|
| "I cannot see the commands for each tool/skill for how to remove them" | **Genuinely missing.** `removalCell()` existed and was called from exactly ONE place: the built-ins table (`tools.js:806`) — the only group on the page nobody should act on. The actionable list, 80 rows of it, rendered no command at all. |
| "separate between built in tools to others that can and should be changed" | **A bug, not a design gap.** `KIND_LABEL` mapped kind `tool` to the literal string **"built-in"**. Claude Code's own tools and a third-party agent's own tools are the *same* kind — the answer lives in a separate `builtin` boolean — so `DesignSync`, `ScheduleWakeup`, `ReportFindings`, `PushNotification` and every other removable client tool rendered a pill reading "built-in" inside the list the page was recommending be removed. |
| "I cannot tell what the statistics mean" | **Genuinely missing.** Every Overview tile had a `TILE_INFO` entry. Not one of the eleven Inventory tiles did. The one tab whose figures authorise deleting things was the one tab that never said what its figures meant. |
| "much more visualized and easy to understand" | **Design.** Addressed below. |

The built-ins section, the `TILE_INFO` mechanism, the `Removal` API, `SelfRemoval`, the
savings decomposition, the prompt-cache panel and the `EventSource` refresh were all verified
present and working. They were not the problem.

## The genuinely missing feature: prompt text

Nothing stored prompt text, so the page could say a capability costs 4,723 tokens a request
and could not show the 4,723 tokens — a figure nobody removes a tool schema on.

Every region of the prefix is now captured with its text: each tool's JSON element, each
skill's listing entry, the listing itself, and the **system prompt** on a new
`kind='system_prompt'` marker row. `GET /api/prompt` serves one session's prefix decomposed
into regions with their weights and shares; the UI shows each capability's own slice behind a
reveal on its row, and the whole prefix region by region behind one on the panel.

**How consent gates it.** The text is the one CONTENT field in a table that otherwise holds
identifiers, so it rides the transcript-capture consent pair — the operator's
`--dashboard-content` **and** the tenant's own opt-in — carried per message on
`RecordInventory(..., text bool)` from `proxy.captureContentFor`. Without it the writer stores
`NULL` and every measurement survives: weights, names and usage counts are identifiers and
operational metrics, so an account that declined transcript capture keeps the whole inventory
feature and loses only the text. The gate is applied on the WRITER, not at the scan, because
`declCache` entries are shared by every tenant declaring the same set and gating there would
poison them.

Four states are told apart, and the UI never shows an empty panel that looks broken:

- text present → the text, with its weight and share of that session's prefix
- **operator** gate shut → "not stored on this deployment … there is nothing for you to enable"
- **tenant** gate shut → "not captured — enable transcript storage to see this", and that it does not backfill
- row predates the column → "not recorded yet", from an explicit coverage count

**No `schemaVersion` bump.** One additive `text_gz BLOB` column with a NULL default, so the
20,074 production rows survive the next restart. Scrubbed with the existing redactor and
capped at 64 KiB on the writer goroutine; system-prompt rows capped at 4 per session because
the text is client-supplied and may carry a clock.

## The design work on Inventory, and the reasoning

**The actionable list is grouped by what ONE action removes.** Flat and sorted by weight,
eighty rows of `$0.0035` read as eighty separate judgement calls with the two that mattered
somewhere in the middle. An MCP server is one decision, not eight. Grouping happens **only
when the server says the mechanism really is group-wide**: a plugin-bundled MCP server has no
`claude mcp remove` name and its removal is a per-tool deny, so those stay ungrouped rather
than sit under a heading claiming a one-tool snippet "removes all 2 of them". Groups sort by
the dollar figure they lead with — not by token weight, which put a 314-token group that cost
$0.04 above a 295-token one that cost $0.13 — and past the eighth the tail folds with its
total stated on the summary. **The page is now shorter than before** (9.1k px against 10.8k)
while carrying the commands, the prompt panel and a text reveal per row.

**The chart form was checked, not assumed** (`dataviz` skill, form heuristic first). Part-to-
whole with long-named categories → horizontal stacked bar, which is what was already there. So
the bar stayed and the *colour* changed: the four groups a reader can act on take the four
categorical slots and the group they must not act on takes a new de-emphasis grey
(`--s-mute`). Colour is what the eye reads before any label, and four hues of equal weight
asserted that Claude Code's own equipment was a fifth removable category — the reading that
gets somebody to delete `Read`. That is the owner's "separate built-ins from the others"
rendered rather than annotated. Segments gained the mark spec's 2px surface gap.

**The system prompt is in the composition**, because it is normally the largest single region
of the prefix and leaving it out made a "composition of your prompt" that was a composition of
the tools array.

**The built-ins danger moved onto the CLOSED bar.** It used to live only inside the expanded
panel, so a first-time reader saw a plain grey strip indistinguishable from the other collapsed
panels. A danger only visible after you open the door is decoration.

## Three arithmetic and copy corrections found on the way

- The composition added each **skill's** weight to the **listing's**, but a skill's weight *is*
  its entry inside that listing. Segments summed to 43,933 while the panel printed 36,950
  beside them as "the whole that each share is a part of", and every legend percentage was a
  share of the inflated figure.
- The prompt coverage count was raw table rows (4,192) against the report's declarations (309)
  — 13x apart on one page with nothing connecting them. Both now count one per
  `(session, kind, name, server)`.
- Both **session-length multipliers** now name their population and the basis panel reconciles
  the pair (plain mean over tool-bearing sessions vs request-weighted mean over captured ones;
  3.9 against 150.6 on production traffic). And the copy asserting "most sessions are
  one-request sidechains" is now conditional on the median actually being small — printed
  unconditionally it told a reader whose median was 244 requests that most of their sessions
  were single calls.
- A `Copy` button was nested inside a `<label>`, so clicking it toggled the checkbox beside it.

## Unreported savings

Re-checked every measured column against what reaches the UI. Everything measured is reported
somewhere, or is one of the three known deliberate omissions (`cache_saved_usd`,
`wasted_tokens`/`bounces`, `filtered_decl_tokens` — measured `SUM` still 0 across all 19,805
requests). The one new measured figure this PR adds — the system prompt's weight — is reported
on the Inventory tab.

**One gap found and deliberately not closed:** Overview's 40-odd tiles never signpost the
Inventory tab or its projected avoidable figure, so a reader of Overview cannot see the largest
lever. I did not add it, because the Inventory figure is a *projection* and Overview is an
account of *realized* savings — folding one into the other is the flattering mistake honesty
rule 2 exists to forbid. It is a discoverability problem, not a reporting one, and worth
solving as a link rather than a number.

## Tests

`go vet` clean, `gofmt -l .` empty. `CGO_ENABLED=1 go test -race -count=1 ./...` — every package
green, including `./proxy/`; the `./dash/` run is being repeated on its own because I terminated
its process by hand mid-run while cleaning up a stray one, and a `signal: terminated` is not a
result worth reporting either way.

Keyboard traversal was checked with real `Tab` presses over the redesigned page: all 9
distinct control kinds reachable by keyboard carry a visible focus ring, and a `Copy` click is
verified not to toggle the checkbox beside it (it used to — the button was nested in a
`<label>`).

New Go tests cover the consent gate (rows written, text `NULL`, weights intact), the coverage-
count path, the system-prompt row not being counted as a removable declaration, redaction and
the size cap, the per-session row bound, and the system prompt being recorded for every
declaration set a session presents — that last one was a real defect found by writing the test:
the writer deduped per session while the reader selects per digest, so a session that changed
its tool set could render a prefix with its largest region silently missing. Verified to FAIL
against the per-session key.

The scan's added hot-path cost is measured rather than assumed, with two new sub-benchmarks:
the system-prompt scan is 22 us warm against the whole warm scan's 212 us, and 208 us cold
against its 1.18 ms — about a tenth added to a scan that is already a small fraction of a
request, paying the tokenizer once per session rather than once per turn. The existing "cold"
benchmark was also cleared of a warm system memo, which had been understating it. Three static checks over the embedded UI assets
guard the defects that shipped and looked fine on screen — `KIND_LABEL` claiming a kind is a
built-in, an Inventory tile with no `TILE_INFO` entry, and the removal command being absent
from the actionable list. All three were verified to FAIL when the original bug is reintroduced.

## Screenshots

Local only — the repo is public and the captures are of real accounts' inventories, so per the
owner's ruling they are not committed. `/tmp/r2/shots/` holds before/after pairs for Overview
and Inventory in light and dark, plus the two not-captured states; `/tmp/r2/slices/` and
`/tmp/r2/slices2/` hold the before and after Inventory pages sliced for panel-by-panel reading.

They were taken against a **copy of a copy** of the production snapshot, scrubbed to the two
authorised accounts with every client session id rewritten to a synthetic one. The capture
driver asserts that no email, no UUID and no unexpected 16-hex account id appears in the
rendered text of any page it shoots and fails the run if one does; the final run reports
`PRIVACY CLEAN, NO JS ERRORS` across all twelve captures.


OsherElhadad:
## Review — `25ba7d3`

**CHANGES-REQUIRED**, on finding 1 alone. 2 and 3 are copy corrections that should ride along.

I served the dashboard on loopback and drove the page in a real browser, so the "both bugs fixed" judgement and findings 3 and 8 come from the rendered page rather than from reading the source. `CGO_ENABLED=1 go test -race -count=1 ./...` green (25 packages, exit 0, no data races; the two known #91 flakes did not fire, and I re-ran `./dash/` and `./proxy/` alone for them). `go vet ./...` clean, `gofmt -l .` empty.

Figures measured against production data are in the local review file rather than here, since this repo is public.

---

### 1. HIGH — `/api/prompt` serves prompt text without the content-visibility gate every sibling content route applies

`dash/promptapi.go:78-95`.

In single-tenant mode (`a.auth == nil`) `a.scope(r)` returns `ok` with `Manager: true` for **any** caller that can reach the dashboard port, and `a.prompt` then writes `PromptRegion.Text` — the whole system prompt and every tool schema — straight out.

The sibling content routes do not. They compute `a.trusted(r)` (loopback, or a `--dashboard-trusted-cidrs` entry) and pass it into the query:

- `dash/api.go:692-700` → `DB.Request(id, trusted)`
- `dash/api.go:317-322` → `DB.SessionEvents(f, session, visible)`

Measured, not inferred — a throwaway test in package `dash`: single-tenant, `CaptureContent: true`, no trusted CIDRs, `RemoteAddr = 10.9.9.9:1234`, synthetic markers:

```
/api/requests/{id} leaks transcript text: false
/api/prompt        leaks system-prompt text: true
/api/prompt body:  {"captured":true,...,"regions":[{"kind":"system_prompt",...,
                    "text":"SYSTEM-PROMPT-SECRET-MARKER","has_text":true},
                    {"kind":"tool","name":"Read",...,"text":"{\"description\":...
```

The documented model says the opposite, in three places this PR did not update:

- `docs/reference/config.md:184` — "CIDRs allowed to view per-request **content** and the effective config. Loopback always is; **aggregates are open to everyone**."
- `docs/dashboard.md:720` — "Per-request **content** (the diff view) | loopback, or an explicit `--dashboard-trusted-cidrs` entry"
- `docs/reference/routes.md:343` — the same table.

`docs/dashboard.md` *is* edited in this PR (line 692) without the visibility table a few lines below it being touched. And the promptapi header says this route "reports the same three states the diff view does" — it reports the same states, but does not apply the same visibility rule.

**Fix, a few lines:** mirror `a.request`. `trusted := a.trusted(r)`; `if a.auth != nil { trusted = true }` (hosted: ownership is the gate); when `!trusted`, clear `Text`/`HasText` on every region and leave the weights, `Rows`/`TextRows` and `BlockedBy` alone. The whole design survives — weights served, text gated — and the panel already renders that state honestly.

**Why nothing caught it.** The scoping test *does* walk this route (`routes()` ends with `append(rs, a.toolRoutes()...)`, so that comment is accurate), but `TestAPIScopesEveryRouteToTheCaller` asserts tenant boundaries, not the CIDR content gate, and nothing exercises `a.prompt` as an HTTP handler at all — the seven new tests all sit at `PromptViewFor` and the writer, i.e. the layers below it. A route-level assertion over the same table ("no route returns content text from an untrusted address in single-tenant mode") would catch this and the next one.

### 2. MEDIUM — the identical unconditional "one-request sidechains" claim survives in the sibling copy

`dash/ui/tools.js:430` (`renderCoverage`), against the fix at `dash/ui/tools.js:1146` (`renderRemovalValue`).

The conditional version carries a comment stating exactly why the unconditional print is wrong — that printed unconditionally it told a reader whose median was 244 requests that most of their sessions were single calls, "which is both false and a very odd thing for a page to say about the reader's own history". Seven hundred lines up, **on the same page**, the same sentence is still unconditional:

```js
+ (rep.totals.requests_per_session_typical || 0).toFixed(0) + ' turns — most sessions are '
+ 'one-request sidechains, so the plain average is not what carrying a declaration costs. '
```

Unchanged from `main:dash/ui/tools.js:256`. One caller guarded, sibling caller left broken — the defect class the commit message itself names. Gate it the same way, or drop the causal clause and keep the two figures.

### 3. MEDIUM — the headline tile calls a declarations-only ratio "of every prompt", and this PR's own page then contradicts it

`dash/ui/tools.js:320`:

```js
pct(t.unused_pct) + ' of every prompt, re-read ' + t.requests_per_session.toFixed(1)
```

`unused_pct = unused_tokens / declared_tokens` — the **declaration** total, not the prefix.

Roughly 500 px below, **this PR adds** the correction that makes it a contradiction: *"Share is of the … tokens of DECLARATIONS, not of the whole prefix above — the system prompt is not a declaration and is not in that denominator."*

Before this PR "prompt" loosely meant the declarations, because that was all the page showed. This PR's own thesis is that the system prompt is normally the largest region of the prefix — so the tile now names a denominator the page explicitly disowns, and errs in the direction that inflates the percentage. The `inv-unused` `TILE_INFO` entry does not correct it either.

Inherited from `main:dash/ui/tools.js:147` rather than introduced here — but `ce62dbb` is titled *"say 'declarations' where declarations are what is counted"*, and this is the instance it missed. "of every declaration you carry" fixes it.

### 4. LOW — the shared declaration memo now holds raw text, and a real corpus overruns its byte ceiling

`dash/toolinventory.go:234` adds `len(d.Text)` to `declsSize`, against `declCacheBytes = 32 << 20` whose comment ("~90 KB of declarations … holds hundreds of genuine sets") was written for the old, text-free accounting: an entry used to be a few tens of bytes per declaration and is now the declaration's whole JSON.

Measured over a real multi-day corpus, the distinct declaration sets' text comfortably exceeds that 32 MB ceiling — and the memo is cleared **wholesale** past the cap (`declCache = map[...]{}`), not LRU, where before it never came close.

The absolute cost is small and worth saying plainly: one cold scan per set afterwards, benchmarked at ~1.1 ms, a few hundred of them per multi-day window. Not a blocker — but the comment should describe the accounting it now has rather than the one it had.

Separately, `sysCache` reuses `declCacheBytes` as its **own independent** ceiling (`sysBytes >= declCacheBytes`), so the memo's total memory bound doubled from 32 MB to 64 MB undisclosed.

### 5. LOW — `scanSystem` does not memoize the negative

On `text == ""` it returns nil *after* the cache miss without storing anything, so a client sending `"system": []`, or blocks with no `text` field, pays the gjson walk on every request — on a function whose doc promises "one hash of the raw system JSON per request on a hit". One line: store a nil-text sentinel, or bail before `systemTextOf` on an empty array.

### 6. LOW — the three new UI tests are source-text greps

`dash/uiinventory_test.go:88` asserts `strings.Count(src, "removalCell(") >= 3` and that the substring appears inside `function removalGroup(`. Both pass if the call sits in a branch that never renders — which is precisely round one's failure mode (mechanism present, page unaffected). I confirmed by rendering that it **is** reached today, so the behaviour is right; the guard against the next regression is weaker than the bug it commemorates. The repo has no JS DOM harness, so this is proportionate rather than wrong — flagging it so nobody mistakes the test for coverage of the rendered surface.

### 7. nit — `client_tool`'s `Effect` asserts more than the same struct's `Note` admits

`dash/toolremoval.go` (unchanged here, but promoted by this PR from one call site to well over a hundred): `Effect: "Removes the tool from the prompt entirely."` sits alongside `Note: "…it comes from whatever agent sent the request. Check what declares it before removing it."` Both cannot be certain — a `claude` CLI flag and a `~/.claude/settings.json` deny do nothing to a declaration an unrelated SDK application assembled. On the traffic I checked the mechanism does apply (those names all arrive under `agent = claude-cli`), but the `Effect` should hedge on the declarer the way the `Note` does, now that this is the primary surface for it.

### 8. nit — "Who owns your system prompt" heads a bar with no system prompt in it

Rendered against real history the panel puts that heading over a bar of declarations only, then honestly discloses *"No session in this window recorded a system prompt, so that region is absent rather than zero — see the panel below."* The disclosure is exactly right. The heading above it is not, in the state **every existing deployment will be in on the day this deploys** — no historical row has a `system_prompt` marker.

---

## The consent gate — explicit judgement

**Sound on the write side. The read side is finding 1.**

- The write gate is `proxy.captureContentFor` (`proxy/proxy.go:1541`) — the operator's `--dashboard-content` **AND** the tenant's `CaptureContent`, no third path. Both `finish` call sites pass it (`proxy/proxy.go:626`, `:1214`).
- Carried **per message** (`invMsg.text`), so a tenant toggling consent between two requests of one session is honoured.
- Applied on the **writer**, not the scan — right call, right reason: `declCache` entries are shared by every tenant declaring the same set. `sysCache` is content-keyed too, which is what makes a cross-tenant memo safe: a hit means the bytes were identical.
- `RedactContent(s, maxDeclTextBytes)` scrubs **before** the cap — `internal/redact.Content` applies `cap` as its last step — so a credential near the end cannot survive by being truncated into place.
- Returns `nil`, not an empty blob, when text is not allowed. NULL is a first-class state and the report counts it via `PromptStat.Rows`/`TextRows`, never a fabricated default; on existing history the page reads "not recorded yet" from that count.
- **`schemaVersion` untouched at 6**, `text_gz` through `additiveColumns`, and `addColumns` runs *before* the version check in `migrate()` — so an existing v6 database gets the column and returns early rather than being renamed aside. Verified by opening the recorder against a copy of a real production-sized database: every request row preserved, column present, version still 6. No deploy-time data-loss hazard.
- Size caps: 64 KiB per region, a 512 KiB refusal on the system prompt, 4 system rows per session, plus the existing 256 KiB `maxDeclSet`.

**Leak hunt — nothing beyond finding 1.** `text_gz` is `SELECT`ed in `dash/promptapi.go` only, and there is no `SELECT *` on the table anywhere. `Inventory`/`Decl` are never marshalled to JSON despite `Decl.Text` carrying a json tag. **No logging at any level** in `proxy/dashcapture.go`, `dash/toolinventory.go` or `dash/promptapi.go` — so `CG_LOG_LEVEL=debug`, which is what production runs, adds nothing on this path. `/stats` and `/metrics` never touch the table, and the text never reaches the `Event` that `/stats` is built from. `dash/archive.go` exports events and content rows, not declarations. `httpErr` returns a fixed string, not the driver error. `scanSystem` cannot panic on a missing, oversized, non-string, non-array or text-free `system`, and `sp != nil` guards the writer.

## Both real bugs, at the surface

Playwright over the served page:

```
removal snippets rendered: 187        groups: 115        page errors: none
pills: client tool | MCP | skill | plugin | MCP server | Claude Code | breaks the agent
```

- **Removal command:** `removalGroup` → `removalCell(g.items[0], …)` at group level (`dash/ui/tools.js:614`), above the member list and **not** behind a disclosure. `copyBox` renders a selectable `<pre>` plus a button that reports its own failure rather than pretending, so an unavailable clipboard API is not a dead end.
- **The "built-in" lie:** `KIND_LABEL` no longer contains the word at all (`tool: 'client tool'`), and `kindPill` checks `t.builtin` **first** (`dash/ui/tools.js:114`) with its own word and colour. The four client tools the PR body names now render `client tool` with a real removal snippet.
- **Single source of truth: yes.** `builtin` is the only authority, from `IsBuiltinTool(kind, name) = kind == KindTool && builtinTools[name]` — the map the sibling round correctly settled on erring toward excluding unknown names. `kind` can no longer carry the claim, so the two fields cannot drift apart again. And the new `system_prompt` kind cannot leak into the removal advice: `buildToolReport` `continue`s on it before any `ToolStat` is built (`dash/toolapi.go:511`, with the reason stated), and `SelfRemovals` restricts candidates to `KindMCPTool`/`KindSkill`.

## Numbers and causal claims

Verified against a database copy by SQL and against the live API:

- both session-length multipliers reproduce **exactly** from the code's stated populations — the plain mean over tool-bearing sessions, the request-weighted `Σn²/Σn` over captured ones, and the median — and both name their population in the copy, with the basis panel reconciling the pair. The request-weighted figure is the one the projection multiplies by.
- `totals.unused_usd` reconciles **exactly** to its three parts (tools + skills + listing), to the cent.
- the prompt coverage count now agrees between `PromptStat.Rows` and `PromptView.Rows`; the 13× gap is closed, both grouping by `(session, kind, name, server)`.
- `pct(…, 1)` on the prefix share, so 99.5% no longer prints as 100%.
- "no system prompt recorded → absent rather than zero" is rendered from an explicit count.

No figure sums `saved_gross`, no token ratio is quoted as a dollar ratio, and the one dollar figure on the page is priced per model at the tier each request actually paid, with the page saying so and refusing a flat-tier comparison.

The two false claims are findings 2 and 3.

## Nothing load-bearing dropped

All four present in the **test-merged** tree, so this is what would actually ship: #86's `keepalive = 0` exclusion in `Filter.where()` (`dash/query.go:76`), the three-state unkeyed disclosure with its 20-row floor (`dash/readvalue.go:263`, `minRowsToJudge = 20`), `SelfRemoval` credit (`dash/toolapi.go:863`, rendered `dash/ui/tools.js:1569`), and the Overview savings decomposition (`dash/ui/app.js:1995` `renderNetReconcile`, `net-unkeyed` state intact).

## `dash/ui.go` −116 / `install.sh` −30 — benign, confirmed independently

The merge-base is `918174f`, i.e. before #92 merged, so the negative counts are #92's *absence*, not a revert. The branch touches **neither file** — `git diff --stat origin/main...HEAD --` on both is empty. And a real test-merge of `25ba7d3` into current `origin/main` (`391bff9`) leaves

```
git diff 391bff9 HEAD -- deploy/service/install.sh dash/ui.go   →   EMPTY
```

with `assetVersion` at 5 hits in `dash/ui.go` and the `in_service_path` / `configured_key_envs` PATH fix present in `install.sh`. No cache-busting regression, no reintroduction of the bug that made round one look broken.

## Request path

`proxy/dashcapture.go` (+17) is comment plus one argument; no new I/O, no new allocation, `noteInventory` still one memoized call, `RecordInventory` still non-blocking with a drop counter.

Cost benchmarked here (`-benchtime 200x`): `system_warm 18,377 ns` and `system_cold 229,365 ns` against `warm 193,340 ns` and `cold 1,090,246 ns` — the PR's 22 / 212 / 208 µs / 1.18 ms are the same measurement within run-to-run noise, and the warm addition is a small fraction of a percent of this deployment's average added latency. Honest.

One precision point on "no latency when capture disabled": the system-prompt scan runs whenever the **dashboard** is on, including when *content* capture is off, because the prompt's SIZE is deliberately not treated as content. That is the documented design, but the claim holds for the dashboard being off entirely, not for content capture being off.

`cb5eedb` does what it says — the cold sub-benchmark now clears `sysCache`/`sysBytes` alongside `declCache`, so "cold" no longer measures a cold declaration scan against a warm system scan.

## Conventions

9 commits, every one `Signed-off-by: Osher-Elhadad`, author == committer == the same identity, conventional-commit titles, and **no AI attribution anywhere** — no `Co-Authored-By`, no `Assisted-By`, no "Generated with", no model or tool name as author, in commits, code, docs or the PR body. No secrets: the only credential-shaped strings in the diff are the redaction test's own synthetic fixtures.

**No screenshots committed** — no image file is added by any commit and the image-glob diffstat is empty. The referenced local directories all exist, with 64 captures between them. The capture driver's allow-list is exactly the two authorized accounts, it forbids email and UUID shapes and any other 16-hex id, and it asserts over `document.body.innerText` **after** opening the prompt reveal and the built-ins panel — so revealed text is inside the assertion rather than outside it. The prompt-text captures I inspected render a synthetic system prompt, not real account text.

---

All database work was done on copies, with the `-wal`/`-shm` sidecars removed first; the live service was not touched and its database was never opened read-write. Both throwaway probe tests were deleted and the review worktree is clean.

PR #95, opened 2026-08-22
fix(dash): store each declaration's prompt text once, not once per session

OsherElhadad:
## The measured problem

`tool_declarations` has one row per **(session × declaration)**, and PR #93 put each
declaration's slice of the prompt on that row. So the same `Bash` schema — one of 239 distinct
declarations — was written afresh for every session that carried it. Measured on a copy of a
real database:

| | |
|---|---|
| `tool_declarations` rows | 488,324 |
| rows carrying text | 328,236 (67%) |
| stored text | **254.5 MiB** (avg 813 B/row, largest blob 12.4 KB) |
| **distinct** text blobs among them | **225** |
| duplication factor | **1,459x** |

The file went from 68 MB to 650 MB in a day, and the byte-eviction rule deleted history before
anyone noticed. All eviction is off; this removes the growth instead of the history.

## The change

- New table `declaration_text(hash, text_gz, ts)` — **one row per distinct text**, keyed by the
  sha256 of the scrubbed, capped text.
- `tool_declarations` gains `text_hash` and stops writing `text_gz`.
- Reads join: `COALESCE(t.text_gz, d.text_gz)`. The three SQL fragments live in one place
  (`declTextJoin` / `declTextCol` / `declHasText`) and every read path uses them.
- A resumable in-process backfill moves the existing rows in bounded batches.
- The writer memoizes which hashes it has committed, so the second session declaring `Bash`
  costs a map lookup rather than an INSERT carrying the whole schema for the conflict to
  discard (~87k avoided inserts over the measured corpus). Only hashes from a **committed**
  transaction enter the memo — remembering a rolled-back one would point later rows at bytes
  that are not there, which has its own test with the failure injected by a trigger.

Note the identity is **not** `tool_declarations.digest`, despite that column already being on
the hot path: `digest` is a `maphash` of the whole declaration *set* under a per-process seed,
so it identifies neither one declaration nor the same set across a restart — 4,035 digests for
239 declarations is that seed, not 4,035 configurations.

## Migration safety

- **`schemaVersion` stays 6.** A new table goes in `additiveDDL` and a new column in
  `additiveColumns`, both of which run before the version check on every open. Verified against
  a copy of the real 650 MB file: no rename aside, `schema_version` still `6`, all 24,636
  request rows and all 488,324 declaration rows still there, no column dropped.
- **Nothing is deleted.** `text_gz` is retained, unread by the writer, still read by the
  reader. No row is removed. The backfill clears a row's blob only after the same text is
  committed to `declaration_text` **and read back and compared as text**, in a second
  transaction — so every stopping point leaves the text readable:

  | state | reads via |
  |---|---|
  | before a batch | `text_gz`, through `COALESCE` |
  | after tx 1 | either — both hold it |
  | after tx 2 | `declaration_text`, through the join |

- A blob that cannot be decompressed is **left exactly as it is**. Hashing its (empty)
  decompression would file every such row under one key, pass verification, and delete bytes
  nobody had understood.
- Batches of 500 with a 200 ms pause, on its own goroutine, checking the recorder's `done`
  channel — the write lock is held for milliseconds at a time, never for the migration.

## Equivalence

The acceptance test (`TestDedupeAgainstADeployedCopy`, skipped without `CG_SNAPSHOT_DB`) reads
the file **before** migrating it, using the driver directly and the deployed read shape, then
compares the reveal of all 328,236 rows afterwards. On a copy of the real database:

```
before: 24636 requests, 488324 declaration rows, 328236 carrying text, 254.5 MiB of text, 328236 readable
INFO dash: deduplicated stored prompt text rows=328236 took=1m55.65s
after:  488324 rows preserved, 225 distinct blobs, 0.2 MiB of text (was 254.5 MiB)
database: 621.3 MiB -> 580.9 MiB
```

Every one of the 328,236 rows reveals byte-identical text. The 40 MiB is what the bounded
`incremental_vacuum` the backfill runs itself can return: `incremental_vacuum` only truncates
free pages at the *end* of the file, and the rest stay on the freelist to be **reused** — which
is what stops the growth. A full `VACUUM` takes the same file to **285.7 MiB** (measured), and
is deliberately left to the operator rather than started under live traffic.

## The read gate is unchanged

`/api/prompt` still computes `trusted := a.trusted(r)`, still forces true under multi-tenant
auth, and still strips `Text`/`HasText` otherwise. The join adds no route and no method that
reaches `declaration_text` except through a tenant-scoped declaration row. Both content-gate
tests pass, and both were confirmed to still fail with the gate removed. The write-side consent
pair and `RedactContent`-before-cap are untouched — the hash is taken *after* scrubbing, so it
is the identity of what is stored and nothing else.

## Cold storage

**The metrics database no longer needs any.** Its bulk was this duplication. Session archival
to Box already exists for history an operator wants off local disk.

What is worth parking is the retained artefacts on local disk. `box-setup.sh park FILE` uploads
one file with `rclone copyto`, stats it back, compares the size, and only then removes the local
copy — after asking. It refuses a live database or any file with a `-wal` beside it. It is
manual, unscheduled, and was **not** run against anything: whether a rollback artefact is still
needed is the owner's call.

## Verification

`CGO_ENABLED=1 go test -race -count=1 ./...` green across all 25 packages, no data races,
`go vet ./...` clean, `gofmt -l` empty. For the record on timing: `dash` under `-race` measures
545 s before this change and 562 s after, against `go test`'s 600 s default — that package was
already close to the limit, and it is worth knowing before it starts failing for someone. Every new test was
mutation-checked: the writer storing per-row text again, a truncated content key, the reader
losing its legacy fallback, the backfill never setting the pointer, keying per row instead of
per text, skipping the read-back comparison, dropping the corrupt-blob guard, and the recorder
never starting the backfill — each one fails the test that owns it.


OsherElhadad:
## Review — `3c26574`

**CHANGES-REQUIRED**, three items, each small. None is a data-loss path.

The question I was asked first: **no, I could not make the backfill lose a byte at any
interruption point.** I aimed a process kill at six distinct points of the two-transaction
sequence on a 436,753-row / 256.8 MiB fixture, and every row's revealed text was byte-identical
afterwards — at the kill point and after resuming. The ordering claim holds as written. What I
found is a hole in the *safety net around* that ordering (F1), an un-shrunk-file trade that is
not what the PR says it is (F2), and two of the eight claimed mutations that kill no test —
including the read-back comparison the PR calls "the whole safety argument" (F3).

### Evidence base, and its limit

**I could not verify any production figure in this PR.** No copy of the current database is
readable from where I am, and every snapshot I can reach predates `text_gz` entirely (v6, 19,805
requests, 36,124 declaration rows, no such column). So 488,324 / 328,236 / 254.5 MiB /
621.3 → 580.9 MiB / 285.7 MiB / 1m55s are unverified by me.

Two do corroborate against what I can read: "68 MB before" matches the real pre-`text_gz`
snapshot (61.5 MiB), and "~87k avoided inserts" matches 45 declarations × its 1,935 sessions
(87,075). The internal arithmetic is consistent throughout (813 B/row, 1,459×, 225 × 813 B ≈
0.2 MiB). No dollar or token figure appears anywhere here, so there is no unit-mixing risk to
chase.

Instead I built a production-shaped fixture from the real snapshot — real row identities and
declaration names, sessions cloned to scale, then a text distribution matched to the measured
shape (67% of rows carrying text, 813 B of gzip each):

| | production copy (author) | my fixture |
|---|---|---|
| declaration rows | 488,324 | 436,753 |
| rows carrying text | 328,236 (67%) | 291,171 (67%) |
| stored text | 254.5 MiB | 256.8 MiB |
| distinct texts | 225 | 155 |
| file | 621.3 MiB | 502.4 MiB |

Plus the classes production may not contain: a text at exactly the 64 KiB cap, one at
cap+marker, one over the cap, a one-byte text, a 12.4 KB text, and three blobs that cannot be
decompressed (raw non-gzip, zero-length, truncated gzip).

---

## F1 — A row whose shared blob does not read back keeps its bytes but loses its reveal

`dedupetext.go:216-224` refuses to clear the row's blob when the stored copy does not decompress
to the same text, and explains why: *"the original is still readable, which is the whole reason
the clear is a second step."* The original is still **present**. It is not still **readable**.

Transaction 1 (`dedupetext.go:149`, `:175`) has already set `text_hash`, and the reader prefers
the shared copy — `declTextCol = COALESCE(t.text_gz, d.text_gz)` (`promptapi.go:136`). So after
the mismatch branch fires, the row points at bytes that do not decompress, and every read path
returns empty for a declaration whose text is intact in the column beside it. Demonstrated with
a legacy row plus one `declaration_text` row under the right hash whose bytes are not gzip:

```
ERROR dash: stored declaration text does not match the row it came from; keeping the original
  row_bytes=69 stored_bytes=16 text_bytes=45 got_bytes=0
--- FAIL: the reveal is "", want "{\"name\":\"Bash\",\"description\":\"run a command\"}"
```

Route in production: a corrupted `declaration_text` blob — the same failure class the
corrupt-blob guard exists for, on the shared side instead of the row side.

**Fix, one word:** `COALESCE(d.text_gz, t.text_gz)`. After the backfill only one of the two is
ever non-NULL, so it changes nothing normally; in the abnormal case it keeps the copy the
backfill has just decided it may not clear. The row's own blob is never stale — the writer
stopped writing that column — so preferring it is strictly safer. Verified.

## F2 — The freed bytes are neither on the freelist nor reused, and the shrink is not a property of the change

Three claims — body, commit message, and `dedupetext.go:252-270` / `:294-295`: *"the rest stay on
the freelist"*, *"to be **reused** — which is what stops the growth"*, and the acceptance test's
assertion that the file must shrink (`zz_dedupe_live_test.go:215-217`).

**(1) The pages never reach the freelist.** After freeing 256.8 MiB — ~65,000 pages at 4 KiB —
the freelist held **62 pages (0.2 MiB)**. At 813 gzip bytes average, most blobs sit *inline* in
the table b-tree (inline threshold ≈1,002 bytes on a 4 KiB page). Nulling an inline blob shrinks
the row inside its page; it frees no page. The space becomes intra-page fragmentation, which the
freelist never sees and `incremental_vacuum` can never touch. The log field at `:295` calls it
`free_pages_kept_for_reuse`, which names it as the opposite of what it is.

**(2) It is not reused.** 60,000 new rows in the shape the new writer produces:

| base | free space present | growth for 60,000 rows |
|---|---|---|
| migrated, not vacuumed (553.0 MiB) | 256 MiB freed | **+2,874 pages (11.2 MiB)** |
| same file, fully VACUUMed (196.3 MiB) | none | +2,947 pages (11.5 MiB) |

Identical. New rows append at the end of the rowid b-tree and cannot land in intra-page gaps in
the middle of the table. The 254 MiB is not working capital; it is dead space until `VACUUM`.

What actually stops the growth is the new write path storing one blob per distinct text. That
argument is sound on its own and does not need the freelist story — which is what makes the
wrong version worth deleting rather than defending.

**(3) The shrink assertion is a property of one file's blob-size distribution.** The acceptance
test **fails** on my fixture:

```
before: 436753 declaration rows, 291171 carrying text, 256.8 MiB of text, 291171 readable
INFO dash: deduplicated stored prompt text rows=291168 took=1m31.56s
    migrated 291168 rows of 291171 carrying text          <- FAIL
    3 rows still carry their own blob                     <- FAIL
after:  436753 rows preserved, 155 distinct blobs, 0.3 MiB of text (was 256.8 MiB)
database: 502.4 MiB -> 553.0 MiB
    the database did not shrink: 526779696 -> 579878912 bytes   <- FAIL
```

The file **grew 50.6 MiB**. Mechanism: transaction 1 adds a 64-char hex `text_hash` to every row
*while the 813-byte blob is still on it*, splitting pages, and transaction 2's removal does not
merge them back. On the author's copy that was outweighed by overflow pages released from the
blobs over ~1 KB; where blobs are mostly inline it is not. So **none of `sizeAfter >= sizeBefore`,
`moved != withText` or `legacy != 0` is a safe assertion** — and the latter two fire on
behaviour this PR deliberately implements: the corrupt-blob guard is *supposed* to leave rows
behind, and the headline acceptance test calls that a failure. Production passes today because
it happens to hold zero undecompressable blobs.

Asks: reword (1) and (2) everywhere they appear; rename the log field; let the three assertions
tolerate skipped rows and drop or invert the shrink one; and **say in `docs/hosted.md` that the
migration returns almost none of the existing bytes and that a `VACUUM` at a quiet moment is
what returns them.** `docs/hosted.md:475-478` currently tells an operator the metrics database
"does not need cold storage… its bulk was duplicated prompt text… that is now stored once",
which reads as a promise the file gets small. It does not. The body mentions the 40 MiB; the
docs, where an operator looks, say nothing.

## F3 — The read-back comparison and the transaction ordering are both untested

The body states all eight mutations "each fail the test that owns it". I checked four, plus the
gate, plus two of my own:

| mutation | result |
|---|---|
| recorder never starts the backfill | **killed** — `TestRecorderMigratesLegacyPromptTextOnStart` |
| corrupt-blob guard removed | **killed** — `TestBackfillKeepsABlobItCannotRead` |
| reader loses the legacy `text_gz` fallback | **killed** — 3 tests |
| `/api/prompt` content strip removed | **killed** — both gate tests, naming both markers |
| **read-back comparison skipped** (`dedupetext.go:216`) | **SURVIVES** every dedupe/backfill/declaration/prompt test |
| **tx 1 clears the blob before tx 2 verifies it** (mine) | **SURVIVES** the same set |

No test writes a conflicting row into `declaration_text`, so the mismatch branch is never
exercised. Consequently the property the whole two-transaction design exists for is unguarded: I
inverted the ordering — `UPDATE … SET text_hash = ?, text_gz = NULL` in transaction 1 — and the
suite stayed green. The code as written is correct; a future edit that collapses the two
transactions ships silently.

**One test closes both, and it is the test F1 needs anyway** — a legacy row plus a mismatched
`declaration_text` row, asserting the row's blob survives *and* the reveal is still the text:

| state | result |
|---|---|
| F1 fix applied, no mutation | **passes** |
| + tx 1 clears the blob before verification | **fails** |
| + read-back comparison skipped | **fails** |
| PR code exactly as it stands | **fails** (that is F1) |

## Smaller

- **A tenant purge no longer removes that tenant's prompt text from disk.** Nothing deletes from
  `declaration_text` (stated at `schema.go:481-484`, confirmed by grep); the GC trigger
  (`schema.go:523`) and `PurgeTenant` (`purge.go:130-135`) both work through request rows and now
  remove only the pointer. The bytes stay in the file and in any backup. The schema comment is
  right that this leaks nothing *between accounts* — a shared row is content-identical by
  construction — but that is a different question from whether a purge is complete. Not asking
  for reference counting; asking that `PurgeTenant` and the `declaration_text` row in
  `docs/dashboard.md` say what survives, so whoever answers a deletion request knows.
- **`text_hash` as 64-char hex TEXT costs ~12% of the saving** (`schema.go:488`): 65 B × 488,324
  rows ≈ 31 MB of the ~254 MiB reclaimed. A 32-byte BLOB halves it with no loss of collision
  resistance. Not worth a migration alone.
- **One transient `SQLITE_BUSY` abandons the rest of the migration until the next restart**
  (`dedupetext.go:85-88`, `:245-248`). There is no `SetMaxOpenConns` in the module, so the
  backfill and the capture writer contend on separate connections under `busy_timeout(5000)`; a
  prune or an `incremental_vacuum` holding the lock past 5 s stops the remainder. Resumable and
  logged, nothing at risk — but a rarely-restarted service can carry the 254 MiB for a long
  time, which is the thing this PR exists to fix. A `continue` past a failed batch would finish it.
- **`capture.go:234` says the no-op start is "one indexed query".** `pendingDedupe`
  (`dedupetext.go:113`) is a full rowid scan filtered on an unindexed column; `dedupetext.go:66`
  gets this right, so two comments in one PR disagree. Also, each undecompressable blob costs an
  *extra* full scan on every start forever, since it matches the predicate and the cursor advances
  one row per pass (measured 15.5 s cold with three such rows, 0.28 s warm with none).
- **Rollback is not mentioned.** Rolling back after this deploys leaves new rows with `text_hash`
  set and `text_gz` NULL, and the old reader looks only at `text_gz` — so prompt text silently
  disappears from the dashboard for everything written since the deploy. Nothing is lost and
  rolling forward restores it; worth a line in the body.
- `park`'s live-database refusal is a two-name basename allowlist, so a cleanly-stopped database
  under another name is caught only by the `-wal` check, which is gone once the service exits
  cleanly. Testing the `SQLite format 3` magic covers every such file in one line.

---

## What holds up

**Interruptions.** `os.Exit(9)` — skipping the deferred rollbacks and never closing the database,
as a SIGKILL does — at six points, each fired at batch 3 of 583:

| kill point | state at the kill | reveal at the kill | after resume | third run |
|---|---|---|---|---|
| before the batch | 1,000 pointed / 290,171 on-row | all 291,171 identical | identical | moved=0 |
| mid transaction 1 | 1,000 / 290,171 (rolled back) | all identical | identical | moved=0 |
| end of tx 1, pre-commit | 1,000 / 290,171 | all identical | identical | moved=0 |
| **after tx 1 commits** | **1,500 pointed / 290,171 on-row** | all identical | identical | moved=0 |
| mid transaction 2 | partially cleared | all identical | identical | moved=0 |
| end of tx 2, pre-commit | 1,500 / 290,171 | all identical | identical | moved=0 |

Zero divergence at production byte scale in every case. The `after tx 1` row is the one that
matters: 1,500 rows holding **both** copies, both read orders agreeing — the pointer transaction
1 writes is honest, so nothing is nulled before the deduplicated copy is committed. Resume moved
the remaining 290,168 and a third run moved 0 every time, so re-running against a fully-migrated
and against a partially-migrated database are both no-ops. The work really is a predicate over
the data.

**Equivalence.** I did not take the test's word for it — my own SQL, `sha256(revealed text)` per
row key, under **two** read shapes that differ in which copy they prefer, so a row pointing at
bytes different from its own blob shows up as a disagreement between them:

```
before (old shape, d.text_gz)         : 291,171 revealing rows
after, preferring the shared copy     : 291,171   IDENTICAL
after, preferring the row's own copy  : 291,171   IDENTICAL
```

Byte-identical, all three, including every class named: the 65,536-byte text, the 65,599-byte
cap+marker text, the 102,400-byte over-cap text, the one-byte text, and the 145,582 rows whose
`text_gz` was NULL (correctly absent from all three — counts match exactly, so nothing gained
text either). The cap is irrelevant to the backfill by construction: it moves the stored gzip
verbatim and never re-scrubs or re-caps. On the writer side the cap is covered by the existing
`TestStoredPromptTextIsScrubbedAndCapped`, now reading through the join.

In the author's favour: the reveal map is keyed by
`(tenant, session, digest, kind, name, server)`, *finer* than the table's primary key (`server`
is not in it), so no rows collapse in the map and a lost row cannot hide behind a surviving
duplicate.

**The undecompressable-blob decision does what the claim says.** All three pathological rows came
through the full migration with bytes unchanged, `text_hash` still NULL, still revealing exactly
what they revealed before, and not counted as migrated (`moved` 291,168 of 291,171). The
reasoning at `:164-168` is the sharpest thing in the PR.

**All four read sites use the shared fragments, and none was missed.** Non-test readers of the
text: `promptapi.go:161`, `:183`, `:207-208`, `toolapi.go:413`. The other five queries against
`tool_declarations` (`toolsuggest.go:232`, `toolapi.go:855`, `:895`, `:966`, the count helper)
select only names, kinds, tokens, sessions and timestamps — none needs the join. `archive.go`
exports Events as JSONL and never reads this table. The `LEFT JOIN` cannot fan out:
`declaration_text.hash` is the primary key.

**The content gate is intact, verified by mutation.** With the strip at `promptapi.go:119-123`
disabled, `TestNoRouteServesContentTextFromAnUntrustedAddress` fails naming **both** markers
(system prompt and tool schema) and `TestPromptRouteStillServesWeightsFromAnUntrustedAddress`
fails too. The route-table walk does cover the joined path; the join added no second ungated
route. No new route or handler in the diff.

**Migration safety.** `schemaVersion` is 6 and unchanged from `da73fed`. New table in
`additiveDDL` (applied at `schema.go:567`, before the version check at `:580`); new column in
`additiveColumns` (`:557`), idempotent via `pragma_table_info`. Against my v6 fixture built from
the real snapshot: no `.v6.bak`, version still `6`, all request rows and all 436,753 declaration
rows intact, no column dropped, `text_gz` retained and still read.

**The writer's memo** is correct and bounded: `fresh` folds into `storedText` only after
`tx.Commit()` (`toolinventory.go:801-804`) — the property the trigger-injected test covers, a
good test for the right reason. `invWriters` is keyed by `*Recorder` (`:614`), so the memo cannot
straddle two database files, and `rememberText` handles the nil map.

**Consent** is unchanged in substance: `declText` returns `("", nil)` without it, `storeText`
inserts nothing and binds NULL, and the hash is taken after `RedactContent`, so it identifies
what is stored and nothing else. A non-consenting tenant gets a NULL pointer, not a shared row.

**`park`.** Every refusal works — exercised with a stubbed `rclone`, so nothing touched the real
remote: `cg.db`, `cg-control.db` and any file with a `-wal` are refused (exit 2, file kept);
stdin from `/dev/null` and closed stdin both keep the file; a failed upload aborts before the
delete; a failed stat-back and a size mismatch both keep the file; only a size match removes it.
The local removal cannot precede a verified upload. No mount is introduced.

**Conventions.** DCO present, identity correct, conventional title, no AI attribution anywhere in
the diff, commit message or body (only hits are the `"anthropic"` dialect string in fixtures and
`CLAUDE.md` as a filename). No secrets or credentials. Scope is tight. `go vet ./...` clean,
`gofmt -l .` empty.

### Timing

`dash` under `-race`, same box, identical heavy contention, 30-minute timeout so both finish:
base `da73fed` **736.5 s**, this branch **802.9 s** (+9.0%). And the run that matters: my
`CGO_ENABLED=1 go test -race -count=1 ./...` **failed** — `panic: test timed out after 10m0s`,
`FAIL dash 601.058s`. The other 24 packages passed. So "green across all 25 packages" did not
reproduce here.

Does it need addressing in this PR? **No, and it is not this branch's debt** — the base is
already at 736 s here and 545 s on a clean box, so ~3-9% is not what breaks it. But someone
should fix it in one line: `Makefile:27` (`make test`) and `Makefile:31` (`make cover`, which is
what CI runs — `-race` **plus** coverage instrumentation) set no `-timeout`, so the 600 s default
applies to both. CI is green today only because its runner is fast enough. `-timeout 20m` on both
targets removes a flake that will otherwise land on whoever is unlucky, and the body should say
"green with `-timeout` raised", since that is what was true.

OsherElhadad:
## Re-run against a copy of a real deployment database — verdict unchanged, two findings sharpened, one added

My first review ran on a fixture I built, because no copy of a real dashboard database was readable
from where I sit. One has since been made available to me (scrubbed: row identities, counts, per-row
lengths, byte totals and distinct-blob count preserved; blob *contents* substituted at identical
length). I will call it **B**, the PR's own measurements **A**, and my synthetic fixture **C**. I
report B only as ratios and as agreement or disagreement with A — its absolute figures are
deployment detail and stay out of this thread.

**Verdict is still CHANGES-REQUIRED, for the same reasons: F1, F2 and F3 are all still live. Nothing
in the data-safety argument moved — it got stronger.**

B is ~20% larger than A (the database grew while this PR was being written) and has the same shape:
about 72% of declaration rows carry text, a couple of hundred distinct texts, a duplication factor
within a quarter of A's, and — like A — **zero undecompressable blobs and no text near the 64 KiB
cap** (its largest text is well under it). Those two classes exist only in C, which is the argument
for keeping both fixtures rather than either.

### The interruption story, now measured on real row identities

`os.Exit(9)` inside `dedupeOneBatch` — skipping the deferred rollbacks and never closing the
database, as a SIGKILL does — at six points inside one batch, 500-row batches. State read with the
WAL honoured:

| kill point | rows cleared | rows pointed | meaning |
|---|---|---|---|
| before the batch | 3,000 | 3,000 | six clean batches |
| mid transaction 1 | 3,000 | 3,000 | tx 1 rolled back whole |
| end of tx 1, pre-commit | 3,000 | 3,000 | tx 1 rolled back whole |
| **after tx 1 commits** | 3,000 | **3,500** | **500 rows hold BOTH copies** |
| mid transaction 2 | 3,000 | 3,500 | tx 2 rolled back; nothing cleared |
| end of tx 2, pre-commit | 3,000 | 3,500 | tx 2 rolled back; nothing cleared |

**`cleared` never exceeds `pointed`, and never advances except through a committed transaction 2
that verified every row it touched.** That is the invariant the PR claims, visible in the data
rather than argued from the code.

| | aimed kill points | checks (at the kill + after resume) | divergences |
|---|---|---|---|
| **B** (real row identities) | 4 — mid tx1, end tx1, after tx1, mid tx2 | 8 | **0** |
| **C** (synthetic) | 6 — adds before-the-batch and end tx2 | 12 | **0** |

Every check compares every text-carrying row's revealed text against the pre-migration capture,
under both read orders. Resume completed the remainder every time and a third run moved 0, so
re-running against a fully-migrated and against a partially-migrated database are both no-ops.

### Equivalence on real data

My own SQL, `sha256(revealed text)` per row key, under two read shapes that differ in which copy
they prefer — so a row pointing at bytes different from its own blob shows up as a disagreement
between them. On B: **every text-carrying row byte-identical, both orders, none lost, none gained.**
Rows preserved exactly; **zero** left carrying their own blob; the stored text collapsed by
**1,307x** (A reported an equivalent collapse); `integrity_check ok` before and after; no `.v6.bak`;
`schema_version` still 6; no column dropped.

### F2, now quantified on real blob sizes — the claim is wrong, and A's number is right

| | share of the bytes the backfill freed |
|---|---|
| reached the freelist | **15.5%** |
| intra-page fragmentation, invisible to the freelist | **84.5%** |

SQLite's inline payload threshold is ~1,002 bytes on a 4 KiB page and A's own average blob is 813
bytes, so most blobs sit *inside* the table page: nulling one shrinks the row and frees no page. Only
blobs above the threshold have overflow pages, and only those come back. The bounded
`incremental_vacuum` returned all of the 15.5% and left the freelist empty; a full `VACUUM` then
returned roughly half the file again.

That vindicates the author's measurement: **A returned 15.9% of what it freed, B 15.5%.** The number
was honest. The *explanation* is wrong, in three ways:

1. *"the rest stay on the freelist"* — 84.5% never reaches it, and after the reclaim loop the
   freelist is **empty**, so `dedupetext.go:294-295` logs `free_pages_kept_for_reuse=0` while the
   bulk of the freed space sits in fragmentation the line does not mention. The field name asserts a
   story its own value contradicts.
2. *"to be reused"* — it is not. 60,000 new rows in the shape the new writer produces grew the
   migrated-but-not-vacuumed file by **+2,874 pages** and a fully-VACUUMed copy by **+2,947 pages**:
   identical, because new rows append at the end of the rowid b-tree and cannot land in intra-page
   gaps mid-table. Confirmed on both B and C.
3. The shrink assertion (`zz_dedupe_live_test.go:215-217`) passes on A and B and **fails on C**,
   along with `moved != withText` and `legacy != 0` — the latter two firing on the corrupt-blob
   guard this PR deliberately implements. All three assert facts about the data, not properties of
   the code. One corrupt blob in production and the headline acceptance test goes red for the right
   behaviour.

What actually stops the growth is the new write path, which needs none of the above — so the
freelist story is worth deleting rather than defending, and `docs/hosted.md` should say plainly that
the migration returns about a sixth of the existing bytes and a `VACUUM` at a quiet moment returns
the rest.

### F4 (new) — this removes roughly half the growth, and the PR reads as if it removes all of it

`dbstat`, as shares of a real pre-migration file:

| object | share of the file |
|---|---|
| `tool_declarations` | 62% |
| `sqlite_autoindex_tool_declarations_1` (the PK) | 11% |
| `idx_tooldecl_tenant` | 3% |
| `idx_tooldecl_name` | 2% |
| **declarations + their three indexes** | **80%** |

Text is about 69% of that table. After dedup collapses it, the row bodies and index entries that
remain come to **504 bytes per declaration row**, plus 65 more for the new `text_hash`. And rows keep
accruing at the post-#93 rate: at that rate **the residual growth is within a few percent of the
text growth this PR removes** — so the change roughly halves the growth rather than ending it.

"This removes the growth instead of the history", and `docs/hosted.md`'s "there is nothing to build
for size", will be read as *solved* — and the next person to look at the file will not know that 80%
of it is a table this PR already touched.

The grain is what is left: **one row per (session × declaration), for roughly 240 distinct
declarations service-wide** — measured at ~299 rows per session — each row carrying a tenant id, a
session id, a digest, a name and now a 64-byte hash, indexed three times. That is the same
duplication this PR just fixed one level down: the text was stored once per session that carried it,
and the *identity* still is. Not asking this PR to fix it; asking it to name the residual with a
number, so nobody reads "growth solved" into it.

### Corrections to my own first pass

- **F5 is measured now, not estimated**: `text_hash` as 64-char hex TEXT costs **8% of what the
  dedup saves**, not the ~12% I projected. A 32-byte BLOB halves it.
- My first post-kill *state* readings came from a tool that opened the file with `immutable=1`,
  which reads only what has been checkpointed and silently ignores committed WAL frames. They were
  stale, and produced one false "divergence" that was my harness, not this branch. Every state
  figure above is read with the WAL honoured, and the corrected numbers match the design exactly.
  Worth knowing for anyone re-running this.

F1, F3, F6 and F7 stand as posted, unchanged.

OsherElhadad:
## Two additions: the grain recommendation, and evidence for the corrupt-blob guard

### Recommendation on the residual growth (F4): follow-up, and the ratio makes it urgent

Measured on the real file: the declaration table holds roughly **130 copies of each distinct
declaration set** — a couple of hundred (session × digest) pairs' worth of identity for every one
distinct set, ~87 declarations per set. A set-level normalisation would cut declaration rows by
about **50x**, taking the ~504-bytes-per-row residual from 80% of the file to a rounding error.

**It should not go in this PR.** Three reasons:

1. **Risk.** This migration is now verified byte-identical at six interruption points on real row
   identities. A second, larger migration in the same PR — every declaration row, all four read
   sites again, the GC trigger again — doubles the surface. One verified migration at a time.
2. **It is a bigger change, not a smaller one.** The text collapsed because it was *byte-identical*:
   one content hash, one row, and the reader cannot tell. The (session × declaration) rows are not
   duplicates — each asserts a relation — so collapsing them needs a **stable set identity** first,
   and `tool_declarations.digest` cannot be it: it is a `maphash` under a per-process seed
   (`toolinventory.go:150`), which is exactly why there are ~130 identity-pairs per distinct set —
   every proxy restart re-digests the same set. The follow-up is a content hash of the set, a
   `declaration_set` table, a `session_set` table, a backfill, and rework of the four read sites and
   the trigger. That is this PR's size again.
3. **Deferring is cheap; holding is not.** This PR stops the text growth today, and the residual is
   the same order as what it removes. Shipping now and following up immediately beats holding a
   working fix behind a redesign.

So: merge this once F1–F3 are addressed, and **file the follow-up with the ratio before it merges**
— 50x is not a someday number, and the PR body naming the residual is what stops someone
rediscovering it in a month.

One fact that reframes the urgency without changing the direction: the row grain is not new. Every
kind (`skill`, `tool`, `mcp_tool`, `skill_listing`) existed at the same grain three days before #93,
at ~5% of today's row count with 40% of the requests. #93 added `system_prompt` rows (19 of them) and
the text — nothing else. What grew is declarations *per session*: `mcp_tool` rows grew ~270x, which
is client configuration, not traffic. And the text load is concentrated in a handful of accounts. So
the grain question is really *what happens when content capture is enabled more widely*, which is an
argument for filing it now while it is cheap.

### A 127,881-row exercise of the corrupt-blob guard

Provenance stated plainly, because it was an accident: the first scrubbed snapshot I was given
preserved blob *lengths* but not the gzip *format*, so 31% of its text-carrying rows were
undecompressable. That has since been corrected — and it produced the best test in this review of the
guard at `dedupetext.go:163-168`, which the author had covered with three synthetic rows.

On that file the backfill moved 284,521 rows and left **127,881 exactly as they were** — `text_hash`
still NULL on every one, not one counted as migrated — while all 412,402 text-carrying rows still
revealed byte-identical text under both read orders, the 127,881 still revealing empty as they did
before, with their bytes intact. All rows preserved, `integrity_check ok`. **The guard holds at
~42,000x the scale it was tested at.**

It also produced F2(3) by a second, independent route: `moved != withText` and `legacy != 0`
(`zz_dedupe_live_test.go:154`, `:191`) both fired. **The acceptance test cannot distinguish "31% of
this database's blobs are corrupt" from "the migration failed"** — and reports the second. That is a
real weakness in the test regardless of how the file came to be that way, and it is the reason those
two assertions should tolerate skipped rows.

### The fair version of the intra-page point

Free space inside a page is not literally dead — it is available to that page's own b-tree inserts.
What I measured is that it is not available to the inserts *this* table receives, because new
declaration rows append at the end of the rowid b-tree; hence identical page growth with and without
it. So the accurate statement is: the freed space is not on the freelist, is not generally reusable,
and is not the mechanism that stops the growth. The new write path is that mechanism, and it needs no
help from the freelist.

Worth naming the shape: two independent measurements agree on the *returned* fraction (15.9% and
15.5%) while the *explanation* of the remainder is wrong. A plausible rationale attached to a correct
number is harder to catch than a wrong number, and it survives review precisely because the number
checks out.

### Everything else stands

F1, F2, F3, F5, F6, F7 unchanged. The byte and size figures in my previous comment were already
measured on the corrected snapshot, so nothing there needs restating.

PR #96, opened 2026-08-22
feat(dash): the keep-alive control tab, and two config-save bugs

OsherElhadad:
## Two config-save bugs, and a keep-alive tab that leads with the downside

Both bugs are data loss in a settings page. Both are fixed at the root, and each has a test
that goes red when the fix is reverted.

### Bug 1 — the keep-alive opt-in did not persist

`config/form.go`'s `child()` returned a **detached** map when the key was absent:

```go
func child(m map[string]any, key string) map[string]any {
	if sub, ok := m[key].(map[string]any); ok && sub != nil {
		return sub
	}
	return map[string]any{}      // never inserted into m
}
```

Its own doc comment said *"creating it when absent"*. Every caller that works compensates by
assigning the result back (`m["components"] = comps`, `m[k] = sub` in `setPath`). The `cache:`
block was the one caller that did not — so the whole block was written into a throwaway map and
discarded, but **only on a document with no `cache:` block already**, which is exactly a tenant
switching the keep-alive on for the first time.

Fixed in the helper, not at the call site: the next block someone adds to `ApplyForm` cannot
repeat it. The two callers that prune an empty block still prune it, so an empty map created
here never reaches the document.

| | before | after |
|---|---|---|
| store a document with no `cache:` block, post `cache.keepalive = true` | output contains **no `cache:` key at all** | `keepalive: true`, and every `CacheForm` field round-trips |

`TestSavingKeepAliveOnADocumentWithNoCacheBlockKeepsIt` is table-driven over every `CacheForm`
field via reflection, and asserts on the **re-parsed form** rather than on a substring — a test
for `"keepalive: true"` in the text would pass on a document that also broke something else.
`TestChildAttachesTheBlockItCreates` is three lines on the helper and is the assertion that
generalises. Reverting the fix fails all seven subtests plus the helper test.

One consequence, and it is intended: a re-save now materialises `cache: {keepalive: false,
head_ttl_1h: false}` on a document that had no block. An existing test asserted "a re-save moves
nothing"; it now allows exactly those two keys, with the reason. No tuning field is written, so
today's defaults are still not frozen into anybody's document.

### Bug 2 — a save silently reverted an operator's pipeline

Given the right payload the server does not drop `linecap`: `ParseForm` → `ApplyForm` preserves
it, and `config/form.go` needs no per-name knowledge, so `grep linecap config/form.go` being
empty is not evidence of the defect. The wrong pipeline was **posted** — `saveSettings` rebuilds
the list from the checkbox grid and the server writes it wholesale, so either a stale render or
a cached bundle whose `/api/options` predates a component produces exactly the observed diff
(`linecap` dropped, `toon` appended last).

Two server-enforced rules, and the client is never trusted to state absence:

1. **`pipeline_known`** — the names the client rendered a control for. Anything in the stored
   pipeline that the client neither sent nor claimed is preserved **at its original index**.
   Empty means "this client declared nothing", and then nothing may be removed: an old bundle or
   a hand-rolled API client can add and reorder but never remove. Removal became an act a client
   has to claim.
2. **`pipeline_base`** — the pipeline the page rendered from. A mismatch against the stored
   document is a **409** with "reload before saving". This is what stops a stale page *adding*
   `toon` back; rule 1 only defends against dropping. `nil` is accepted (older bundle), and rule
   1 is then the only defence, which is the safe direction.

The membership test is on what the client **sent**, and the insertion is into the **resolved**
pipeline. Those are deliberately different: testing membership on the resolved list would undo
`applyExtractLLMCoupling`'s own removal of `extract_llm` — a server decision, not a client
omission — and put back a component its constructor refuses to build.

| | before | after |
|---|---|---|
| stored `[format, linecap, extract]`, client posts `[format, extract]` without claiming `linecap` | `linecap` gone | `[format, linecap, extract]`, at its own index |
| same, with `linecap` in `pipeline_known` | gone | gone — a declared removal still removes |
| client declares nothing, posts `[format]` | everything else gone | nothing removed |
| page rendered from a pipeline that has since changed | applied | **409**, stored document byte-identical |

Two in-repo callers had to state their claim, which is the honest expression of what they do:
`POST /api/toolfilter` claims the one component it models (without it, an emptied removal list
could no longer take `toolfilter` out of the pipeline), and the switch-off test declares the
document it read.

**Comment preservation is out of scope, with the reason.** It means abandoning `map[string]any`
for `yaml.Node` throughout `ApplyForm`, `setPath`, `delPath`, `child` and `readBlocks` — a
rewrite of the exact code whose two previous rewrites each shipped a data-loss bug (the regex
pipeline mangling, and Bug 1 above). Mitigated instead, since the prior document is already
verbatim and durable in `tenant_config_audit.before`:

- a **"Copy the document as it was"** button on every `config_yaml` audit row, so comment loss
  stops being unrecoverable. No new storage — the row already holds the whole text.
- honest copy beside the Save button: *"Saving rewrites your configuration document from these
  fields. Comments and key order are not preserved — the previous document is kept verbatim in
  the audit log below, with a button to copy it back."*

---

## The tab

`data-view="keepalive"`, after Usage, visible to every principal for their own data, and it
respects the time-range filter (so it is deliberately **not** in `UNFILTERED_VIEWS`). Six
panels, in the order a decision is actually made:

1. **Verdict** — six stat tiles: your position (Winner / Losing / No pings yet, in words beside
   the colour), the net, pings and what they cost, re-creations avoided, what is still on the
   table, and the ping rows' own footprint on disk.
2. **Am I in the losing majority?** — a status bar (winners `--good` / losers `--bad`, both
   labelled in words), the worst session named in dollars, and a sortable table of every session
   with its own net. It renders **unconditionally**, including on a positive net. When the
   account's own net is negative it becomes a `banner bad` with a **"Turn keep-alive off"**
   button.
3. **Your TTL expiries** — expiries/day and $/day as **two separate charts** (a count and a
   dollar figure share no scale); idle-gap bands in fixed edge order with everything beyond the
   current `K·X + TTL` in the de-emphasis gray; prefix-size bands with the p50 called out; 24
   hour bins **in the viewer's own timezone with the offset stated**; and the gap between
   expiries **per account**, never blended.
4. **Calculator** — the K = 1..4 ladder at a chosen X, each rung showing its coverage
   arithmetic, the account's own convertible expiries, the share of reachable dollars, the ping
   count and the net, plus one bar per rung so reach and cost are read together.
5. **Keep one session warm** — arm a session with a mandatory expiry; the live armed list.
6. **What should I set?** — a range with its n, or a refusal.

The verdict is first and the downside second, **above** the calculator: 85 of the 119 sessions
this mechanism touched lost money funding a rebate on 34, and a page that led with the
calculator would be selling. `TestTheLosingMajorityIsOnThePage` asserts that ordering and that
the split is not rendered inside a negative-net branch.

Forms reuse what exists — stat tiles, `barRows`, `lineChart`, the status bar, tables. No new
chart library, no new hue, no pie, no treemap, no gauge, no second y-axis;
`TestTheKeepAliveTabAddsNoNewChartVocabulary` enforces that against the code (not the prose,
which names the forms it refuses). `barRows` gained a hover title and an optional click-through,
which every bar on this dashboard should have had.

### The calculator's arithmetic

Computed on the server, in Go, in one place. The browser posts inputs and renders answers: it
has twice duplicated a table the server owns and drifted from it, and money is the worst case of
that class.

```
pings_per_span(gap)  = min(K, floor((gap − X)/X) + 1)      for gap > X, else 0
ping_cost            = P × price.CacheRead + price.Output   # one output token; max_tokens=1
coverage_seconds     = K × X + TTL                          # TTL = 300 s
avoided_miss         = (price.CacheWrite − price.CacheRead) × P
expected_saving      = avoided_miss × (addressable misses whose gap ∈ (X, coverage])
expected_ping_cost   = Σ over the account's OWN spans of pings_per_span(gap) × ping_cost
net                  = expected_saving − expected_ping_cost
```

`coverage = K·X + TTL` is load-bearing — the last ping is itself a read, and a read refreshes
the entry for the provider's full lifetime. Writing `K·X` is the single error that made an
earlier analysis of this mechanism wrong by a factor of 4.4, so
`TestCoverageIncludesTheFinalPingsTTL` pins seven golden values ((280,2)→860, (280,1)→580,
(240,2)→780, …) and a `K·X` implementation fails every one.

Ping cost is priced from **the row's own model**, per the operator's list. An unpriced model
renders "not priced" and produces **no dollar figure at all** — never a blended average, which
is the defect class this project has hit five times. The prefix is a **real** one, prefilled in
order: the selected session's last billed prefix → the account's own median at a lapse → nothing
(and the panel asks). Never a service-wide average: per-ping cost is bimodal (p50 $0.0004, p99
$0.2275).

Two caveats the panel states in words: it is a replay of the account's own past gaps and not a
forecast, and it counts only **addressable** expiries. A `ttl_expiry` row with `cache_write = 0`
is a phantom — 385 of 742 in production, every one an HTTP 400 — and they are counted apart so
this panel and the Usage tab's `ttl_expiry` count can be reconciled.

### How the saving joins the totals without contaminating the aggregates

One line: `o.TotalSavedUSD = o.NetSavedUSD + o.CachesplitSavedUSD + o.KeepAliveNetUSD`.

**The net, never the gross.** `o.CostUSD` excludes ping rows, so the ping spend appears nowhere
else in the walk — adding the saving alone would present a saving without the spend that bought
it. PR #86's predicate (`r.keepalive = 0` in `Filter.where()`) does not change; every new
statistic where a ping is the subject runs its **own** query with `withKeepAlive(f)` and joins in
Go, the same two-query pattern the overview already uses. There is no `CASE` over `keepalive`
inside an agent-traffic aggregate anywhere in the new code.

`TestPingRowsStayOutOfAgentAggregates` runs one fixture twice, with and without the ping rows,
and compares the whole `Overview` **key by key** through JSON — so a field added later is covered
without anyone remembering — plus `/api/breakdown` in every dimension and `/api/series`. Only
the four keep-alive keys and `total_saved_usd` may move, and `total_saved` must move by exactly
the change in the net. `TestTotalSavedTakesTheKeepAliveNetNotItsGross` makes a window where the
pings cost more than they saved and requires the headline to go **down**;
`TestTheWaterfallReconcilesWithTotalSaved` requires the signed steps to sum to the cent.

Confirmed on the capture database: `Requests = 1,770` and `8 sessions` are identical with and
without 81 ping rows present, while `Total dollars avoided` moves from $1.47 to $6.61 — exactly
the $5.15 net.

Tiles that changed: the Overview headline's sub-label (`compaction + prefix cache + keep-alive`),
`TILE_INFO['total-saved-usd']`'s catch (three addends; the net, not the gross; and the ceiling),
the `keepalive-net` tile (both halves in its sub-line, "a ceiling", and a link to the tab),
`total_saved`'s waterfall description, and one new **Keep-alive net** column on Sessions —
blank, not `$0`, where it never ran.

### Per-session overrides

In memory on the keeper, under the existing lock, with **no new table**. The dash DB may not grow
unboundedly and a schema bump renames history aside; but the deciding argument is semantic — an
override is an **authorization to spend**, and one that silently survives a restart is worse than
one that does not. The UI says so, and the durable half is an audit row per arm and per disarm
carrying the resolved parameters and the expiry.

One hook in `record`, immediately after `pol := tn.Cache`. Everything downstream — `pol.on()`,
`pingable()`, the hard deadline, the per-ping guard, the kill switch, the no-audit-sink refusal —
is untouched.

| rule | value |
|---|---|
| `until` mandatory | max **12 h** (session lifetime p99 is 12.9 h), default 1 h |
| `Idle` | 60–290 s (past 290 the first ping lands after the TTL has lapsed and pays a **write**) |
| `MaxPings` | 1–11, **and `(K+1)·X ≤ 60 min`** — the credential-hold ceiling, shown to the caller |
| `MaxUSDPerPing` | **not settable**; the body's value is ignored and the answer reports the one in force |
| per account / total | 64 / 512 |
| arm rate | 30/hour/account, via a new `Limiter.AllowAction` |

Every bound is a 4xx **with the reason**, never a silent clamp: a clamp on a spend authorization
tells someone they got what they asked for when they did not. An override may switch the
mechanism **on** for one session while the account default is off — that is the point, and the
arming request is the consent act — but it cannot reach around the kill switch or the audit-sink
refusal. `POST` answers with the computed credential hold in minutes and the worst-case spend,
priced from that session's own last prefix. Arming without showing what it can cost is the thing
this feature is trying not to be.

An override holds **no material**: arming creates no `kaEntry`, so one left on a dead session
costs exactly zero (`TestOverrideOnADeadSessionCostsNothing`). Disarming releases what is held
**now** rather than at the deadline, zeroized — `TestDisarmRetiresTheHeldMaterialNow` extends the
no-output-surface bar to that path at debug level. `keeper.forget` (which every Settings save
calls) drops a tenant's overrides too, so unticking the account box kills armed sessions.

**One gap closed.** The account-level off switch was drawn disabled for a non-manager
(`ka.disabled = inherited || !mgr`) because it lives in the manager-owned configuration document
— so an account owner whose own key was being spent could not switch it off from the page.
`DELETE /api/me/keepalive` is open to **any** principal on their own account and is deliberately
one-way: it clears the stored consent, drops every override, and calls `forget` so retention
stops now. Turning it back **on** keeps the manager gate.

### Routes

Five reads in `dash/api.go`'s `routes()` table, so `TestEveryMountedRouteDeclaresItsScope` and
`TestNoRouteServesContentTextFromAnUntrustedAddress` both walk them. All `scopeTenant`, all
returning numbers and enum labels only — no prompt text, no transcript, no tool schema — so the
content gate passes on them by construction and no CIDR gate is needed. Four writes in
`ctlRoutes()`, sharing `PUT /api/me`'s authentication and CSRF path exactly; the arm route
declares `ctlManager`, which is the honest statement of the gate it enforces.

### Migration and storage

**`schemaVersion` stays 6. No new table, no new column, no `additiveDDL`, no `additiveColumns`.**
Every statistic is a query over `requests` plus the three columns PR #86 already added. Nothing
deletes: no eviction, retention or purge path is touched.

Verified by opening a copy of the production snapshot (which predates those columns): the version
stayed **6**, the three keep-alive columns were added additively, and all **1,770** rows survived.
Historical rows read "not recorded yet" through an explicit coverage count — never a fabricated
default — and the UI renders three distinct states: *the keep-alive has not run on this account*,
*recorded on N of M requests in this window*, or the plain figures. All three were exercised on
the capture database.

Storage growth is ping rows and nothing else. Measured with `dbstat`: 294 B of row plus ~103 B of
index = **≤ 400 B per ping**.

| policy | pings/day | bytes/day |
|---|--:|--:|
| shipped gate | 204 | ~82 KB |
| blanket (ungated) | 1,995 | ~800 KB |

The tab shows the account's own rate and projected growth, because per-session overrides are the
one way a user could raise it. Nothing here stores text.

### Honesty, and where it surfaces

Every headline is a **range with its n**. The recommendation is **refused** unless all three
hold: ≥ 20 addressable expiries, ≥ 200 requests, and a 90% bootstrap interval over **sessions**
(2,000 draws, deterministic seed; sessions, not requests, because expiries within a session are
not independent draws) that excludes zero **on the positive side**. On the production corpus that
admits 4 of 13 accounts and refuses 9. The wire carries **no point estimate field at all** — a
field that exists gets rendered — and `TestRecommendationPayloadCarriesNoPointEstimate` asserts
its absence in the JSON. `K = 1` is never recommended (−$71 service-wide: one ping reaches ~4.7
min, inside the free TTL). K = 2 and K = 3 are reported as a measured tie unless K = 3's interval
clears K = 2's entirely.

Every tile carrying `keepaliveSavedUSD` states it is a **ceiling**: the provider's cache is
content-keyed, so another session sending a byte-identical prefix would have refreshed the same
entry for free. The counts beside it (pings sent, pings that read nothing, pings that wrote) are
what an operator checks it against.

The panel above the session table says, in the measured numbers, that per-session selection is
not a strategy: a list fitted on one half of a week earned **$5.27** in the next half where the
plain account-wide rule earned **$53.32**.

### Found while building this

`Overview`'s replay-ceiling subquery had **no predicate at all** on its inner count, so a
keep-alive ping counted as a later turn that could replay a reduction. It cannot: a ping is a
verbatim resend carrying no new transcript, and PR #86's invariant is that a ping is not a turn.
On a three-ping fixture it inflated the ceiling from 30 to 50 tokens and dragged
`replay_realized_pct` down with it — which reads as compaction realising less of its value than it
does. On the capture database it over-counted by 517k tokens (3.4%). Caught by the key-by-key
comparison, not by an aggregate anyone thought to check.

**The obvious fix was a 20% regression on `Overview`, and measuring caught that too.** Adding
`AND p.keepalive = 0` inside the correlated count is right logically and wrong operationally:
`keepalive` is not in `idx_requests_session`, so it turns an index-only count into a row fetch per
candidate. Measured over three alternating runs against `main`:

| | main | `AND p.keepalive = 0` | shipped |
|---|--:|--:|--:|
| `Overview`, 10k-row fixture | ~231 ms | ~277 ms | ~235 ms |
| same under `-race` | 10.6 s | 15.3 s | 10.4 s |
| `Overview`, capture DB (81 pings) | 147–274 ms | — | 163–230 ms |

Under `-race` that 44% was enough to blow `TestOverviewStaysFastOnALargeWindow`'s budget on a
loaded box. The shipped form keeps `main`'s index-only count and subtracts a correction driven
from the **ping** side — one indexed sum per ping over its own session's earlier rows, which is
`O(pings × session)` rather than a second full pass — behind an `EXISTS` gate, so it costs
literally nothing where the mechanism has never run.
`TestTheReplayCeilingCountsAgentTurnsNotPings` pins all of it: the uncorrected value, that a ping
does not move it, and that a ping *before* a reduction is not over-subtracted either.

(The inner count is still not window- or tenant-scoped, and both statements agree about that;
session ids are `tenant:uuid` so a cross-tenant collision is not reachable, and the ceiling is
deliberately about the session's whole life. Noted, not changed.)

### Four defects that only gave up when the page was opened

Reading the code was not enough; each rendered plausibly and said something false.

- **The K-ladder table was absent.** `barRows` clears the host it is handed, and the share chart
  was given the panel's own host — so the table it replaced simply was not there, under a row of
  bars carrying all the same numbers.
- **The recommendation offered a policy whose own 90% interval was −$38.63 to −$4.54.** "Excludes
  zero" is a two-sided reading; entirely below zero is not inconclusive, it is a finding, and it
  is now its own refusal with its own wording and its own `banner bad`.
- **Band edges rendered as `5m–9.666666666666666m`**, and a band with no expiries drew a 2px stub
  that reads as "a little" rather than "none".
- **The calculator said "not priced" on a deployment whose price list was fine**, because nothing
  had named a model. It now prefills the account's own **modal** model on its own expiries — an
  answer about that account, not a blend.

Plus: a "top 7 of these hold $81.40 of $81.40" tautology; a last-prefix of `0` rendering as a size
rather than as an absence; an arm confirmation that reported the credential hold only *after* the
authorization; and — on a single-tenant deployment, which mounts no control plane — a "Keep warm"
button on every row whose only possible outcome was a 404. The panel and the whole action column
now go once that is known.

## Verification

`go vet ./...` clean, `gofmt -l .` empty, and `CGO_ENABLED=1 go test -race -count=1 ./...` green
across all **25** packages — **with `-timeout 25m`**. That timeout is not this branch's doing:
`origin/main`'s own `dash` package takes **671 s** under `-race` on this box, already past Go's
default 10-minute per-package limit, so a default-timeout run is load-dependent on `main` too.
This branch's `dash` is **680 s** — a 9 s delta. Worth raising separately.

46 new tests (16 dash query/attribution, 5 dash UI-asset, 11 keeper override, 7 control-plane, 5
config). Every fix was mutated and its test confirmed red — the two config fixes, the one-sided
recommendation rule, and all five UI honesty checks (gating the split on a negative net, dropping
the service-wide figures from the markup, removing a `TILE_INFO` entry, rendering the
recommendation as a point, putting the tab in `UNFILTERED_VIEWS`, and collapsing the two charts
into one).

The dashboard was served against a copy of a copy of the production snapshot and the tab was
**looked at**, in light and dark, on two databases: one where the keep-alive has never run (the
absence state) and one with a simulated run over those accounts' real gaps (the populated
state). Screenshots are at the local path in the review notes and are **not committed** — this
repo is public and they carry real figures. The capture database was built from the two
authorized accounts only, every content table was emptied, session UUIDs were remapped, and a
programmatic scan for emails, session UUIDs, foreign tenant ids and host paths reports **0
findings** on both databases.

### DIFF
--- a/config/form.go
+++ b/config/form.go
@@ -57,7 +57,31 @@ type Form struct {
 	// not enablement, and reading "no block" as "off" silently removed components from
 	// bare-preset accounts on save.
 	Pipeline []string `json:"pipeline"`
-	Mode     string   `json:"mode"`
+	// PipelineKnown lists the component names the CLIENT rendered a control for. Any name in
+	// the STORED pipeline that is not in this list is preserved at its original index,
+	// whatever Pipeline says — so removing a component is an act a client has to CLAIM, and a
+	// client that cannot see one cannot take it out.
+	//
+	// It exists because a save silently reverted an operator's pipeline: `linecap` was
+	// dropped and `toon` re-added. The server never knew either name — the pipeline is a
+	// wholesale write of whatever the browser rebuilt out of its checkbox grid — so a stale
+	// render or a cached bundle whose /api/options predated a component produces exactly that
+	// diff, and the server had no way to tell it from a deliberate removal. Now it does.
+	//
+	// Empty or absent means "this client declared nothing", and then NOTHING may be removed.
+	// That is the safe direction for an old bundle and for a hand-rolled API client: both can
+	// still add and reorder.
+	PipelineKnown []string `json:"pipeline_known,omitempty"`
+	// PipelineBase is the pipeline the page RENDERED from. A mismatch against the stored
+	// document means the document moved under the page, and the save is refused with a 409 —
+	// which is what stops a stale page ADDING a component back. Preserving the unmodelled
+	// only defends against dropping.
+	//
+	// nil is accepted (an older bundle does not send it) and then PipelineKnown is the only
+	// defence, which is again the safe direction. Checked at the HTTP layer, where the stored
+	// document is: see proxy.Handler.ctlUpdateMe.
+	PipelineBase []string `json:"pipeline_base,omitempty"`
+	Mode         string   `json:"mode"`
 	// Cache is the `cache:` block — host-level prompt-cache policy (the idle keep-alive and
 	// the mixed-TTL head), which is not a component and so has no descriptor to be drawn
 	// from. nil means the form does not state it and ApplyForm leaves the document's own
@@ -67,6 +91,12 @@ type Form struct {
 	// has nothing to say about the cache" are different instructions, and with a value type
 	// every save from a page that had not drawn the control would switch a tenant's
 	// keep-alive off.
+	//
+	// Reachable only from a direct Go or API caller, in practice. ParseForm's strict path
+	// always fills it in (cacheForm is unconditional), and its loose path leaves it nil but
+	// also sets ParseError — which the save route answers with a 409 before ApplyForm is ever
+	// called. So the nil branch in ApplyForm is a contract for library callers, not a state
+	// the settings page can produce.
 	Cache *CacheForm `json:"cache,omitempty"`
 	// Components holds, per component name, the DOTTED key paths that component's block
 	// actually states — `{"extract_llm": {"cold_cache.min_tokens": 800}}`. Only keys the
@@ -352,6 +382,11 @@ func ApplyForm(doc string, f Form) (string, error) {
 
 	pipeline := append([]string(nil), f.Pipeline...)
 	pipeline = applyExtractLLMCoupling(pipeline, f.Components["extract_llm"])
+	// Before the component loop, not after: a preserved component is ON, and the loop below
+	// CLEARS the declared keys of anything it reads as switched off. Resolving the pipeline
+	// once, here, is what stops a save preserving a component in the run order while deleting
+	// the block that configures it.
+	pipeline = preserveUnmodelled(was, pipeline, f.Pipeline, f.PipelineKnown)
 
 	comps := child(m, "components")
 	for name, decls := range components.AllFields() {
@@ -438,6 +473,45 @@ func applyExtractLLMCoupling(pipeline []string, vals map[string]any) []string {
 		return insertBefore(pipeline, name, "extract")
 	}
 	return pipeline
+}
+
+// preserveUnmodelled re-inserts every component the STORED pipeline ran that the posted one
+// does not mention AND the client did not declare a control for.
+//
+// This is the rule that makes a save unable to drop what the client cannot see. The posted
+// pipeline is a wholesale replacement built from a checkbox grid, so a name missing from it is
+// ambiguous: the operator may have unticked it, or the page may never have drawn a box for it
+// (a cached bundle whose /api/options predates the component) or may have drawn it from a
+// document that has since changed. `known` disambiguates — it is what the client says it drew —
+// and absence from `known` resolves the ambiguity in favour of the stored document.
+//
+// Position is preserved, not appended: a pipeline is an ORDER, and `linecap` reappearing at the
+// end is a different configuration from `linecap` where the operator put it. Each survivor goes
+// back at the index it held in `was`, which for the common case of one unmodelled component in
+// the middle restores the document exactly.
+//
+// A DECLARED removal still removes: a name in `known` and absent from `sent` is gone.
+//
+// `sent` is what the CLIENT posted and `resolved` is that after applyExtractLLMCoupling, and
+// the two are deliberately different arguments. The membership test is on `sent`, because this
+// rule is about what the client said; the insertion is into `resolved`, because that is the
+// pipeline being written. Testing membership on the resolved list instead would undo the
+// coupling's own removal of extract_llm — a SERVER decision, not a client omission — and put
+// a component back that its own constructor refuses to build.
+func preserveUnmodelled(was, resolved, sent, known []string) []string {
+	out := append([]string(nil), resolved...)
+	for i, name := range was {
+		if contains(out, name) || contains(sent, name) || contains(known, name) {
+			continue
+		}
+		// Clamped, because `was` may be longer than what is being written.
+		at := i
+		if at > len(out) {
+			at = len(out)
+		}
+		out = append(out[:at:at], append([]string{name}, out[at:]...)...)
+	}
+	return out
 }
 
 // normalize fills enum fields a caller left EMPTY with the component's own default, and
@@ -687,14 +761,26 @@ func delPath(m map[string]any, path string) {
 	}
 }
 
-// child returns m[key] as a map, creating it when absent and replacing it when it is
-// something else (a `components:` that decoded as nil because the key was written with no
-// body is the common case).
+// child returns m[key] as a map, creating it when absent — INSERTED INTO m — and replacing
+// it when it is something else (a `components:` that decoded as nil because the key was
+// written with no body is the common case).
+//
+// The insertion is the whole contract and it used to be missing: the function returned a
+// fresh detached map and left m untouched, so a caller that did not assign the result back
+// wrote its whole block into a throwaway. `components:` and setPath both happened to assign
+// it back and were fine; the `cache:` block did not, and a tenant switching the keep-alive on
+// for the first time — the one case where the key is absent — had their consent silently
+// discarded. Fixed here rather than at that one call site: the next block someone adds to
+// ApplyForm cannot repeat it. The three callers that do assign are unaffected, and the two
+// that prune an empty block (`components:` and each component's own) still prune it, so an
+// empty map created here does not reach the document.
 func child(m map[string]any, key string) map[string]any {
 	if sub, ok := m[key].(map[string]any); ok && sub != nil {
 		return sub
 	}
-	return map[string]any{}
+	sub := map[string]any{}
+	m[key] = sub
+	return sub
 }
 
 func contains(xs []string, v string) bool {
--- a/config/form_test.go
+++ b/config/form_test.go
@@ -603,8 +603,14 @@ func TestSwitchingAComponentOffClearsItsDeclaredKeysAndNothingElse(t *testing.T)
 			if err != nil {
 				t.Fatalf("configuring every key at once failed: %v", err)
 			}
-			// Now switch it off: out of the pipeline, values still posted.
+			// Now switch it off: out of the pipeline, values still posted. The removal is
+			// DECLARED — PipelineKnown is what the client says it modelled, and a save may
+			// only remove what it claims to have drawn a control for (see
+			// preserveUnmodelled). This caller read the whole document, so it claims all of
+			// it; a client that claims nothing removes nothing, which is the case
+			// TestASaveFromAClientThatDeclaresNothingRemovesNothing covers.
 			off := mustParse(t, configured)
+			off.PipelineKnown = append([]string(nil), off.Pipeline...)
 			off.Pipeline = remove(off.Pipeline, name)
 			if name == "extract_llm" {
 				// The one component with a coupling: it is switched off by its two
@@ -672,8 +678,18 @@ func TestTheColdCacheKeysSurviveASaveThatDoesNotMentionThem(t *testing.T) {
 			t.Errorf("a plain re-save lost %q:\n%s", want, out)
 		}
 	}
-	if paths := changedPaths(t, osherDoc, out); len(paths) != 0 {
-		t.Errorf("re-saving an unchanged form moved %v", paths)
+	// The `cache:` block is the one thing a re-save MATERIALISES, and that is the point of
+	// Bug 1's fix: the two consent booleans are written explicitly, so an absent block stops
+	// being indistinguishable from a stored `false`. Before the fix they were written into a
+	// detached map and thrown away, which is how a first-time keep-alive opt-in was lost.
+	// Nothing else may move, and no tuning field may appear — writing zeroes for those would
+	// freeze today's defaults into the document (R3).
+	for _, path := range changedPaths(t, osherDoc, out) {
+		switch path {
+		case "cache.keepalive", "cache.head_ttl_1h":
+		default:
+			t.Errorf("re-saving an unchanged form moved %s", path)
+		}
 	}
 }
 
--- a/config/formbugs_test.go
+++ b/config/formbugs_test.go
@@ -1,0 +1,173 @@
+package config
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+
+	_ "github.com/rossoctl/context-guru/components/all"
+	"gopkg.in/yaml.v3"
+)
+
+// Two data-loss bugs in the settings save path, and the tests that fail without their fixes.
+//
+// Both are the same shape: a save wrote something the operator did not ask for and there was
+// no test at the layer where the decision is made. The first threw away a consent the user
+// had just given; the second threw away a component the client could not see.
+
+// Bug 1: the keep-alive opt-in did not persist on a document with no `cache:` block.
+//
+// The one case that breaks, and the reason it went unnoticed for so long: on a document that
+// ALREADY has a cache block every field round-trips correctly, because child() then returns
+// the real map. Absent key, detached map, whole block discarded — and "switching keep-alive on
+// for the first time" is by definition the absent-key case.
+//
+// Asserted through a ParseForm round trip rather than a substring. A test for
+// `"keepalive: true"` in the output text would also pass on a document that persisted the
+// consent and dropped something else.
+func TestSavingKeepAliveOnADocumentWithNoCacheBlockKeepsIt(t *testing.T) {
+	const doc = "pipeline:\n  - format\n  - extract\n"
+	if strings.Contains(doc, "cache") {
+		t.Fatal("the fixture must have NO cache block; that is the only case that breaks")
+	}
+	// Every field of CacheForm, one case each, so a field added later is covered by the same
+	// table. reflect over the struct rather than a hand-listed set: a hand-listed set is how
+	// the form drifted from its own declarations in the first place.
+	full := CacheForm{
+		KeepAlive: true, KeepAliveIdleSeconds: 280, KeepAliveMaxPings: 2,
+		KeepAliveMaxUSDPerPing: 0.05, KeepAliveMinPrefixTokens: 20000,
+		HeadTTL1h: true, HeadTTLMinTokens: 50000,
+	}
+	rt := reflect.TypeOf(full)
+	for i := 0; i < rt.NumField(); i++ {
+		name := rt.Field(i).Name
+		t.Run(name, func(t *testing.T) {
+			// One field at a time, on an otherwise zero form: a field that only survives
+			// because a neighbour happened to create the block is not a field that works.
+			var one CacheForm
+			reflect.ValueOf(&one).Elem().Field(i).Set(reflect.ValueOf(full).Field(i))
+			out, err := ApplyForm(doc, Form{Pipeline: []string{"format", "extract"}, Cache: &one})
+			if err != nil {
+				t.Fatalf("ApplyForm: %v", err)
+			}
+			got, err := ParseForm(out)
+			if err != nil {
+				t.Fatalf("ParseForm: %v", err)
+			}
+			if got.Cache == nil {
+				t.Fatalf("the whole cache block was discarded; the saved document is:\n%s", out)
+			}
+			if *got.Cache != one {
+				t.Errorf("%s did not round-trip: sent %+v, read back %+v\nsaved document:\n%s",
+					name, one, *got.Cache, out)
+			}
+		})
+	}
+}
+
+// Bug 1's CLASS, on the helper. Three lines, and it is the assertion that generalises: the
+// next block added to ApplyForm inherits it for free.
+func TestChildAttachesTheBlockItCreates(t *testing.T) {
+	m := map[string]any{}
+	child(m, "x")["k"] = 1
+	sub, ok := m["x"].(map[string]any)
+	if !ok {
+		t.Fatalf(`child(m,"x") did not insert a map into m; m = %#v`, m)
+	}
+	if sub["k"] != 1 {
+		t.Errorf("the returned map is not the one in m: m[x] = %#v", sub)
+	}
+	// And it still adopts what is already there rather than replacing it.
+	m2 := map[string]any{"y": map[string]any{"keep": true}}
+	child(m2, "y")["added"] = true
+	y := m2["y"].(map[string]any)
+	if y["keep"] != true || y["added"] != true {
+		t.Errorf("child replaced an existing block: %#v", y)
+	}
+}
+
+// Bug 2: a save must never drop a component the client did not render a control for.
+//
+// The reported incident: `linecap` vanished from an operator's pipeline and `toon` was
+// re-added at the end. The server writes the posted pipeline wholesale, so a stale render or
+// a cached bundle whose /api/options predates a component produces exactly that — and the
+// server could not tell it from a deliberate removal. PipelineKnown is the client CLAIMING
+// what it drew, so absence from it resolves the ambiguity in favour of the stored document.
+func TestASaveNeverDropsAComponentTheClientDidNotRender(t *testing.T) {
+	const doc = "pipeline:\n  - format\n  - linecap\n  - extract\n"
+	// The client draws format and extract, does not know linecap, and posts without it.
+	out, err := ApplyForm(doc, Form{
+		Pipeline:      []string{"format", "extract"},
+		PipelineKnown: []string{"format", "extract"},
+	})
+	if err != nil {
+		t.Fatalf("ApplyForm: %v", err)
+	}
+	if got := pipelineOf(t, out); !reflect.DeepEqual(got, []string{"format", "linecap", "extract"}) {
+		t.Errorf("linecap was not preserved at its own index: got %v\n%s", got, out)
+	}
+	// And a DECLARED removal still removes. Without this the fix would be "the pipeline is
+	// append-only", which is a different bug.
+	out, err = ApplyForm(doc, Form{
+		Pipeline:      []string{"format", "extract"},
+		PipelineKnown: []string{"format", "linecap", "extract"},
+	})
+	if err != nil {
+		t.Fatalf("ApplyForm (declared removal): %v", err)
+	}
+	if got := pipelineOf(t, out); !reflect.DeepEqual(got, []string{"format", "extract"}) {
+		t.Errorf("a declared removal did not remove: got %v\n%s", got, out)
+	}
+}
+
+// The old-bundle direction: a client that declares nothing removes nothing. A cached app.js
+// and a hand-rolled API client are the same case, and both can still add and reorder.
+func TestASaveFromAClientThatDeclaresNothingRemovesNothing(t *testing.T) {
+	const doc = "pipeline:\n  - format\n  - linecap\n  - extract\n  - toon\n"
+	out, err := ApplyForm(doc, Form{Pipeline: []string{"format"}})
+	if err != nil {
+		t.Fatalf("ApplyForm: %v", err)
+	}
+	got := pipelineOf(t, out)
+	for _, want := range []string{"format", "linecap", "extract", "toon"} {
+		if !contains(got, want) {
+			t.Errorf("%s was removed by a client that declared nothing: got %v", want, got)
+		}
+	}
+}
+
+// A preserved component keeps its CONFIGURATION too. The component loop reads the resolved
+// pipeline to decide what is switched off, and switched-off clears the declared keys — so
+// preserving a name while wiping its block would leave a run order the document cannot explain.
+func TestAPreservedComponentKeepsItsOwnBlock(t *testing.T) {
+	const doc = `pipeline:
+  - format
+  - linecap
+components:
+  linecap:
+    max_line_chars: 4000
+`
+	out, err := ApplyForm(doc, Form{
+		Pipeline:      []string{"format"},
+		PipelineKnown: []string{"format"},
+		Components:    map[string]map[string]any{"linecap": {"max_line_chars": 4000}},
+	})
+	if err != nil {
+		t.Fatalf("ApplyForm: %v", err)
+	}
+	if !strings.Contains(out, "max_line_chars: 4000") {
+		t.Errorf("the preserved component's configuration was cleared:\n%s", out)
+	}
+}
+
+// pipelineOf reads the pipeline back out of a saved document.
+func pipelineOf(t *testing.T, doc string) []string {
+	t.Helper()
+	var d struct {
+		Pipeline []string `yaml:"pipeline"`
+	}
+	if err := yaml.Unmarshal([]byte(doc), &d); err != nil {
+		t.Fatalf("the saved document does not parse: %v\n%s", err, doc)
+	}
+	return d.Pipeline
+}
--- a/dash/api.go
+++ b/dash/api.go
@@ -236,7 +236,10 @@ func (a *API) routes() []route {
 	// The tool/MCP/skill inventory's route, declared beside its handler in toolapi.go and
 	// appended here rather than mounted separately: this table is what the scoping test walks,
 	// so a route mounted around it would be a route whose scope nothing checks.
-	return append(rs, a.toolRoutes()...)
+	rs = append(rs, a.toolRoutes()...)
+	// The keep-alive tab's reads, declared beside their handlers in keepaliveapi.go and
+	// appended here for the same reason: this table is what both scoping tests walk.
+	return append(rs, a.keepAliveRoutes()...)
 }
 
 // Mount registers every dashboard route on a mux under the given prefix
--- a/dash/keepalive.go
+++ b/dash/keepalive.go
@@ -249,17 +249,19 @@ func (d *DB) windowDays(cond string, args []any) float64 {
 // Band is one bucket of an ORDERED histogram: idle-gap bands, prefix-size bands, hour bins.
 //
 // Label carries the meaning and the order is the data — a bar chart of these must preserve it
-// and must not sort by size, because the order IS what the reader is reading. Since/Until make
-// each bar a drill-down into the requests behind it.
+// and must not sort by size, because the order IS what the reader is reading.
+//
+// Deliberately NO since/until here. These bands bucket by GAP LENGTH, by PREFIX SIZE and by HOUR
+// OF DAY — none of which is a time range, so a bar's drill-down is `reason=ttl_expiry` plus the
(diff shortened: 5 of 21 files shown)


OsherElhadad:
## Review: CHANGES-REQUIRED

The engineering is good and both config bugs are fixed correctly at the root — I mutated each fix
and confirmed it goes red. What blocks this is not code quality. **The tab's flagship control does
not work at all**, **the page states two refuted numbers as its own justification** and
contradicts itself two panels away, **one agent aggregate is still contaminated by ping rows**,
and **the arm path reports a $0.00 per-ping ceiling while enforcing none**. All four are in the
class this PR exists to police.

Ten findings; F1–F5 blocking. Every number below is from my own run: full suite green
(25 packages, `-timeout 25m`, `dash` 639.7 s), `gofmt` empty, `vet` clean, and a scrubbed copy of
the production corpus (19,805 requests) with a 1,060-ping replay of the shipped policy.

---

### F1 — BLOCKING. `Keep warm` throws. The per-session override is unreachable from the UI.

`app.js:7654` calls the global `prompt(...)`. `tools.js:1358` declares

```js
const prompt = { state: 'idle', view: null, byName: new Map() };
```

at the top level of a **classic** script — a global *lexical* binding, which shadows
`window.prompt` for every bare `prompt(...)` in every classic script on the page, `app.js`
included, regardless of load order. In Chromium 131 against the served bundle:

```
typeof window.prompt = function
typeof prompt        = object          <-- the shadow
armSession throws:   TypeError: prompt is not a function
```

Clicking any of the 21 arm buttons: no dialog, no alert, nothing armed,
`GET /api/me/keepalive/sessions` → `{"armed":[]}`. **Silent.** The server is fine — the same arm
via `fetch` returns 200. So panel 5 of 6, and all of design Part 2, ships dead.

Same root cause kills three more call sites, two of them new here:

| `app.js` | dead |
|---|---|
| `:7654` | `armSession` — **new in this PR** |
| `:6213` | "Copy the document as it was before this save" — **new in this PR**, and this PR's stated mitigation for comment loss |
| `:6943`, `:6948` | "Name this token" / "Copy this token now — it cannot be recovered". **Pre-existing** (the shadow landed with #93). A token is shown once and the showing never happens. |

`tools.js` is untouched here, so the shadow isn't this PR's doing — but this PR adds two call
sites into it. Rename it (`promptView`, matching `promptWaiters` beside it) and add one static
check that no `dash/ui/*.js` declares a top-level binding shadowing a `window` method it calls.
A `data-testid` click on `ka-arm-*` would also have caught it.

### F2 — BLOCKING. Two refuted numbers, and the page contradicts itself.

`app.js:8016` and `TILE_INFO['ka-rec-policy'].catch` (`:885`):

> One ping is never suggested: it reaches about **4.7 minutes**, which is inside the free
> five-minute lifetime, and on this service it is **$71 worse** than not running at all.

- **"4.7 minutes"** is `K·X` (280 s). The reach is `K·X + TTL` = **580 s = 9.67 min**, and the
  page's own K-ladder three panels up prints `580s = 1×280 + 300`. `dash/keepalive.go:722` carries
  the warning that `K·X` *"is the single error that made an earlier analysis wrong by a factor of
  4.4"* — the code enforces the right formula, the copy explains itself with the wrong one.
- **"$71 worse"** is the superseded figure. The adjudication's own table gives K=1 at X=280 as
  `−71.39` **as published by the analysis it overturned**, `−72.22` under the wrong coverage, and
  **+$101.56** under correct coverage. K=1 is positive at every X ≥ 180 there.
- And this page's calculator reports **K=1 net = +$43.17**, directly above the sentence saying
  K=1 loses $71.

The *rule* survives — K=2 dominates K=1 in the adjudicated sweep at every X. Keep the behaviour,
replace the reason: *"two pings reach nearly twice as many of your expiries (51% vs 30% of your
reachable dollars) for a proportional cost."* The design doc carries the same phrasing in Parts 6
and 11; it needs the same correction.

### F3 — BLOCKING. A ping row still contaminates an agent aggregate: `prefix_change`.

`overview.go:424-435`. The prefix-change diagnostic asks whether *the session's previous turn*
mutated something:

```sql
SELECT p.id FROM requests p
WHERE p.session_id = r.session_id AND (p.ts < r.ts OR (p.ts = r.ts AND p.id < r.id))
ORDER BY p.ts DESC, p.id DESC LIMIT 1
```

No `p.keepalive = 0`. A ping becomes "the previous turn", has no `request_components` rows, so
`EXISTS(... mutated = 1 ...)` is false and the row drops out. **This is the same defect, in the
same SQL statement, twenty lines above the one the PR fixes.**

The key-by-key comparison across every mounted read route, with and without the 1,060 ping rows:

```
DIFF /api/stats                          diffs=8  unexpected=2
      prefix_change_cost_usd: without=101.81492917  with=97.48978051
      prefix_change_requests: without=94            with=92
OK   /api/series, /api/components, /api/facets, /api/sessions
OK   /api/breakdown x model provider agent preset mode session accounting cache_miss stop_reason tenant
```

Direct confirmation: `n=92 / $97.4898` shipped vs `n=94 / $101.8149` with the predicate — a 4.2%
under-report, from 4 `prefix_change` rows whose previous row is now a ping. $101.81 is exactly the
established "49.7% of prefix-change cost is adjacent to one of our own mutations", so enabling the
keep-alive makes context-guru look **less** responsible for prefix changes than it is — the
flattering direction, on the one diagnostic that measures our own harm.

`TestPingRowsStayOutOfAgentAggregates` misses it because the fixture has no `prefix_change` row
adjacent to a ping. One predicate plus one fixture row.

Everything else in the family is safe and I checked each: `splitMoved` (`p.split_tail_hash <> 0`),
`CompactionResets` (`p.tokens_before > 0`) and `cachehistory.go:67` all exclude pings by
construction, because a ping Event never sets those fields.

**One more, pre-existing, please check:** `capture.go:547-550` `SeedSessions` takes each session's
**latest** row with no `keepalive` predicate, so a ping as the last row re-dates `lastSeen` across
a restart — which is precisely what `keepalive.go:882` says must never happen (*"never touches the
session-recency map: doing so would make the NEXT real request's gap read as four minutes instead
of the twenty it actually was"*). The majority of pings under a blanket policy are session-final,
and this corrupts the `ttl_expiry` attribution this whole tab rests on. My injected data does not
exercise it, so I have no number — only the code.

### F4 — BLOCKING. Arming reports a $0.00 per-ping ceiling and enforces none.

`cmd/context-guru-proxy/main.go:720` returns the **zero** `CachePolicy` when the account's
keep-alive is off, so `MaxUSDPerPing = 0`; `keepalive.go:301` treats `<= 0` as **no cap**;
`overrideFor` deliberately leaves the field alone.

So on an account whose keep-alive is off — which design Part 2 names as *the* use case (*"it is
how someone tries it without a blanket spend"*) — an armed session pings with **no per-ping cost
guard at all**. The guard exists because ping cost is bimodal, p99 $0.2275, max $0.3780, against a
$0.25 default that never applies here. Observed:

```
POST /api/me/keepalive/sessions -> 200
{"max_usd_per_ping":0, "min_prefix_tokens":0, "hold_minutes":14, ...}
```

`keepalivectl.go:176` therefore tells the operator the ceiling is **$0.00** while enforcing
infinity — worse than omitting it. The design says the guard may not be widened; by omission it is
widened without bound, in exactly the case it was written for. Resolve it from
`config.DefaultKeepAliveMaxUSDPerPing` when `<= 0`, and never report 0. The same response silently
drops the 20k prefix floor to 0 without saying so.

### F5 — BLOCKING. The calculator's `PINGS`/`NET` are not a policy anyone could run, and the sign flips.

`keepalive.go:834` builds `spans` from `LAG(ts)` within a session, so a span exists only *between*
two requests. It charges **no pings** for a session-final request — where a live policy must send
K, since it cannot know the session ended — and applies **no gate**, charging pings on the turn-0
and small-prefix spans `turn ≥ 1 AND prefix ≥ 20k` would never touch. Replayed at X=280, K=2:

| policy | pings |
|---|--:|
| the calculator (no gate, no session-final) | **1,452** |
| a blanket live policy (no gate, session-final charged) | **9,234** |
| the shipped gate, session-final charged | 1,060 |

9,234 reproduces the adjudicated blanket count to the ping, so the replay is sound. The calculator
under-counts the policy it nominally models by **6.4×**. Since `NET = SavedUSD − pings ×
PingUSDEach`, the panel's **+$70.41** at K=2 becomes **−$785** with the pings a blanket policy
actually sends. **The sign of the headline flips**, and F2's "+$43.17 at K=1" is positive only
because of this.

**The same omission sits inside the recommendation's bootstrap, where it decides whether to
recommend at all.** `keepalive.go:1085` resamples only the sessions that had an addressable
expiry — **53 of 3,891** here — and charges ping cost only from those, omitting **42.4% of even
the intra-session ping cost** (837 of 1,452) on top of every session-final ping. Both biases push
the interval **upward**, i.e. toward firing when the rule should refuse; the route returns
`[$101.33, $263.05]` where the adjudicated service-wide interval is `[$95, $237]`.

The spec asked for `Σ over the account's own spans`, so the implementation is faithful — but this
is a dollar figure on a page whose purpose is not to publish one it cannot defend, and there is no
caveat naming which pings are counted. Either charge session-final spans and apply the gate, or
**drop `PING COST` and `NET`** and keep reach only, which is all the spec's golden table pins.
The second is smaller and honest.

---

### Nits

**F6.** `app.js:7630` — *"The top 8 of these hold $428.26 of **$649.27**"*. `$649.27` is the sum of
the 20 rows fetched; the account's addressable total is **$732.10** and is already on the page in
the "Still on the table" tile. Concentration reads 66% where it is 58%, and the denominator moves
with `limit`. The `$81.40 of $81.40` tautology was this same defect degenerate — the numerator got
fixed, the denominator did not. Spec Part 5 states the intended figure. One line:
`const all = (kaState.ledger || {}).addressable_usd;`

**F7.** `kaState.canArm` is cleared only on a 404 (single-tenant). In a manager's **service-wide**
view every row belongs to another tenant, arming keys on the authenticated principal by design, so
the override lands inert and the response is a cheerful 200 — with
`last_prefix_tokens: 0, priced: false` because `worstCase` looks the prefix up under the *arming*
principal's filter, for a session whose row shows a 987k prefix. The alert then says *"at worst 8
pings — this model is not on the price list, so the cost cannot be stated"* for something that will
never ping. This is the defect `0a90dcf` fixed one case short of. Disable the button unless
`s.tenant_id === account.tenant.id`, and 400 rather than arming nothing and reporting success.

**F8.** `keepalivectl.go:202` — `worst_case_pings = (until / ((K+1)·X)) × K` is not a ceiling. A
span ends when a real request arrives, so the worst case is requests just after each ping: `until /
X` pings. Over 1 h at the defaults, reported **8**, true **12**. Ratio `(K+1)/K` — 2× at K=1.
Labelling an under-estimate a CEILING on a spend authorization is the wrong direction.

**F9.** `maxOverrideHold` (60 min) is unreachable: `(K+1)·X` maxes at `12 × 290 s = 58 min`.
`TestOverrideRespectsTheCredentialHoldCeiling` says so mid-test (*"reach the ceiling through X"*)
and then cannot, and ends up asserting `hold()` arithmetic on **accepted** overrides. Design test
14 is unmet in both halves. Delete the constant and state 58 min as a consequence of the two live
caps, or drop it from the cap table.

**F10, grouped.**
- The `EXISTS` gate buys nothing. `keepalive` is in no index, so with zero pings it is a full scan.
  Timed alternately on the 10k fixture, 5 trials: `EXISTS` alone **2.3 ms**; gated **52.7 ms** vs
  ungated (always correcting) **53.0 ms** at 0 pings, 58.0 vs 57.3 at 100. Drop the gate or drop
  the *"costs literally nothing"* claim.
- `overview.go:527` says *"three pings ... 30 to 70 tokens"*. The test has **one** ping and the
  value is **50** — I reproduced it by deleting the correction. The PR body has it right.
- The widened re-save test can no longer detect Bug 1 **in either direction**. `changedPaths`
  compares path *names*, and the new switch neither requires the two paths to move nor pins them to
  `false`. With `child()` reverted it passes; with `cb["keepalive"] = true` forced (a save flipping
  consent **on**) it also passes. The sibling test catches the second, so this is a nit — but the
  widening should have been exact: assert the changed set **is** those two paths and that both
  re-parse to `false`. A test loosened to accommodate a change is how regressions get in; this one
  lost its teeth rather than narrowing them.
- `sameNames`'s comment insists order is part of the meaning; replacing it with a multiset
  comparison keeps the whole `proxy` package green. A reorder-only stale render sails through the 409.
- The `EXISTS` gate's unscopedness is correct and load-bearing but untested — a window-scoped gate
  keeps the shipped suite green (see below).
- `Not enough history to recommend. **y**ou have 0 cache expiries...` — lower case after a full stop.
- `keepalivectl_test.go:65` argues with itself mid-sentence: *"`extract` survives because the client
  did not claim to have drawn it... it did claim it, so it is removed."*
- `keepaliveapi.go:118` sets `prefix_source = "account_median"` even when the median came back 0.
- `CalcRow.PingUSD/SavedUSD/NetUSD` have no `omitempty`, so they go on the wire as `0` when
  `priced:false`. The UI gates on `priced`, but *"a field that exists gets rendered"* is this
  project's own rule.
- "46 new tests": I count 45 top-level `func Test` in the five new files, and the PR's own breakdown
  sums to 44.

---

### The five deviations — all five are right

**(a) One-sided recommendation rule: correct, and it is a spec fix rather than a deviation.** Read
two-sidedly, `[−38.63, −4.54]` "excludes zero" and would be recommended. Part 6 opens with
*"Recommend only when..."* and Part 7's subject is not raising the bill, so the intent is
unambiguous. The refusal fires, has its own wording, and is tested. The rule is also conservative
in the right direction: a two-sided 90% interval with `lo > 0` is a 95% one-sided test.

**(b) `PipelineKnown` empty ⇒ remove nothing, callers made to state their claim: correct, no
over-claim.** All three production `ApplyForm` callers are handled and I found no fourth.
`toolfilterctl` claims exactly the one component it models — and cannot over-claim, since every
other stored name is already in its `f.Pipeline`. The other two claim the whole pipeline they just
read, which is the honest statement for a caller that read the document.

I also built the adversarial test for the interaction you flagged, and it holds: with the client
sending `extract_llm` while its booleans switch it off, `applyExtractLLMCoupling`'s removal
**stands**. Deleting `contains(sent, name)` turns the existing
`TestApplyFormRemovesTheComponentWhenBothSwitchesAreOff` red, so the clause is load-bearing and
already covered. The remaining shape — a client declaring nothing that posts a switched-off
`extract_llm` — is refused by `Validate` with the component's own message, so no invalid document
can be stored. A dedicated regression test would still be cheap.

**(c) `Band.since/until` dropped: correct.** The bands bucket by gap length, prefix size and hour
of day; none is a time range, and Part 1's drill-down spec was wrong about its own data. Two
always-zero fields on the wire is how a date gets rendered from nothing.

**(d) `AccountMedianPrefix` also returning the modal model: correct and necessary.** I reproduced
both states. With no price list the route returns `priced:false` and refuses to state a dollar;
with one it prefills `aws/claude-opus-5` and 289,161 tokens (spec p50 292,527) and prices
`ping_usd_each = 289161 × 0.1 × 3.8e-6 + 19e-6 = $0.10990`, `avoided_usd_each = 289161 × 1.15 ×
3.8e-6 = $1.26363` — both exactly the spec formula. The modal model of the account's own expiries
is an answer about that account; a blend would not be. See F10 on `omitempty`.

**(e) `PingsPerSpan(gap = 2X) = 2`: correct in isolation, and swamped by F5.** Counting the racing
ping over-states spend, which is conservative, and cannot touch `Convertible`. But it is a
conservative rounding inside a function whose caller then omits 84% of a live policy's pings — a
documented conservatism buried under an undocumented 6.4× optimism. Fix F5 and (e) stands.

---

### The honesty requirements do genuinely surface

Served against a copy of a copy of the snapshot (every content table emptied, tenant and session
ids re-hashed), in both themes, at UTC and UTC+05:30, on a populated and a never-ran database.

All five Part 12 items are on the page. Every headline is a range with its n
(`$101.33 – $263.05 · 90% interval from 356 expiries in 53 sessions`) and the wire carries no point
estimate. Thin history is refused, naming the count and the service-wide interval for scale. The
saving is stated as a ceiling in the tile sub-line, in the session table's own column header
(`AVOIDED (CEILING)`), in the waterfall description and in `TILE_INFO`. The losing majority renders
**unconditionally on a positive net** (`+$105.89`), with the service-wide 34 of 119 and $2.42 and
this account's own 38/87 labelled in words beside the colour. Inter-expiry gaps are per-account
only, 0.64 h – 6.56 h across 11 rows.

The three coverage states are distinct and none fabricates a zero. Band labels are fixed
(`5m–9.7m`). The K-ladder table and its bars both render. The hour bins are genuinely rotated
client-side, not relabelled: at `TZ=Asia/Kolkata` the header reads `UTC+05:30` and UTC 19:00's
bucket appears at `00:30`.

**Of the six claimed mutation-proven UI honesty checks, five bite and one does not.**

| mutation | result |
|---|---|
| gate the split on a negative net | RED ✓ |
| drop `34 of 119` from the markup | RED ✓ |
| rename a `ka-` `TILE_INFO` key | RED ✓ |
| `keepalive` into `UNFILTERED_VIEWS` | RED ✓ |
| collapse the two `lineChart` calls | RED ✓ |
| **render the recommendation as a point** | **GREEN — not enforced** |

`TestTheRecommendationRendersARangeAndRefusesThinHistory` only asserts that the strings
`rec.lo_usd` and `rec.hi_usd` appear. Replacing the tile value

```diff
-      usd(rec.lo_usd) + ' – ' + usd(rec.hi_usd),
+      usd((rec.lo_usd + rec.hi_usd) / 2),
```

with a single hero midpoint leaves the **entire `dash` package green**. That is Part 12 item 1 and
the "never a hero number" rule, unenforced by the check written to enforce it. Pin the separator.

`TestEveryKeepAliveTileHasAnExplanation`'s ceiling clause is also weak: it searches the whole of
`app.js` for `"CEILING"` / `"content-keyed"` / `"content, so"` and returns on the first hit — seven
occurrences, one unrelated. Neutralising the wording in `ka-saved` and `ka-net` keeps it green. The
copy really is on screen, so this is a test-strength nit only.

### The ceiling correction, verified independently

Exact. On my scrubbed corpus with 1,060 pings:

```
uncorrected 198,405,117 | correction 8,939,690 | corrected 189,465,427 == the API's value
over-count 8,939,690 tokens = 4.72%
replay_realized_pct  5.550% (uncorrected) -> 5.812% (corrected)
```

and `replay_projected_tokens` is **identical with and without** the pings. The 517k/3.4% figure is
the same effect on a smaller database. The algebra holds too: the correction equals
`Σ_{r∈cond, saved>0} saved_r × |later pings in session|`, precisely the ping subset of the main
count with the same `(ts, id)` tie-break — exact by construction, not approximately. I wrote two
harder tests than the shipped one (two sessions, pings before and after reductions, a window that
clips rows; and a window containing no ping at all) and both pass.

**The `EXISTS` gate cannot silently skip**, and it is deliberately unscoped, which is load-bearing:
the main query's inner count spans the whole session, so a ping *outside* the window still inflates
it. A window-scoped gate returns 15 where the answer is 5 — and the shipped suite does **not**
catch that mutation. Correct as shipped, untested; happy to hand over the test.

**The performance claim holds and is stronger than stated.** Alternating trials on the 10k fixture:
`main` 50.6 ms, the inline `AND p.keepalive = 0` **100.3 ms** (2.0× on the query itself, not 1.2×),
shipped 52.7 ms. Both figures are consistent — the PR timed whole-`Overview`.

### Aggregates, schema, storage, conventions

- PR #86's `r.keepalive = 0` is intact and sufficient everywhere except F3 — the key-by-key sweep
  is the evidence. The waterfall reconciles exactly: signed non-total steps `−115.058120` against
  `total_saved_usd = 115.058120`.
- **Schema stays 6.** Opening my own copy of the snapshot (which predates #86's columns) with the
  branch binary: version `6`, three `keepalive*` columns added additively, **19,805 of 19,805 rows
  survived**, no new table, no fourth column.
- **Storage is over-stated in the safe direction.** `dbstat` page-delta between the same corpus with
  and without 1,060 ping rows: **216 B table + 35 B index = 251 B per ping row** against a claimed
  ≤400 B, because entries land in partially-full pages. ~82 KB/day at the shipped gate is a
  conservative bound, and the tile's own `237 × ~400 B = 95 kB/day` checks out. (Do not use the
  626 B all-rows average — a ping row is far narrower.)
- **Consent asymmetry works.** Non-manager: `POST` arm **403**; `DELETE /api/me/keepalive`
  **200 `{"keepalive":false}`**; `DELETE .../sessions/{id}` **200**. Part 7's gap is genuinely
  closed. Caps refuse with reasons and never clamp: idle 400 s → 400, `max_pings` 20 → 400,
  `until` +13 h → 400.
- **The content gate bites on the new routes.** Making `/api/keepalive` serve a real `PromptViewFor`
  ungated turns `TestNoRouteServesContentTextFromAnUntrustedAddress` red on both markers. (An
  *empty* text field does not fail it — correct: it catches leaked content, not string fields.)
- All 8 commits: correct author/committer, `Signed-off-by:` present, conventional titles, **no AI
  attribution anywhere** in the diff, messages or body. **No secrets or PII**: no account ids, no
  host paths, no screenshot path, no key shapes; the only emails are `boss@ibm.com` / `a@ibm.com`
  test fixtures whose shape already exists on `main`. The permission layer's block was correct and
  nothing leaked.
- `gofmt -l .` empty, `go vet ./...` clean, `CGO_ENABLED=1 go test -race -count=1 -timeout 25m ./...`
  green across 25 packages. `dash` 639.7 s here; the 25-minute timeout is required and is not this
  branch's doing.

Not merged, nothing pushed.

PR #102, opened 2026-08-23
fix(dash): stop tools.js shadowing window.prompt, which killed "Mint a token"

OsherElhadad:
## What was broken

Settings → **Mint a token** did nothing when clicked. No dialog, no request, no token, no
error on screen. The button, its click handler, its route (`POST /api/me/tokens`), its scope
gate and `tenant.MintToken` were all correct and all covered by passing tests.

## Root cause

`dash/ui/index.html` loads `app.js` and `tools.js` with a plain `<script src>` — **classic**
scripts, not modules — so they do not get a scope each. They share one global lexical scope.

`tools.js` declared a top-level binding for the Inventory tab's prompt-text cache (added in
#93):

```js
const prompt = { state: 'idle', view: null, byName: new Map() };
```

That `const` shadowed `window.prompt` for every **bare** `prompt(...)` call in *both* files.
Only the bare form — `window.prompt` itself stayed a function — which is why nothing looked
wrong in `app.js`, in the markup, in the endpoint, or in any existing test. The only symptom
was `TypeError: prompt is not a function`, thrown inside whichever handler called it.

Measured in the page, before the fix:

```
typeof prompt         -> 'object'      // the tools.js cache
typeof window.prompt  -> 'function'    // untouched
```

## Blast radius — four call sites in `app.js`, three broken features

| Where | What happened |
|---|---|
| Settings → **Mint a token** | threw on the handler's first statement: no dialog, no request, no token |
| Tenants → **Reissue token** (manager) | threw **after** its POST had already succeeded — the token was minted and then never shown. Tokens are stored hashed, so every click left that account one live credential nobody holds |
| Keep-alive → *"how many hours?"* | threw before it could ask, so arming a session did nothing |
| Settings → audit → *"copy the document as it was"* | a clipboard **fallback**, so only a browser that refuses clipboard access reached it |

## The fix

Rename the binding to `promptState` (`dash/ui/tools.js`, 17 lines). The Inventory tab is its
only user; nothing else referenced it. `window.prompt` is one function again for both scripts.

An audit of **all 302 top-level declarations** across both scripts against the real
`Window` property names found exactly one collision — this one.

## The guard

`TestUIScriptsDoNotShadowBrowserGlobals` (`dash/uishadow_test.go`) walks every top-level
declaration in both UI scripts and fails on any name the browser already defines. It targets
the **class** of defect, not this instance:

- it **fails on the original code** and passes on this branch (verified both ways);
- it refuses to pass vacuously — if the matcher stops finding declarations, it fails instead;
- a sibling test pins the premise, so if the dashboard ever moves to `<script type="module">`
  the guard says it can go.

## Verification

Built the binary and drove the real dashboard in headless Chromium against a hosted-mode
proxy, signed in as a real account.

**Before:** click → `pageerror: prompt is not a function`, 0 dialogs, 0 requests, token list unchanged.

**After:**

```
DIALOG prompt  "Name this token (e.g. laptop, ci):"  default="new-token"
RESP   POST /api/me/tokens -> 201
DIALOG prompt  "Copy this token now — it is shown once and cannot be recovered:"
               default="cg_live_BCXXWLQKXQYMOXNBK67B5YTF7M"
token list: 3 rows, newest first
```

The minted token is a real credential, not just a string: the proxy answers it with
`no provider credential; send your own API key` (i.e. it authenticated and stopped for lack
of an upstream key), while a bogus token gets `unknown or revoked context-guru token`.

Also confirmed in the page after the fix: `prompt === window.prompt` and both are functions,
which is what un-breaks the other three call sites. The renamed Inventory state object still
works — `loadPrompt()` transitions `idle → ok` and `repaintPrompt()` runs, with zero page
errors across the Inventory, Settings, Keep-alive and Overview tabs.

`go test ./...` green (25 packages), `go vet` clean, `gofmt` clean.

## Docs

- `docs/dashboard.md` — the "no CDN, no build step" box claimed the UI is **three** files; it
  is five. Corrected, and given the shared-global-scope constraint that this bug is an
  instance of.
- `docs/dashboard-inventory-page.md` — said `tools.js` "mounts itself" with "no edit to
  `app.js`", which reads as *has its own scope*. Now says what mounting itself does not buy.
- `dash/ui/tools.js` — the renamed binding carries a comment saying why it may not be called
  `prompt`, so it does not get renamed back.

### DIFF
--- a/dash/ui/tools.js
+++ b/dash/ui/tools.js
@@ -1354,8 +1354,28 @@ function copyBox(text, label) {
  *
  * `state` is one of: 'idle' (not asked for yet), 'loading', 'ok', 'absent' (the endpoint is
  * not there — an older proxy), 'error'.
+ *
+ * NOT named `prompt`, and nothing at this level may be. app.js and this file are both
+ * CLASSIC scripts, so they share one global lexical scope: a top-level `const prompt` here
+ * shadowed window.prompt for every bare prompt() call in BOTH files, while leaving
+ * window.prompt itself a function — so nothing looked wrong in app.js, in the markup or in
+ * any test, and the only symptom was a TypeError thrown inside a click handler. It shipped
+ * that way, and it broke three things by taking out four call sites in app.js:
+ *
+ *   - Settings → "Mint a token" threw on its FIRST statement, so the button did nothing at
+ *     all: no dialog, no request, no token.
+ *   - Tenants → "Reissue token" threw AFTER its POST had succeeded, so the replacement
+ *     token was minted and then never shown. Tokens are stored hashed, so each click left
+ *     the account one live credential nobody holds.
+ *   - the keep-alive's "how many hours?" threw before it could ask, so arming a session
+ *     did nothing.
+ *
+ * The fourth, the audit trail's "copy the document as it was before this save", is a
+ * clipboard fallback, so only a browser that refuses clipboard access reached it.
+ *
+ * TestUIScriptsDoNotShadowBrowserGlobals now fails on any such name.
  */
-const prompt = { state: 'idle', view: null, byName: new Map() };
+const promptState = { state: 'idle', view: null, byName: new Map() };
 
 /**
  * loadPrompt fetches the prefix text once and repaints only the reveals that are open.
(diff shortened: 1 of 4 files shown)


OsherElhadad:
Superseded — closing. The defect was real on df85918, but the same fix (the `promptView` rename, plus `TestNoUIScriptShadowsAWindowMethod`) had already landed on `main` in #100 while this branch was in flight. Verified against current `main`: Mint a token works end to end, and no top-level declaration in the UI scripts shadows a window global. Nothing here is still needed.

PR #117, opened 2026-08-27
feat(extract_llm): never ask the model to transport text while sweeping history

amiddavid:
Ports the reasoning behind removing the merged design's `trim` verdict (`cc1aa9f`, on
`feat/coref-compaction`) to the per-output path that `main` actually ships, and applies it where
it matters most: the cold-cache sweep.

## The principle, transferred

`cc1aa9f` removed `trim` because it was **the only verdict that asked the model to transport
text**, on this evidence: chosen zero times in 21 probe opportunities, metrics identical without
it, and in production accepted **once against eight rejected as invented**.

There is no `trim` verdict on `main` — it belongs to the merged contract, where one call disposes
of many outputs and needs a word per output. But the same division exists in the per-output
strategies, and nothing had drawn the line:

| Strategy | What the model returns | Transports text |
|---|---|---|
| `code` | a Starlark program, executed against the real text | no |
| `deterministic` | nothing — no model call | no |
| `single` | the selected JSON | **yes** |
| `rlm` | selected content, per chunk | **yes** |

A sweep now allows only the first group.

## Why the sweep specifically

Two reasons that compound:

- **`strategy: auto` orders `rlm` ahead of `code` once a body is large**, and large bodies are
  precisely what a sweep works on. So on a sweep the transporting mode is not a corner case, it is
  the default pick.
- **A sweep rewrites content deep in history** that the model has already reasoned about. An
  invention there contradicts the transcript, rather than degrading one recent tool result.

## Two details taken from the original rather than invented

**The narrowing is counted** (`sweep_strategy_narrowed`). This is the half of the trim change
that is easy to lose: there, a model answering `trim` degraded to `keep` *and was counted*,
because an unjudged output is otherwise indistinguishable from silence. Here, "the operator asked
for `rlm` and got `code`" must not read as "the operator asked for `code`".

**`auto` is excluded by absence, not by filtering.** It is a strategy *order*, and
`extract.intersectAllowed` narrows the order it produces — so an `auto` configuration still
resolves, to `code` then `deterministic`, instead of being emptied and failing open to no
compaction at all.

## Verification

Three tests. Two pin the classification against `extract.Modes`, the engine's own list of accepted
strategies, so a strategy added later cannot silently default to "does not transport" and have its
narrowing go uncounted. The third runs **one** `strategy: rlm` config over two turns that differ
*only* in whether the prompt cache is cold, so it asserts the sweep condition rather than
something general about `rlm`.

Verified by removing the narrowing: the cold arm fails with `gates=map[]`, the warm arm still
passes. Full suite: 27 packages, 0 failures, `gofmt` clean. No benchmarks run.

## Intentionally not in this PR

**Splitting `extract_llm` into `extract_llm` + `extract_llm_sweep`.** Agreed as the better design
and scoped, but it is a config-surface migration and belongs in its own PR. The sweep already
diverges from the tail path in nine places — its own floor (`cold_cache.min_tokens` replaces
`min_tokens`), its own call cap, cadence gate exempted, pressure trigger exempted, request trigger
exempted, cached-prefix tail gate bypassed, `skip_file_reads` auto-mode inverted, model escalation
on prompt overflow, and serialised cache-write ordering. That is two components sharing one config
struct, and `per_output: false` plus the *"leaves the component with nothing to do"* error is the
seam showing.

When that lands, a transporting `strategy` under `extract_llm_sweep` should become a **config-time
error** rather than a runtime narrowing — strictly better feedback. The narrowing here then
becomes defence-in-depth, following the precedent set in `2d6902d`, which kept
`dropOrphanedToolResults` "as a defensive net rather than the primary mechanism."

**Whether the sweep's `rewrite` should default to `false`** (verified deletion-only containment
rather than the `derivationRatio` backstop). It is the adjacent question — proof *strength* rather
than text *transport* — and it would reduce sweep savings by an amount only measurement can
establish, so it is deliberately unchanged here.

### DIFF
--- a/components/offload/extract_llm.go
+++ b/components/offload/extract_llm.go
@@ -226,6 +226,27 @@ const (
 	unknownModelInputLimit = 32768
 )
 
+// nonTransportingStrategies are the strategies that never hand tool-output text back THROUGH the
+// model: `code` has it write a Starlark program executed against the real text, and
+// `deterministic` makes no model call at all. `single` (returns the selected JSON) and `rlm`
+// (returns selected content per chunk) are the transporting ones, and are what a sweep excludes.
+//
+// `auto` is deliberately absent rather than filtered: it is a strategy ORDER, and
+// extract.intersectAllowed narrows the order it produces, so an `auto` configuration still resolves
+// — to code, then deterministic — instead of being emptied.
+var nonTransportingStrategies = []string{"code", "deterministic"}
+
+// transportsText reports whether a configured strategy would have asked the model to reproduce
+// content, so narrowing it can be counted rather than applied silently. `auto` counts, because its
+// first pick on a large body — the sweep case — is `rlm`.
+func transportsText(strategy string) bool {
+	switch strategy {
+	case "single", "rlm", "auto":
+		return true
+	}
+	return false
+}
+
 // staticWindows is the last-resort model→window table modelinfo already maintains. Held as
 // a package var because DefaultStatic allocates.
 var staticWindows = modelinfo.DefaultStatic()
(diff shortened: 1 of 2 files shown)


amiddavid:
Closing — superseded before merge, and by a better shape.

This restricted the sweep's strategies to `code` and `deterministic`, which stops the model **returning content** but leaves `code` **rewriting** it: the model still writes a Starlark program that filters lines and trims within them. Re-reading the post-trim contract in `internal/extract/bulk.go` on `feat/coref-compaction` shows the intended shape is stronger:

> `"verdict"` — keep (still needed, or you are unsure — this is the default) or drop (its information is spent; a short descriptor of its shape will remain in its place). A verdict of `"drop"` REQUIRES needed_by `"none"`.

So once trim is gone the operation is **binary** — keep verbatim or drop entirely — with no column stripping and no elision markers. That removes the transporting *operation* rather than just the transporting *strategies*, which is the point of the trim measurement (accepted once against eight rejected as invented).

Two things follow, and both are why this PR should not land as a half-step:

- The sweep's `rewrite` question is **moot**, not open. `rewrite` governs how a rewritten result is validated; if nothing is rewritten there is nothing to validate. I had been asking for a decision on a knob that should not exist on that path.
- The `extract_llm` / `extract_llm_sweep` split stops being config hygiene and becomes the actual design: the sweep is not a compactor with different thresholds, it is a different kind of component. `strategy`, `rewrite`, `aggressiveness` and `max_chars` all stop applying to it.

Replaced by a per-output spent-ness adjudicator, in a new PR. Nothing here is lost that is worth keeping: the strategy classification and its tests are subsumed, since an adjudicator never selects a compaction strategy at all.

PR #126, opened 2026-08-29
feat: bulk-create keep-alive strategies from KV-cache suggestions (strategy campaigns)

itay-nakash:
## Summary

- Bulk-create/activate keep-alive strategies from a batch of `GET /api/kvcache/suggest`
  cells in one manager action, restricted to arms with a real enforcement path today.
- Freeze what each cell predicted at creation time; read the real, live saving back
  later without ever storing it — the same convention `dash/cachehistory.go` already
  follows for "computed on read, never written."
- New dashboard tab (`Campaigns`): fetch live suggestions or upload a JSON file,
  preview, create, and drill from an aggregate per-tenant overview into any tenant's
  hour-by-hour historical/predicted-vs-real breakdown.

## How it works

- **Enforcement is restricted to arms with a real runtime path.** Most suggest arms
  (`historical-probability`, `sticky-session-1h`, `observed-policy`, ...) are
  simulation-only and are recorded with a reason, never activated — see
  `campaignArmFor` in `proxy/campaign.go`.
- **`tenant.Strategy` gains `HeadTTL1h`/`HeadTTLMinTokens`** so the 1h-tier arms
  (`keepalive-1h`, `keepalive-1h-once`) can be enforced too. The write-tier decision is
  read on the request path (`applyMode`), not inside `keeper.record` where a strategy's
  ping schedule resolves, so `resolveHeadTTL` is a second resolution beside
  `applyStrategy`'s existing one. Additionally gated **per tenant** by which model that
  tenant's traffic runs on: this deployment's own measurement (`apply/headttl.go`) shows
  Bedrock granting the 1h tier only to `aws/claude-haiku-4-5` and silently downgrading
  everything else — activating it blindly would pay the tier's write premium for zero
  benefit on the traffic that dominates.
- **Cells are coalesced by (arm, exact hour set)**, not just by arm: a strategy's
  `Target` and `Windows` both apply to the whole strategy, so merging two tenants
  recommended for different hours would activate an hour for one of them it was never
  recommended for. Windows are tiled in UTC, matching exactly the hour suggest computed
  each cell's saving from.
- **A cell with no tenant id is excluded outright** — single-tenant traffic or a
  pre-tenancy row, never a safe campaign target in a hosted deployment.
- **Real saving reuses `StrategyLedger`'s two-disjoint-query shape**
  (`dash/campaignsavings.go`): a ping's cost carries its strategy id, the real request it
  later rescues does not, so cost and saving can never come from one query. Generalized
  to a per-(tenant, hour) grid, time-bounded to each campaign's own activation, with a
  new covering index.
- **The create flow always submits the exact payload its preview showed** (`source:
  "upload"`), even when that payload came from "Fetch live suggestions" — re-fetching at
  commit time would let traffic between preview and commit change the numbers a manager
  just reviewed.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean on all touched files
- [x] `go test ./...` — full suite green
- [x] Each commit independently builds and passes its own package tests (spot-checked
      via detached-HEAD checkout), so the history bisects cleanly
- [x] New tests cover: the enforceable-arm mapping stays within `validStrategyBounds`,
      empty-tenant-id exclusion, non-activatable arms recorded with a reason, the 1h-tier
      per-tenant model gate, window tiling has no gap/overlap and never wraps hour 23
      into hour 0, coalescing only merges tenants sharing the *exact* same hour set, an
      end-to-end campaign create through the real HTTP route (fixture suggest payload →
      correct strategies → correct cells), and the per-tenant real/predicted aggregate
      wiring
- [x] Manually verified in a real browser: signed in through the actual sign-in form,
      uploaded a suggest fixture, created a campaign, opened the aggregate drawer,
      drilled into a tenant, navigated back — caught and fixed one real bug along the
      way (`omitempty` on the aggregate totals made a genuine `$0` render as `—`)


itay-nakash:
## Multi-agent review + fixes

Ran a comprehensive review of this PR: 8 independent dimensions (engine logic,
request-path/hot-path risk, persistence & migrations, dash SQL/schema, security &
authz, frontend JS, completeness-vs-the-PR's-own-claims, test-coverage audit), each
finding adversarially verified before acting on it (39 agents total, several of which
wrote and ran real reproduction tests against the code rather than reasoning from
reading alone). 31 findings raised, 28 confirmed or plausible, 3 refuted with evidence
(one via an actual scope-check experiment, one via a real SQL trace, one by compiling
and running the exact input in question).

Fixed everything that was a real bug or a meaningful gap, re-verified with new
automated tests plus a live pass in a real browser (signed in through the actual
sign-in form, not a shortcut). Commits `255a535`, `2e3d9d3`, `3c8881d`, `fc09a8a`.

### Fixed

**The 1h-tier arms didn't actually work** (`fc09a8a`) — `keepalive-1h`/`keepalive-1h-once`
campaign strategies were created with `HeadTTLMinTokens` left at 0. The apply layer's
own gate is `HeadTTL1h && HeadTTLMinTokens > 0`, so every 1h-tier campaign strategy
silently never promoted the head to the 1-hour tier while still paying for the 1-hour
ping schedule — full cost, zero benefit, on exactly the arms this feature exists to
activate. Now carries a real token floor (50000, mirroring
`config.DefaultHeadTTLMinTokens`), and `validStrategyBounds` refuses the zero-floor
combination outright for any future caller too. Added an end-to-end test that seeds
*real* recorded traffic under both an honored and a downgraded model and exercises the
actual DB-backed gate (`tenantHonorsHeadTTL1h`) — every previous test of that gate
injected a fake closure instead.

**An out-of-range hour created a live-but-permanently-dead strategy** (`fc09a8a`) — the
create route never validated a coalesced strategy's windows/target/bounds before
persisting it (unlike the sibling manual-strategy route, which always does), and the
hour-tiling function had no bounds check at all. An uploaded, hand-edited suggest
payload with `hour_utc: 24` produced a strategy that was created successfully (201,
`activatable: true`) but could never match anything, ever — pinging and saving nothing
forever with no error anywhere. Now bounds-checked up front, plus defense-in-depth
validation before every strategy create.

**Cross-tenant data leak in the per-tenant drill-down** (`fc09a8a`) — two tenants
sharing a coalesced strategy (an ordinary outcome of this feature's own coalescing) both
appear in the underlying real-savings query's un-tenant-scoped cost rows; the drill-down
handler collapsed them into a map keyed on hour alone, so whichever tenant's row landed
last silently overwrote the other's. A manager viewing tenant A's numbers could see
tenant B's ping cost and credited saving. Now filtered to the tenant actually being
viewed.

**Predicted vs. real compared two different populations** (`fc09a8a`) — the predicted
total summed every cell's frozen number regardless of whether it was ever activated,
while real can only ever cover activated cells — so a campaign dominated by
simulation-only arms (the common case, by this feature's own design) showed an inflated
"Predicted" tile next to a "Real" tile computed over a much smaller set, with nothing
saying so. Predicted now only sums cells that actually got a strategy.

**Audit-write loop stopped on the first failure** (`fc09a8a`) — skipping every
already-live, already-cell-committed strategy after it, not just the one that failed.
Now attempts every strategy regardless and reports the aggregate outcome.

**Two clocks that were supposed to be one** (`fc09a8a`) — the write-tier resolution
(`resolveHeadTTL`, called from the request path) used `time.Now()` directly while the
ping-schedule resolution (`applyStrategy`, called from `keeper.record`) uses the
keeper's own injectable clock — both sides' doc comments assert they resolve the same
matching strategy "so the two can never disagree," which only holds if they see the
same instant. They happened to agree only because the keeper's clock defaults to
`time.Now` in production. Added a nil-safe `clockNow()` so this is provably true
instead of coincidentally true.

**`$0` rendering as absent** (`fc09a8a` + `3c8881d`) — the two normalized real-saving
fields (`$/1k requests`, `$/active day`) were plain `omitempty float64`s, so a
genuinely-zero rate was indistinguishable on the wire from "never computed," and the UI
rendered it as `—`. Changed to `*float64`, non-nil exactly when the denominator was
actually nonzero; UI updated to check `!= null` instead of truthiness.

**`CampaignRealSavings`' two query halves weren't independent** (`255a535`) — the early
return required *both* `strategyIDs` and `tenantIDs` to be non-empty, even though each
half only needs its own list. A campaign whose every cell resolved to a non-activatable
arm has no strategy ids at all, and would silently skip the saving-half query too, even
with real tenant traffic to report.

**Duplicate `(tenant, hour)` cells surfaced a raw SQL error** (`2e3d9d3`) — an uploaded,
hand-edited payload with two cells at the same tenant+hour collides on
`campaign_cells`' own primary key; the rollback was already correct (verified: zero
orphan strategies), but the caller saw a raw SQLite constraint message after strategies
had already been created and torn back down. Now rejected up front with a clear error
naming the duplicate.

**`skip_reason` stored `''` instead of `NULL`** (`2e3d9d3`) — inconsistent with the
sibling `strategy_id` column, which correctly uses `NULL` for "unset." Fixed, and
checked at the raw SQL level (the Go API normalizes both back to `""`, so this needed a
direct query to actually catch).

**A redundant index** (`2e3d9d3`) — `campaign_cells`' own primary key already creates
SQLite's automatic covering index for the one query that index was meant to serve;
measured via `EXPLAIN QUERY PLAN`, the planner never chose it. Dropped from the (still
unmerged, so safe to edit directly) migration.

**Frontend: double-submit, stale drawer renders, a bad-upload crash** (`3c8881d`) — "Create
campaign" had no re-entrancy guard, so a double-click created two full, independent
sets of live strategies — verified with an actual rapid double-click in a real browser
(confirmed exactly one campaign resulted after the fix). The drawer's async renders had
no cancellation guard against a slow, now-stale fetch overwriting whatever's currently
on screen after navigating elsewhere in the same drawer. Uploading literal JSON `null`
(or a bare number/array) threw an uncaught `TypeError` with no visible error.

### Not fixed (flagged, not acted on)

The audit surfaced ~14 test-coverage-only suggestions with no actual bug behind them —
things like "no test creates two campaigns to check `ListCampaigns` ordering," "the
live-fetch source path has no end-to-end test," "an archived campaign's routes aren't
re-tested post-archive." Closed the highest-value ones (the real 1h-tier gate, the
duplicate-cell path, the out-of-range-hour path), left the rest as known gaps rather
than writing tests for their own sake.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` clean, `go test ./...` green — the whole
  repo, every package, not just the touched ones.
- Every new/changed test passes, including a genuine end-to-end reproduction of each
  fixed bug (not just the fixed code path in isolation).
- Live browser pass after the fixes: the null-upload guard, a rapid double-click on
  Create (exactly one campaign resulted, not two), and the `$0`-vs-`—` rendering all
  confirmed visually, zero console errors.

itay-nakash:
## Second review round: fresh-eyes multi-agent pass + fixes

Ran a second, independent multi-agent review of this PR — 28 agents across 8
dimensions (regression-in-fixes, all-arms-real-e2e, concurrency-and-races,
scale-and-performance, security-fresh-eyes, frontend-fresh-eyes, docs-pr-consistency,
meta-test-quality), each finding adversarially verified by an agent that tried to
REFUTE it against the actual current code (several with real reproduction tests, run
and reverted). 20 findings raised, 19 confirmed or plausible, 1 refuted (a
cross-campaign strategy-collision concern that turned out to be this codebase's own
documented, pre-existing precedence design, unrelated to this feature).

Two of the confirmed findings were incomplete fixes from round 1 — closing them
properly this time. The rest are new: one critical correctness bug in an arm nobody
had exercised end-to-end yet, one real concurrency gap, one activation-safety gap, and
several smaller correctness/security/test-coverage issues. Commits `5a0f71a`,
`9d78bcc`, `e79c72c`.

### Fixed

**`stop-reason-gated` pinged unconditionally, identically to `keepalive-5m`**
(`9d78bcc`, `5a0f71a`) — `campaignArm` had no field for `PredictorThreshold` at all,
so every campaign-created stop-reason-gated strategy was persisted with threshold
0.0. `predictorFor`'s probability function only ever returns 0 or 1, so
`p(stopReason) >= 0.0` is true either way — the predictor gate was a complete no-op.
The arm silently behaved exactly like the plain 5-minute arm while being labeled and
billed as the predictor-gated one, for however long it went unnoticed. Now carries a
real threshold (0.5, matching this codebase's own tests for this exact predictor), and
`validPredictorRef` refuses a named predictor paired with a non-positive threshold
outright, so this class of bug can't recur silently for a future predictor either.
Added an end-to-end test driving `keepalive-5m-once`, `stop-reason-gated`, and
`keepalive-1h-once` through the real create route with an assertion on every field of
the resulting strategy — previously only `keepalive-5m` and `keepalive-1h` got that
depth, which is exactly why this shipped unnoticed.

**Predicted vs. real still compared two different populations, on the REAL side this
time** (`9d78bcc`) — round 1 scoped the *predicted* total to activated cells only; the
*real* side was never narrowed to match. A tenant with zero activated cells in a
campaign (every cell simulation-only, or model-gated-out) could still show a nonzero
real-saving figure — in both the aggregate overview and the per-tenant drill-down — if
it happened to have unrelated real keep-alive credit from anywhere else. Now the real
side is scoped to exactly the tenants that received a strategy from *this* campaign,
matching the predicted side.

**A thin cell could still become a real, live strategy** (`9d78bcc`) — `suggest`'s own
`insufficient_data` flag (as few as 1-4 requests can "win" an arm by chance) was never
checked before activation. A manager clicking straight through "fetch → create" with
no hand-editing could get a real ping schedule created off a sample the suggest engine
itself says not to trust. Baseline arms stay exempt — doing nothing carries no risk
regardless of sample size.

**A rolled-back strategy could keep matching live traffic** (`9d78bcc`) — a
campaign-create failure after some strategies were already created deleted them from
the registry but never reloaded the keeper's in-memory list, and a newly created
strategy is matchable to live traffic (pinging, toggling KeepAlive/HeadTTL1h) the
instant it exists. A concurrent, unrelated strategy mutation reloading the keeper in
that narrow window could pick up the about-to-be-deleted strategy and keep it live
until some later reload happened to evict it. Now reloads once after every rollback;
delete failures are logged instead of silently swallowed.

**A duplicate-cell upload created strategies before being rejected** (`9d78bcc`) —
round 1's duplicate-`(tenant, hour)` guard lives inside `tenant.CreateCampaign`, which
this package only calls *after* already creating a live strategy for every coalesced
group. A duplicate-cell upload still caused real strategies to be created and then
rolled back, rather than never created at all. Checked up front now, before any
strategy exists.

**An omitted or fully-misspelled weekday list silently meant "every day"** (`9d78bcc`)
— `tenant.Window` treats an empty `Days` as no restriction at all, so a hand-edited
upload missing `weekdays_included` (or with every name misspelled) produced a live
strategy firing on Friday/Saturday too — traffic the frozen predicted numbers were
never backtested against (this suggester only ever reads Sunday-Thursday). Falls back
to that canonical week now instead of silently expanding scope.

**Internal errors leaked to the client** (`9d78bcc`) — three DB-backed failure
branches (the live suggest call, strategy create, campaign create) concatenated raw
driver/runtime error text into the HTTP response. Logged server-side now; the client
gets a fixed message, matching the hand-authored validation errors that were already
safe.

**Frontend: three more unguarded overlapping-fetch races** (`e79c72c`) — round 1's
drawer generation-token fix only bumped the token on drawer *open*, not on in-drawer
navigation, so two rapid clicks inside the same open drawer (two tenant rows, or an
overview/drill-down bounce) still shared one token and remained exactly as vulnerable
to a slow-response-overwrites-a-newer-one race as before the fix. Now bumped at the
start of every render, not just at open. Two more of the same shape had no guard at
all: the live-fetch/file-upload preview (`camp.previewGen`) and the campaigns list
refresh (`camp.listGen`).

**A test whose outcome depended on the wall-clock hour it happened to run at**
(`9d78bcc`) — `TestCtlGetCampaignTenantDoesNotLeakAnotherTenantsRealSavings`'s
recorded events defaulted to `time.Now()`, but the leak it guards only lands under the
cell's own hardcoded hour (9). Reverting the fix it guards still passed the test
unless the suite happened to run during UTC hour 9 — roughly 1 run in 24. Events are
now pinned to that hour explicitly.

### Not fixed (flagged, not acted on)

Two test-coverage-only findings with no bug behind them: the round-1 audit-write-
continues-past-failure fix and the `clockNow()`/`applyMode` two-clocks invariant both
have no dedicated regression test, but neither has an available fault-injection point
in the current test fixtures to exercise them deterministically — writing one would
mean adding new test seams for their own sake rather than testing real behavior. Left
as known gaps rather than force-fitting a brittle test. A third (a redundant,
non-load-bearing first assertion in `TestCreateCampaignRejectsDuplicateTenantHourCells`)
is genuinely inert but the test as a whole already correctly catches its regression via
its second assertion — no functional change needed.

A `frontend-fresh-eyes` finding on `refreshCampaignsList`'s missing generation guard
came back `PLAUSIBLE` rather than `CONFIRMED` (the specific failure scenario as
narrated had its causality backwards — archive already awaits its own DELETE before
refreshing), but the underlying gap was real and easily reachable a different way, so
it's fixed anyway (`camp.listGen`, see above) even without a reproduced failing test.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` clean, `go test ./...` green across the
  whole repo (`-count=1`, not cached), including every fix's own new/updated test.
- Every fix in this round was verified two ways before being trusted: read the exact
  current code and traced the mechanism, then (for the ones with a test) confirmed the
  new test fails against the pre-fix code and passes against the fix.
- Live browser pass after the fixes: signed in through the actual sign-in form,
  uploaded a suggest file, created a campaign, opened its drawer, and rapidly bounced
  between the overview and a tenant drill-down (back → tenant → back → tenant again) —
  correct final state each time, zero console errors.

OsherElhadad:
Reviewed on current head (`e79c72c`), which already includes a second round of fixes beyond the 4 commits (`255a535`/`2e3d9d3`/`3c8881d`/`fc09a8a`) referenced in the existing 39-agent review comment — `5a0f71a`/`9d78bcc`/`e79c72c` fixed a stop-reason-gated no-op threshold bug, a real-side population-scoping gap, a rollback-doesn't-reload-keeper race, weekday-fallback, error leakage, and 3 frontend races. `go build`/`go vet`/`gofmt`/`go test ./... -count=1`: all clean.

## Verified: all 6 originally-reported fixes hold, not superficially

Confirmed by reading current code and exercising it live against a hosted instance:
- `HeadTTLMinTokens` — `campaign.go:57`, `campaignDefaultHeadTTLMinTokens = 50000`, matches `config.DefaultHeadTTLMinTokens` exactly.
- Hour bounds — checked first in `resolveCampaignCell`; live-tested `hour_utc: -1` and `hour_utc: 25` → campaign recorded, zero strategies created, visible `skip_reason`.
- Cross-tenant drill-down leak — `campaignRealSavings` now filters by `tenantID` before building the per-hour map; confirmed live in the browser.
- Predicted-vs-real population — predicted only sums cells with a `StrategyID`; the real side is separately scoped to `tenantsWithStrategy`.
- Audit-write loop — no early `break`/`return`; always attempts every strategy.
- Two clocks — `resolveHeadTTL` and `keeper.record`'s `applyStrategy` call both resolve off the same `k.now()`.

## Adversarial JSON-upload pass (beyond the existing review) — no new vulnerability found

Tested directly against a live instance: negative/out-of-range `hour_utc` (handled, zero strategies), a typo'd arm name (handled, `activatable:false` with a named reason, nothing hidden), a 3,000-cell payload (rejected in 16ms at the 256KB body-size limit, no DoS surface), type confusion and unknown fields (clean 400s with informative, non-leaky messages), duplicate `(tenant,hour)` cells (rejected up front, zero strategies — the round-2 fix holding correctly, better than round 1's create-then-rollback). One item in the original brief doesn't apply as framed: a manager naming an arbitrary tenant in an uploaded cell isn't a scope bypass, because this app has exactly two roles and a manager already has global scope for campaigns *and* the existing manual-strategy form — this PR doesn't introduce a new privilege, it just makes "arbitrary tenant in a file" more discoverable than the old one-at-a-time form. Worth a UI copy addition (see suggestions below), not a code fix.

## Campaigns tab: audited against the dashboard's own tab-mounting convention and "the four rules" — zero deviations

Checked `dash/ui/campaigns.js` against every point of the pattern `kvcache.js`/`tools.js` establish: tab button insertion, own `<section>`, registration in the shared `loaders` table, `UNFILTERED_VIEWS` membership, and using only verified `app.js` exports (`el, clear, api, ctl, usd, num, when, tile, tileGroup, openDrawer, emptyState, errorState, loadingState` — all confirmed present in `app.js`). The one apparent gap (no self-appended stylesheet) isn't a deviation: the file introduces no new visual components, reusing existing `style.css` classes throughout — adding a pointless stylesheet link would be the actual violation.

Checked against `kvcache.js`'s "four rules": (1) no arithmetic in JS — every `[+\-*/]` hit is an integer UI counter, every dollar value is a direct `usd()` wrap of a server-computed field; (2) N/A, no idle-time concept in this tab; (3) zero-vs-absence — `real_saved_usd_per_1k_requests`/`per_active_day` are checked `!= null` against server-side `*float64` pointers, and this was independently confirmed live (a zero-activity tenant tile showed literal `$0`, an unmeasured hour showed a `"no data yet"` pill — two different states rendered correctly); (4) negative-not-clamped — styled `bad-text`/`good-text` by sign, going through the shared `usd()` helper that already implements this rule once for the whole dashboard. This is the most conservative of the three appended tabs — no new CSS, no new helpers, pure composition of existing primitives.

## Real browser pass (Playwright) — clean

Full walk against a live instance (register → verify → sign in via the real cookie flow): Campaigns tab lists existing campaigns correctly; "fetch live suggestions" on zero traffic shows a correct empty state, not a crash; a hand-crafted 4-cell/2-tenant upload (mixing a normal cell, a thin-data cell, and an out-of-range hour) previewed correctly with a thin-data pill, and Create surfaced "2 activated, 2 recorded but not activated" grouped by skip reason — exactly the visible, grouped surfacing the design should have; a rapid double-click on Create produced exactly one campaign (the round-1 guard holds live); the Stats drawer and per-tenant drill-down both rendered correctly, including the zero-vs-absence distinction above; a 3x rapid tenant-drilldown/back-to-overview bounce left the UI in a stable, correct final state with zero console/page errors; malformed/null/bare-int uploads all produced a clean visible error, zero uncaught exceptions. The UX goal the user asked for (easy to fetch-or-upload, preview, create, and see why something didn't activate) is met.

## Oracle-gap measurement — blocked at the permission layer; here's what's honestly available instead

Direct and copy-based access to the live production DB (the only path to a real oracle-gap number on production traffic) was correctly denied by the tool-permission layer as insufficiently authorized for production tenant data — on two independent review agents, in two different framings, neither routed around. So there is **no honest production oracle-gap number for campaign-eligible arms in this review**, and — separately confirmed — **no repo doc has ever published one**: `docs/results/kv-ttl-predictor-*.md` explicitly say "not reported" for the closest analogous comparisons.

What's genuinely available:
- **A synthetic methodology demo** (n=40, one tenant, one model, reusing `kvcache/registry_test.go`'s own `TestOptimalIsALowerBoundOnEveryOtherArm` fixture, run once and deleted — `git status` clean): confirms the `Simulate`-vs-`NewOptimal` machinery correctly discriminates arms and that 1h-tier arms close roughly a third more of the gap than 5m-tier ones on rows engineered to need it. Not production data.
- **A real, non-privileged, non-oracle aggregate** already on disk (`/tmp/kv_ttl_arms_result.json`, 66,779 requests / 17 tenants): pooled, `stop-reason-gated` (campaign-eligible) saves **$111.29 vs fixed-5m (1.54%)**; `historical-probability-tenant-tuned` is **net negative (−$2.52, −0.035%)** pooled. This file has no `optimal` column, so it can't be combined with the synthetic number into one real oracle-gap figure — reporting them separately, as here, is the honest version of this measurement.

The historical-probability-tenant-tuned result is worth the author's direct attention regardless of the oracle question: a campaign-eligible arm that's net negative on real pooled traffic shouldn't ship into `campaignArmFor` without either a fix or an explicit exclusion.

## Suggested fixes, by importance

1. **(Product, the actually-lazy fix)** `dash/kvcachesim.go` already computes every registered arm — including `optimal` — via `Simulate`/`NewOptimal`. Wire that existing computation into the Campaign drawer's "Predicted" tile (or an adjacent one) so a manager sees "$X predicted, $Y oracle ceiling" without any offline script. No new pipeline needed, just surfacing data the code already produces.
2. **(Data)** Investigate why `historical-probability-tenant-tuned` is net negative pooled before it's relied on anywhere in a campaign context.
3. **(Process)** If the real oracle-gap number matters, it needs either explicit user sign-off for the live-DB read (not a relayed claim), or a `context-guru` deployment whose dashboard already has real traffic recorded, so the existing live comparison panel can be read instead of anything new being computed.
4. **(Minor, cosmetic)** Add a line to the Campaigns UI copy noting that an uploaded file can create strategies for any tenant it names, regardless of who uploaded it — not a bug, but worth being explicit about given the upload path makes this more discoverable than the old manual form, with no confirmation step before Create.

OsherElhadad:
## Acted on the suggested-fixes list (items 1 and 4) — commit `63eedfe`

Reviewed on top of `e79c72c` (the second review round). Implemented the two concrete,
in-scope suggestions from the review above; the other two were a data question and a
process question, not code fixes — see below for why they're left as-is.

### Fixed

**Oracle ceiling now surfaces beside the Predicted tile** (suggestion 1) — `dash/kvcachesim.go`
already runs `optimal` through `Simulate` for the page-wide simulator, but the per-cell
suggester (`kvSuggestCell`) never did, so a campaign's own "Predicted" figure had no
headroom number beside it without an offline script. `kvSuggestCell` now also builds and
simulates `optimal` on exactly the same rows every other candidate is scored on — never
added to `Candidates` (`kvSuggestCandidates()` already excludes it there on purpose, since
it reads the true next-request time) — and freezes the saving as
`KVCacheSuggestion.OptimalSavingUSD`. Threaded through `resolveCampaignCell` into a new
`CampaignCell.OptimalSavingUSD` column (migration v11, still unmerged so edited directly,
the same way round 1 did for `skip_reason`/the redundant index) and summed in
`campaignTenantSummaries`/`ctlGetCampaign` over **exactly the same population**
`PredictedUSD` already is (`StrategyID != ""`), so the two are always apples-to-apples. The
Campaign drawer's aggregate overview gets one new tile, "Oracle ceiling (predicted)",
right beside Predicted.

**UI copy: uploaded/fetched payloads can target any tenant** (suggestion 4) — added a
sentence to the Campaigns tab's intro copy noting a fetched-or-uploaded payload can name,
and create strategies for, any tenant it lists, regardless of who is uploading. Not a new
privilege (a manager already had this scope through the existing manual-strategy form),
but worth being explicit about given there's no separate per-tenant confirmation step
before Create.

New tests: `dash/kvcachesuggest_test.go` asserts the ceiling is frozen, never below the
cell's own winner's saving, and never reachable as a candidate or a winner;
`proxy/campaign_test.go` and `tenant/campaign_test.go` cover the population-scoping and the
new column's round trip through a real migration.

### Not acted on (data/process, not code)

- **Suggestion 2** (`historical-probability-tenant-tuned` net negative pooled on real
  traffic) — confirmed this arm has no live enforcement path in `campaignArmFor` (it's
  simulation-only, recorded with a reason, never activated), so it isn't a campaign-safety
  issue for this PR specifically. Still worth the tuning follow-up the review flagged, just
  not something this PR's scope should fix.
- **Suggestion 3** (a real oracle-gap number against live production traffic) — needs
  either explicit sign-off for a live-DB read or a deployment with real recorded traffic to
  read from the existing comparison panel; neither is something to decide unilaterally here.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` clean, and `go test ./dash/... ./proxy/...
  ./tenant/...` green, `-count=1`.
- One pre-existing, timing-sensitive flake observed in
  `TestCtlGetCampaignAggregatesPredictedAndRealPerTenant`, but **only** under `go test ./...`
  full-repo load (never in isolation, never in the `proxy` package alone) — reproduced it
  identically on the unmodified `e79c72c` base in a separate worktree, so it predates this
  commit and isn't touched by it. Worth a follow-up look (likely WAL-visibility timing under
  load, the same class of issue `da51c890` already worked around once), but out of scope for
  this round's fix list.

PR #130, opened 2026-08-30
docs(spec): propose local distribution for a one-command Claude Code trial

amiddavid:
## What this is

A proposal for review, docs only — no code, no behavior change. It asks for approval on **how we distribute context-guru to someone who finds the repo and wants to try it on their own Claude Code sessions.**

Spec: `docs/superpowers/specs/2026-08-30-local-distribution-design.md`

## The problem

An evaluator today has to install Go 1.26, install a C toolchain, build with the right tags, run a binary, work out `ANTHROPIC_BASE_URL`, and trust that routing their coding agent through an unknown local proxy won't break it. Most stop at step one.

Step one turns out to be avoidable: **the C toolchain isn't required.** `CGO_ENABLED=0` produces a fully static binary for all four release targets — `go-tree-sitter` is the only cgo dependency and it's already behind `//go:build cg_skeleton`. Our quickstart tells every evaluator otherwise.

The value we want them to feel first is **KV-cache**, not the offloaders. Cache work requires being on the wire, so this is about making *the proxy* trivial to install — not about avoiding it.

Target:

```
/plugin marketplace add rossoctl/context-guru
/context-guru:install
```

## What the user does

Nothing is copied into their project, and the skill isn't something they have to place. `/plugin install` writes to `enabledPlugins` in `~/.claude/settings.json` (user scope is the default = "available across all projects"), so the plugin and its skills are installed **once per machine** and reachable from any repo.

```
1. /plugin marketplace add rossoctl/context-guru    # once per machine; prompts to trust the source
2. /plugin install context-guru@context-guru        # once per machine; may ask for /reload-plugins
3. /context-guru:install                            # once per repo (or once, with --global)
```

Step 3 is the whole reason a skill exists rather than a shell script. It detects the platform and installs the binary (one Bash approval), asks which scope to route, reads the target settings file and backs it up before adding **one** key — `env.ANTHROPIC_BASE_URL` (one file-edit approval), notices if the user already has a base URL set and asks rather than overwriting, then starts the proxy, verifies `/healthz` and prints the dashboard URL.

**Three to four approval prompts**, of which steps 1–2 happen once, ever. Afterwards `/context-guru:status` reads `/stats` and `/context-guru:uninstall` removes the key and stops the proxy.

| Thing | Scope | Frequency |
|---|---|---|
| Plugin + its skills | global — user settings | once per machine |
| Binary | global — `brew` / `~/.local/bin` | once per machine |
| **Routing (the `env` block)** | **per-project by default**, `--global` opt-in | **once per repo** |
| Proxy process | per-machine, started on demand | automatic |

Only the *routing decision* is per-repo, and deliberately so — it's the one with blast radius.

## What's proposed

| # | Piece | Why |
|---|---|---|
| 1 | Pure-Go release binaries + Homebrew tap | Removes the toolchain gate |
| 2 | `--idle-exit` self-terminating proxy | Nothing left running on the machine |
| 3 | `cache` preset (`cachesplit` alone) | The KV-cache pitch, fully lossless |
| 4 | Plugin: install skill + `scripts/` + `SessionStart` hook | One command, agent handles the judgment calls |
| 5 | Gateway conformance fixes | Routing must not break their agent |

## Three things worth your attention

**Our quickstart is wrong, and fixing it is the cheapest win here.** `docs/get-started/quickstart-proxy.md` instructs every evaluator to set `CGO_ENABLED=1` and install a C toolchain. That's only true for a `cg_skeleton` build. One-line change, removes the largest onboarding gate in the project, and **depends on nothing else in this proposal** — worth landing even if the rest is rejected.

**A rejected `cache_control` silently switches off prompt caching.** Per the gateway protocol reference, when the upstream rejects a `cache_control` marker Claude Code retries and disables the capability *for the rest of the conversation*. So a breakpoint-budget mistake isn't an error the user sees — it turns off the thing we're selling, and the demo reads as negative. This is the strongest argument for shipping `cache` (`cachesplit` only, no `cacheinject`) as the funnel default.

**A subscription needs no API key.** Setting `ANTHROPIC_BASE_URL` *without* a credential variable keeps the user's claude.ai login active — their Pro/Max limits and billing continue to apply. An evaluator can trial context-guru with no API key at all. That's the single most important adoption fact about this project and it appears nowhere in our docs.

## Design decisions I'd like challenged

**Default scope is project-local, not global.** Precedence is managed → `--settings` → `.claude/settings.local.json` → `.claude/settings.json` → `~/.claude/settings.json`. A global base URL pointing at `localhost` means a dead proxy breaks Claude Code everywhere, including repos the evaluator never meant to experiment in. That blast radius is the biggest risk in this proposal — bigger than the build. Local is safer, global is the better demo; I picked safer and am happy to be overruled.

**`SessionStart` hook rather than launchd/systemd.** Session-scoped lifetime leaves nothing on the machine and needs no privileged install. The spec documents the five details that make it correct — never `async: true` (reintroduces the race with the first request), idempotent because the event also fires on `clear`/`compact`/`resume`/`fork`, fixed port because the env block and the hook must agree, the clone-time trust prompt, and:

**The hook self-gates on `ANTHROPIC_BASE_URL`.** Because the plugin installs at user scope, its hooks run in *every* project — including ones the user never routed, where starting a proxy is pure waste. Settings `env` values are written into the process environment and hooks inherit it, so the hook exits immediately unless the variable names our port. No extra configuration, no second copy of the port to drift, and it degrades correctly: remove the env key by hand and the hook stops firing on its own.

**`--idle-exit` has a floor at `store.ttl_seconds`.** Exit wipes the in-memory Store, and a frozen decision dying mid-session is the 11.5× cache-write bug `FrozenLost` exists to catch. 24h is safely past the 10000s default; a 30-minute threshold would be actively destructive. Proposed as config validation, not a doc comment. Also: the keepalive inverts what "idle" means — pinging is what it does *while* there's no client traffic — so the watchdog must know about pending ping schedules or it kills the feature it ships.

**Skill + `scripts/`, not a script alone.** A skill doesn't reduce permission prompts, it concentrates them. The real reason for it is the part a script does badly: merging one key into a `~/.claude/settings.json` that already holds the user's theme, model, permission rules and possibly their own base URL. That's judgment work — read, back up, add one key, detect a conflicting base URL, verify, report. Our own settings files are the proof.

## Non-goals (stated explicitly in the spec)

- Not a replacement for the proxy — this is packaging *for* it, and cache work stays there.
- No offloaders in the funnel default — lossy-by-design is the wrong first impression for a cache pitch. Separate story in #129.
- No measurement-only mode — considered and rejected: without pings it diagnoses a problem it cannot fix.
- Not the DAM integration — DAM is harness-plural with its own gateway. Separate proposal.

## Open questions

All judgment calls — nothing here is an unknown that could invalidate the plan.

1. **Default scope: project-local or global?**
2. **Idle-exit default 24h**, and should the floor be `max(2 × store.ttl_seconds, 1h)`?
3. **Homebrew tap** as a new `rossoctl/homebrew-tap` repo — who owns release signing?
4. **Does the 30 MB artifact bother anyone?** Mostly tokenizer tables, the embedded dashboard UI and `modernc.org/sqlite`. A `-slim` build without the dashboard is possible, but it costs the demo its best surface.
5. **Ship `skeleton` in v1?** It's the only thing needing cgo, so a `-skeleton` variant means per-platform CI for one component that isn't in `codesmart` and isn't in the cache story. Proposal: omit from the first release, document the source build.

## Staging

- **Stage 1** — GoReleaser + tap + `cache` preset + fix the CGO claim in the quickstart. **Shippable alone, and most of the adoption win.**
- **Stage 2** — `--idle-exit` + the plugin.
- **Stage 3** — conformance items with tests.

---

<details>
<summary><b>Appendix: what was verified</b></summary>

Two assumptions the plan rests on were checked rather than trusted. Both came back the way the proposal already assumed, so neither changes a decision — they're recorded so you don't have to take the claims on faith, and the scripts are committed so you can re-run them. Full detail in the spec's Verification appendix.

**The binary needs no C toolchain** — `scripts/gate-a-purego.sh`, in a `golang:1.26` container so it needs no local Go:

| Check | Result |
|---|---|
| default tags, `CGO_ENABLED=0` | PASS |
| `-tags cg_skeleton`, `CGO_ENABLED=0` | fails, as intended — `build constraints exclude all Go files in …/go-sitter-forest/typescript`, the cgo-disabled signature |
| `GOOS`/`GOARCH` × {linux,darwin} × {amd64,arm64} | PASS, all four |
| artifact | 30.5 MB stripped, 38.3 MB unstripped, `ldd` → *"not a dynamic executable"* |

The first run of the `cg_skeleton` check failed on `proxy.golang.org` timeouts rather than on cgo, which would have been recorded as a confirmation it wasn't. The script now fetches the grammar modules first and reports `INCONCLUSIVE` if it can't.

**`env` blocks merge per key** — `scripts/gate-b-envmerge.sh`. Relocates the config dir via `CLAUDE_CONFIG_DIR` so it never touches a real `~/.claude`, and reads the result from a `SessionStart` hook rather than a model self-report. With `CG_USER`+`CG_BOTH` in user scope and `CG_BOTH`+`CG_PROJ` in project scope, the hook saw `CG_BOTH=project`, `CG_PROJ=project`, `CG_USER=user` — the higher-precedence file won only the key it sets, and the user-scope object survived.

Still unverified: the five gateway conformance items in piece 5.
</details>

Related: #129 (Claude Code plugin as a transport for the offloaders — different question, same plugin surface).

### DIFF
--- a/docs/superpowers/specs/2026-08-30-local-distribution-design.md
+++ b/docs/superpowers/specs/2026-08-30-local-distribution-design.md
@@ -45,17 +45,26 @@ build:
 
 So without `-tags cg_skeleton` there should be no cgo dependency at all.
 
-**Gate A — verify before anything else** (30 seconds, and it decides the shape of all of
-piece 1):
+**Gate A — measured, 2026-08-30. It passes.** Reproduce with `scripts/gate-a-purego.sh`
+(runs in a `golang:1.26` container, so no local Go needed):
 
-```sh
-CGO_ENABLED=0 go build -o /tmp/cg ./cmd/context-guru-proxy
-```
+| Check | Result |
+|---|---|
+| default tags, `CGO_ENABLED=0` | **PASS** — builds clean |
+| `-tags cg_skeleton`, `CGO_ENABLED=0` | **PASS (fails as intended)** — `build constraints exclude all Go files in …/go-sitter-forest/typescript`, the cgo-disabled signature. Confirms tree-sitter is the only C dependency |
+| `GOOS`/`GOARCH` × `linux,darwin` × `amd64,arm64` | **PASS, all four** |
+| artifact | **30.5 MB stripped** (`-ldflags="-s -w"`), 38.3 MB unstripped, `ldd` → *"not a dynamic executable"* |
 
-If it passes we get a **static single binary**: a plain `GOOS`/`GOARCH` matrix in one CI
-job, no C cross-toolchains, no zig, no libc coupling, ~20 lines of GoReleaser. If it
-fails, cross-compiling four platforms needs per-platform runners and the binaries carry
-libc requirements — still doable, several times the work.
+So we get a **fully static single binary**: a plain `GOOS`/`GOARCH` matrix in one CI job,
+no C cross-toolchains, no zig, no libc coupling, ~20 lines of GoReleaser. 30 MB is on the
+large side — the o200k_base tokenizer tables, the embedded dashboard UI and
+`modernc.org/sqlite` account for most of it — but it is a fine download for `brew` or
+`curl`, and it needs no runtime dependencies at all.
+
+**Consequence for the docs:** `docs/get-started/quickstart-proxy.md` currently instructs
+every evaluator to set `CGO_ENABLED=1` and install a C toolchain. That is only true for a
+`cg_skeleton` build. Fixing that paragraph is a one-line change that removes the largest
+onboarding gate we have, independent of everything else in this proposal.
 
 `skeleton` is the only casualty, and it costs nothing here: it is not in `codesmart` (the
 default) and not in the cache story at all. Ship pure-Go binaries for everyone; a
@@ -167,6 +176,27 @@ that blast radius is the main risk in this proposal — bigger than the build. H
 to project-local, `--global` as an explicit opt-in, and `--idle-exit` paired with the
 `SessionStart` hook so "is it running?" stops being the user's problem.
 
+**Gate B — measured, 2026-08-30: `env` blocks MERGE per key.** Reproduce with
+`scripts/gate-b-envmerge.sh` (relocates the config dir via `CLAUDE_CONFIG_DIR`, so it never
+touches a real `~/.claude`; a `SessionStart` hook dumps the environment it was handed).
+
+With `CG_USER`+`CG_BOTH` in user scope and `CG_BOTH`+`CG_PROJ` in project scope, the hook saw:
+
+```
+CG_BOTH=project      <- project won the key it sets
+CG_PROJ=project
+CG_USER=user         <- survived; the user object was NOT replaced
+```
+
+So a higher-precedence file wins only the keys it actually sets, and **a user-scope install
+survives a repo that ships its own `env` block.** That removes the failure mode this gate
+existed to catch.
+
+It does *not* change the default-scope recommendation. Per-key merge was one argument for
+project-local; the blast-radius argument above stands on its own and is the stronger one. It
+does mean `--global` is safe to offer without a per-repo caveat, and that `init` only has to
+reason about one conflict: an `ANTHROPIC_BASE_URL` the user already set themselves.
+
 **Why `SessionStart` and not a launchd/systemd service.** Session-scoped lifetime leaves
 nothing running on the machine and needs no privileged install. Mechanics check out:
 `SessionStart` supports `type: "command"`, hooks are awaited unless `async: true` is set, and
@@ -239,26 +269,42 @@ do not match a subscriber's bill.
 - **Not the DAM integration.** DAM is harness-plural with its own gateway; a Claude Code
   plugin covers one harness. Separate proposal.
 
+## Resolved before review
+
+Both gates that would have reshaped the plan were run rather than left open. Scripts are in
+this PR so a reviewer can re-run them.
+
+| Gate | Question | Answer |
+|---|---|---|
+| A | Does `CGO_ENABLED=0` build? | **Yes** — static 30.5 MB binary, all four targets, no C toolchain. `scripts/gate-a-purego.sh` |
+| B | Does `env` merge per key or get replaced wholesale? | **Merges per key** — a user-scope install survives a repo's own `env` block. `scripts/gate-b-envmerge.sh` |
+
+Neither answer changes a decision in this proposal; both remove a way it could have been
+wrong. Gate A is the load-bearing one — it is what makes piece 1 twenty lines of GoReleaser
+instead of a per-platform CI build.
+
 ## Open questions for reviewers
 
-1. **Gate A** — does `CGO_ENABLED=0` build? Everything in piece 1 scales off this answer.
-2. **Does an `env` block merge per-key across settings files, or is the whole object
-   replaced by the highest-precedence file?** The docs say *lists* merge and call `env` "an
-   ordinary key". If it is replaced wholesale, a user-scope install silently stops working in
-   any repo that ships its own `env` block — i.e. exactly the most-configured repos. Needs a
-   10-minute test.
-3. **Default scope: project-local or global?** Local is safer, global is the better demo. The
-   proposal picks local; happy to be overruled.
-4. **Idle-exit default: 24h?** And should the floor be `max(2 × store.ttl_seconds, 1h)`?
-5. **Homebrew tap: `rossoctl/homebrew-tap` as a new repo, and who owns release signing?**
+1. **Default scope: project-local or global?** Local is safer (a dead proxy costs one repo),
+   global is the better demo. The proposal picks local; happy to be overruled. Gate B means
+   global carries no per-repo caveat.
+2. **Idle-exit default: 24h?** And should the floor be `max(2 × store.ttl_seconds, 1h)`?
+3. **Homebrew tap: `rossoctl/homebrew-tap` as a new repo, and who owns release signing?**
+4. **Does the 30 MB artifact bother anyone?** It is mostly tokenizer tables, the embedded
+   dashboard UI and `modernc.org/sqlite`. A `-slim` build without the dashboard is possible if
+   it matters, but it costs the demo its best surface.
+5. **Ship `skeleton` at all in v1?** It is the only thing needing cgo, so a `-skeleton`
+   variant means per-platform CI for one component that is not in `codesmart` and not in the
+   cache story. Proposal: omit it from the first release and document the source build.
 
 ## Staging
 
-- **Stage 0** — Gate A + the `env` merge test. Half a day. Decides the rest.
+- ~~**Stage 0** — Gate A + the `env` merge test.~~ **Done** (see Resolved above).
 - **Stage 1** — GoReleaser + tap + `cache` preset + fix the CGO claim in the quickstart. The
   funnel already works here for anyone willing to run two commands by hand.
 - **Stage 2** — `--idle-exit` + the plugin (skill, scripts, `SessionStart` hook).
 - **Stage 3** — conformance items 1–5 with tests.
 
 Stage 1 is shippable alone and is most of the adoption win; stages 2–3 make it pleasant and
-safe.
+safe. With Gate A answered, the smallest useful change in the whole proposal is a one-line
+quickstart fix telling people they do not need a C compiler.
--- a/docs/superpowers/specs/2026-08-30-local-distribution-design.md
+++ b/docs/superpowers/specs/2026-08-30-local-distribution-design.md
@@ -8,6 +8,10 @@ right tags, run a binary, work out `ANTHROPIC_BASE_URL`, and trust that routing 
 coding agent through an unknown local proxy will not break it. Most evaluators stop at
 step one.
 
+Step one turns out to be avoidable. **The C toolchain is not required** — measured, not
+assumed (Gate A below): `CGO_ENABLED=0` produces a fully static binary for all four release
+targets, and our own quickstart is telling people otherwise.
+
 The value we want them to feel first is **KV-cache**, not the offloaders. That decides
 almost everything below, because cache work requires being on the wire — so this proposal
 is about making *the proxy* trivial to install, not about avoiding it.
@@ -19,31 +23,44 @@ Target: **two commands, no toolchain, no guide, reversible.**
 /context-guru:install
 ```
 
+## Findings up front
+
+Two questions could each have reshaped this proposal. Both were run before review rather
+than left open; the scripts ship alongside so a reviewer can re-run them.
+
+| Gate | Question | Answer | Reproduce |
+|---|---|---|---|
+| **A** | Does `CGO_ENABLED=0` build? | **Yes.** Static 30.5 MB binary, all four release targets, no C toolchain. `cg_skeleton` is the only cgo path. | `scripts/gate-a-purego.sh` |
+| **B** | Does an `env` block merge per key, or does the highest-precedence file replace the whole object? | **Merges per key.** A user-scope install survives a repo shipping its own `env` block. | `scripts/gate-b-envmerge.sh` |
+
+Neither answer changes a decision below; both remove a way this could have been wrong. Gate
+A is the load-bearing one — it turns piece 1 from a per-platform CI build into a plain
+`GOOS`/`GOARCH` matrix, and it means **the single smallest useful change in this whole
+proposal is a one-line quickstart fix telling people they do not need a C compiler.**
+
 ## What we ship
 
 Five pieces. Each is independently useful; the order is the order they unblock each other.
 
-| # | Piece | Why |
-|---|---|---|
-| 1 | Pure-Go release binaries + Homebrew tap | Removes the toolchain gate entirely |
-| 2 | `--idle-exit` self-terminating proxy | Nothing left running on the machine |
-| 3 | `cache` preset | The KV-cache-only pitch, fully lossless |
-| 4 | A plugin: install skill + `scripts/` + `SessionStart` hook | One command, and an agent that can make the judgment calls |
-| 5 | Gateway conformance fixes | Routing must not break their agent |
+| # | Piece | Why | Gated on |
+|---|---|---|---|
+| 1 | Pure-Go release binaries + Homebrew tap | Removes the toolchain gate entirely | Gate A ✅ |
+| 2 | `--idle-exit` self-terminating proxy | Nothing left running on the machine | — |
+| 3 | `cache` preset | The KV-cache-only pitch, fully lossless | — |
+| 4 | A plugin: install skill + `scripts/` + `SessionStart` hook | One command, and an agent that can make the judgment calls | Gate B ✅ |
+| 5 | Gateway conformance fixes | Routing must not break their agent | — |
 
 ### 1. Pure-Go binaries
 
 `docs/get-started/quickstart-proxy.md` tells every evaluator to set `CGO_ENABLED=1` and
-install a C toolchain. Reading the dependency tree, that looks unnecessary for the default
-build:
+install a C toolchain. The dependency tree says that is unnecessary for the default build,
+and Gate A confirms it:
 
 | Dependency | cgo? |
 |---|---|
 | `tiktoken-go/tokenizer` | **No.** `internal/tokens/tokens.go` says it outright: *"o200k_base is embedded in the binary (pure-Go, offline, no CGO)"* |
 | `modernc.org/sqlite` (dashboard) | **No** — pure Go by design |
-| `tree-sitter/go-tree-sitter` | Yes — but gated behind `//go:build cg_skeleton`, with `internal/treesitter/stub.go` for `!cg_skeleton` |
-
-So without `-tags cg_skeleton` there should be no cgo dependency at all.
+| `tree-sitter/go-tree-sitter` | Yes — the only one. Gated behind `//go:build cg_skeleton`, with `internal/treesitter/stub.go` for `!cg_skeleton` |
 
 **Gate A — measured, 2026-08-30. It passes.** Reproduce with `scripts/gate-a-purego.sh`
 (runs in a `golang:1.26` container, so no local Go needed):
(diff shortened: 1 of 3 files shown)


amiddavid:
## Both gates run — neither blocks, and Gate A is better news than expected

Pushed `9f7e221` with the results folded into the spec and the check scripts committed so you can re-run them.

### Gate A — static binary, no C toolchain: **yes**

`scripts/gate-a-purego.sh` (runs in a `golang:1.26` container, no local Go needed):

| Check | Result |
|---|---|
| default tags, `CGO_ENABLED=0` | **PASS** |
| `-tags cg_skeleton`, `CGO_ENABLED=0` | **fails as intended** — `build constraints exclude all Go files in …/go-sitter-forest/typescript` |
| `GOOS`/`GOARCH` × {linux,darwin} × {amd64,arm64} | **PASS, all four** |
| artifact | **30.5 MB stripped**, `ldd` → *"not a dynamic executable"* |

So: one CI job, plain `GOOS`/`GOARCH` matrix, no C cross-toolchains, no libc coupling, ~20 lines of GoReleaser. The `cg_skeleton` failure message is the cgo-disabled signature, which confirms tree-sitter is genuinely the only C dependency and the build tag is doing its job.

One process note: the first run of check 2 "failed" for the wrong reason — `proxy.golang.org` timeouts, not cgo. I'd have reported a confirmation that wasn't one. The script now fetches the grammar modules first and labels the outcome inconclusive if it can't, because a network error must not be mistakable for the error we're testing for.

**The most valuable thing here is not the release tooling.** `docs/get-started/quickstart-proxy.md` tells every evaluator to set `CGO_ENABLED=1` and install a C toolchain, and that's only true for a `cg_skeleton` build. Fixing that paragraph removes the largest onboarding gate in the project and depends on nothing else in this proposal. Worth landing on its own even if the rest is rejected.

### Gate B — `env` merges per key: **yes**

`scripts/gate-b-envmerge.sh` relocates the config dir via `CLAUDE_CONFIG_DIR` so it never touches a real `~/.claude`, and reads the answer from a `SessionStart` hook that dumps the environment it was handed. With `CG_USER`+`CG_BOTH` in user scope and `CG_BOTH`+`CG_PROJ` in project scope:

```
CG_BOTH=project      <- project won the key it sets
CG_PROJ=project
CG_USER=user         <- survived; the user object was NOT replaced
```

A higher-precedence file wins only the keys it actually sets, so a user-scope install survives a repo that ships its own `env` block. That kills the failure mode the gate existed to catch — an install silently dying in exactly the most-configured repos.

It does **not** change the default-scope recommendation. Per-key merge was one argument for project-local; the blast-radius argument (a dead proxy on a global base URL breaks Claude Code in every repo) stands on its own and is the stronger one. What it does change: `--global` needs no per-repo caveat, and `init` has one conflict to reason about rather than two — an `ANTHROPIC_BASE_URL` the user set themselves.

### Where that leaves the proposal

Open questions are still five, but they're now all judgment calls — default scope, idle-exit floor, tap ownership, the 30 MB artifact, whether to ship `skeleton` in v1 — rather than unknowns that could invalidate the plan. Stage 0 is done; Stage 1 is unblocked and shippable alone.

OsherElhadad:
Reviewed by re-running both committed gate scripts independently (not trusting the issue comment's reported results) and fact-checking the two claims the doc itself calls "most consequential" against the current, live docs rather than memory.

## CI is green, no action needed

`mkdocs build --strict` passes in ~5s. The new spec page (`docs/superpowers/specs/2026-08-30-local-distribution-design.md`) falls under the existing `exclude_docs: superpowers/` rule, so no nav change was needed here — unlike PR 129, which has the opposite problem.

## Gate A (pure-Go static binary) — re-run independently, all four checks confirmed

Docker was available but couldn't pull `golang:1.26` from this host (no network to Docker Hub), so the containerized script as written couldn't run as-is. Substituted the local Go 1.26.4 toolchain for every check in it — same Go version the script pins, so this is a full substitute, not a partial one:
- `CGO_ENABLED=0 go build ./cmd/context-guru-proxy` — succeeds; `ldd` confirms "not a dynamic executable."
- `CGO_ENABLED=0 go build -tags cg_skeleton ./cmd/context-guru-proxy` — fails exactly as claimed, with `build constraints exclude all Go files` against every `go-sitter-forest/<lang>` package, confirming tree-sitter is genuinely the only cgo dependency.
- Cross-compile matrix, `{linux,darwin}×{amd64,arm64}`, all `CGO_ENABLED=0` — all four succeed.
- Binary size — stripped 30.5 MB, unstripped 38.3 MB, matching the PR's claimed figures exactly.

## Gate B (settings.json per-key env merge) — re-run, confirmed exactly

Ran `gate-b-envmerge.sh` directly (`CLAUDE_CONFIG_DIR` relocated so it never touches the real `~/.claude`, as the script itself requires): exit 0, dump showed `CG_BOTH=project` / `CG_PROJ=project` / `CG_USER=user` — an exact match to the claimed result. A higher-precedence settings file only overrides the specific keys it sets; a user-scope install survives a repo shipping its own conflicting `env` block.

## The two "most consequential" claims — checked against current, live docs, not training-data memory

Both confirmed verbatim against `code.claude.com/docs/en/llm-gateway-protocol`, fetched live specifically because CLI behavior is exactly the kind of thing that can drift between versions:
- *"A rejected `cache_control` marker silently switches off prompt caching for the rest of the conversation."* Confirmed verbatim: *"When the upstream rejects the ... `cache_control` marker on one of those messages, Claude Code retries the request and disables the rejected capability for the rest of the conversation."* No error surfaces to the user — matches the doc's "the demo reads as negative" framing exactly.
- *"Setting `ANTHROPIC_BASE_URL` without a credential variable keeps the user's claude.ai subscription login active."* Confirmed and still true on current CLI-version docs (not assumed from memory): the gateway-protocol page independently describes the OAuth capability this depends on and what breaks if it's stripped. Worth noting as a bonus: the same page also independently corroborates two of the doc's other gateway-conformance claims (the SSE-buffering-stalls-the-client 300s abort behavior, and the attribution-block positional-strip behavior) that weren't specifically asked to be checked.

## Assessment

Every claim checked out — both gates, both "most consequential" facts, plus the incidental corroborations. No red flags. The doc's own framing of the quickstart's `CGO_ENABLED=1` instruction as "the cheapest win here, worth landing even if the rest is rejected" is itself now validated rather than just asserted — that one-line fix stands on its own regardless of how the rest of the proposal lands. The five "open questions" the doc lists (default scope, idle-exit floor, tap ownership, artifact size, whether to ship `skeleton` in v1) remain genuine judgment calls rather than unknowns that could invalidate the plan — nothing found in this review changes that.

amiddavid:
Implemented in #141 (all three stages, based on `main` with this branch merged in, so the spec shows up in its diff until this lands).

Decisions taken on the open questions, all reversible in review: routing scope **project-local** by default; `--idle-exit` **24h** with a `max(2 × store.ttl_seconds, 1h)` floor enforced at startup; **no Homebrew tap** (tarball + checksum, since the tap repo and signing are unowned); `skeleton` omitted from releases.

Two notes on the spec itself:

- The pure-Go claim re-checked independently on go 1.26.4 — all four targets static, 27.1–33.5 MB, and the binary serves `/healthz`. Confirmed.
- Conformance item 3 (the attribution block) is safe, but **not primarily for the reason the spec gives**. The `minSplitTokens` floor is only part of it — the block is not a split candidate at all, because it carries no volatile marker, and the third property is that untouched blocks are re-emitted from their original bytes rather than re-encoded. A test that relied on the floor alone passed with the floor removed; it now covers all three. So `CLAUDE_CODE_ATTRIBUTION_HEADER=0` is not needed.

Also split out of the implementation: #142 (the preset table was stale in every row) and #143 (the same drift in prose, unfixed).

OsherElhadad:
Reviewed against main `e88f5d8`, with both gate scripts actually executed, and audited against
#141 — which claims to implement this spec — and #129, the analysis that motivated it.

The spec is well written and the review round improved it (the self-gating `SessionStart` hook is
verified working on #141's shipped script: no `ANTHROPIC_BASE_URL` → exit 0 silent; wrong port →
exit 0 silent; matching port → starts, waits for `/healthz`; second run → no-op; proxy detached
`ppid=1`; no DB litter in the CWD. Two concurrent sessions cannot race the port, because the hook is
synchronous and probes `/healthz` first, so the loser exits at step 2).

The findings are about the **gates**, which are the executable part of this PR and the whole point of
calling something a gate.

## 1. Blocking — `gate-a-purego.sh` exits 0 with every single check FAILED. It cannot fail.

```
== 1. default tags, CGO_ENABLED=0 (want SUCCESS) ==
default tags / CGO=0                   exit=1   FAIL
cg_skeleton / CGO=0                    INCONCLUSIVE (could not fetch grammars)
== 3. cross-compile matrix ==
linux/amd64  exit=1 FAIL   linux/arm64  exit=1 FAIL
darwin/amd64 exit=1 FAIL   darwin/arm64 exit=1 FAIL
EXIT=0
```

`:19` sets `set -uo pipefail` **without `-e`**, and `result()` at `:30-38` only *prints* — it never
accumulates, there is no summary, and there is no `exit`. The script's status is whatever the last
`tee` returned, which is always 0. So a gate wired into CI would pass unconditionally.

**And the failures were spurious, which makes it worse.** Every module fetch died on
`dial tcp: lookup proxy.golang.org … i/o timeout` — the containerization ("so it needs no local Go
install") discards the host's warm module cache. The local equivalent passed in **1.588s** and
produced a statically linked binary. **So the spec's central claim is true, and its own gate reported
FAIL on it while exiting 0.** Both directions of wrong at once.

The author already learned this lesson and applied it to the wrong check: the appendix `:340-343`
records fixing check 2 so *"a network error cannot be mistaken for the cgo error"*. Checks 1 and 3
have no such guard. Check 2's old bug produced a false *confirmation*; check 1's produces a false
*refutation*.

Fix: `set -e` or an accumulator with a real `exit`, plus the same network-error guard on checks 1 and
3 that check 2 already has. Also `:20`'s unguarded `cd "$(dirname "$0")/.."`, no cleanup trap, and it
leaks a named Docker volume `cg-gomodcache`.

## 2. Blocking — gate A tests the cheap half, and the real risk of shipping pure-Go is not tested at all

- **It never runs the binary.** No `/healthz`, no startup — build-only, plus size and linkage.
- **It never covers the `unknown component "skeleton"` startup-exit path**, which is the actual
  production risk of a pure-Go build. `config/config.go:365-368` records that `coding` named
  `skeleton` until 2026-08 and *"failed to build with `unknown component "skeleton"` for every user
  who selected it"*. That class is guarded by `TestEveryPresetBuilds` — a **Go test, not this gate**.
- **Check 4 (`:80-89`) asserts nothing.** It prints size and `ldd` output with no PASS/FAIL. That is
  reporting, not a check.

Worth noting how cheap the missing half is: I confirmed independently that all **15 presets start and
answer `/healthz` on a `CGO_ENABLED=0` binary**. That loop is a few lines and it is the check that
would actually have caught the `skeleton` incident this spec's own reasoning cites.

Related and worth fixing here: `ci.yaml:15` runs the suite only with `CGO_ENABLED: "1"`, so
`TestEveryPresetBuilds` never executes in the configuration being shipped. `CGO_ENABLED=0 go test
./config/ -run TestEveryPresetBuilds` closes it.

## 3. Blocking — #141 cites this gate as its proof, and does not contain it

`scripts/gate-a-purego.sh` is named at `docs/setup.md:12`,
`docs/get-started/quickstart-proxy.md:23`, `.goreleaser.yaml:5`, `Makefile:13` and `install.sh:4` in
#141 — two of them user-facing docs that PR edits — but `git ls-files` in #141 finds no such file. It
is added *here*. So either #141 declares a dependency on #130 landing first, or it cites its own
workflow `purego` assert step, which does exist and works. As things stand the headline claim of #141
points at a script that is not in the tree.

## 4. The spec describes a different design than the one that shipped

#141's entire hook manifest is one `SessionStart` hook running `start-proxy.sh`, plus `settings.py`
writing `env.ANTHROPIC_BASE_URL` into the user's real `settings.json`. That matches this spec. What it
does **not** match is #129, the analysis cited as motivation, which analyses and recommends a
`PostToolUse` → `updatedToolOutput` offload hook plus an MCP expand server — a genuinely different
transport. So this spec quietly picked the option #129 never evaluated. That is defensible and
probably correct, but the spec should say so explicitly, because right now a reader who follows the
citation chain gets two incompatible designs and no statement of which won or why. (Raised on #129 as
well.)

Scope beyond the spec: #141 also ships a `cache` preset, `--idle-exit`, a `count_tokens` route and
gateway conformance tests. If the spec authorized those, point at where; if not, they are scope creep
that should be split out — I have recommended a split on #141 for independent reasons.

## 5. The trial's honesty problem — a first-run user may see nothing

The spec's goal is a one-command trial, and #141 makes `cache` (`{cachesplit}`) the preset that trial
installs. This repo's own measurements say that is close to zero in exactly the regime a trial user is
in: `docs/dashboard.md:204` puts it at **$0.0298 across 1,127 sessions / 11,361 requests**, because the
env snapshot is captured once per session and 1,105 of 1,127 first requests read zero from cache. It
is **exactly zero** for an agent outside a git repo, under the 1,024-token `minSplitTokens` floor, or
on a non-Anthropic backend (`config/config.go:379-380`). The `-34.1%` figure #141 quotes comes from
harnesses running tasks back-to-back inside the cache TTL — the opposite regime.

A spec that designs a trial should state what the trial will show, and either pick a preset whose
value is visible on a first cold session or tell the user plainly that the first session is the worst
case. Right now the design optimizes for *verifiability* ("read one line of `config.go`") at the cost
of showing a stranger a number near zero — and #141's implementation does not even hold up the
verifiability half, since the `cache` preset does inject a tool despite the docs saying it does not.

## 6. Security items the spec should cover

- **Teardown.** #141 implements "stop the proxy" as `pkill -f "context-guru-proxy.*${PORT}"`, which
  cannot match the proxy (the port is passed via `LISTEN_ADDR`, never on the cmdline) and instead
  matches the Claude Code session's own Bash shell — and the natural broadening of that pattern would
  match a host's production service. If the spec specifies pattern-based teardown, that is a
  spec-level defect, not just an implementation bug; specify PID-from-socket or a pidfile.
- **Artifact integrity.** The install path downloads a release tarball. `install.sh` does handle a
  `checksum_mismatch` outcome, but the release is unsigned and the workflow floats its actions on
  tags. For a "one command, no toolchain" install that runs a binary proxying all of a user's LLM
  traffic, the spec should state what integrity guarantee it is offering.
- **Credential handling.** #141's `settings.py` resets `~/.claude/settings.json` from mode 600 to 644
  on a file containing `ANTHROPIC_AUTH_TOKEN`, its timestamped backup clobbers itself within one
  second so the user's undo is destroyed, and uninstall does not restore a pre-existing
  `ANTHROPIC_BASE_URL` it overwrote. A spec whose install edits a credential file should state the
  required properties (mode preserved, backup unique, prior value restorable) so an implementation can
  be checked against them.

## 7. Does anything run the gates?

This is the question that decides whether they are gates or documentation. If no workflow invokes
`scripts/gate-*.sh`, then combined with finding 1 they are neither run nor capable of failing — and
the spec's acceptance criteria are unenforced prose. Please wire them into `ci.yaml` (cheap, since
gate A's local form takes ~1.6s once it stops containerizing away the module cache), or state
explicitly that they are one-off manual checks and record their results in the spec instead.

## Verdict

**Needs changes.** The spec's reasoning is sound and its central technical claim is true — I verified
the pure-Go build independently and every preset starts on it. But the two artifacts that make this
more than a design doc are a script that cannot fail and, apparently, nothing that runs it. Fix the
`set -e`/accumulator, add the network-error guard to checks 1 and 3, make gate A start the binary and
loop the presets, and either wire both gates into CI or stop calling them gates.

`bash -n` clean on both scripts; no `shellcheck` on this host, so those were hand-reviewed. The real
`~/.claude/settings.json` was never written to.

OsherElhadad:
Three additional findings after running the second gate and checking CI.

## 8. Blocking — `gate-b-envmerge.sh` makes a live model call, and its own header says it does not

`:12` claims *"no model call, no network."* `:64` runs:

```
claude -p … --model claude-haiku-4-5 --permission-mode bypassPermissions
```

I did not execute it for that reason. A gate documented as hermetic that in fact spends money and
requires credentials cannot run in CI, and the header is the thing a reader trusts when deciding
whether it is safe to run. Either fix the header or make the gate hermetic.

Read-only review of the rest:

- **Same defect as gate A: it cannot fail.** `:22` is `set -uo pipefail` with no `-e`. It does exit 2
  on INCONCLUSIVE, but all three verdict branches at `:85-95` fall through to **exit 0** — including
  `REPLACE`, which is the exact failure the gate exists to catch.
- **Its real-config safety is sound, and better than I expected.** `:24-28` builds a `mktemp -d` lab
  and `:64` relocates with `CLAUDE_CONFIG_DIR="$CFG"`. That is the correct mechanism: in this
  environment `CLAUDE_CONFIG_DIR` **wins over `HOME`**, so a temp-`HOME` approach would not have
  protected the real config and this does. Worth stating in the spec, since anyone writing a similar
  test will reach for `HOME` first.
- **It proves none of the four properties the install design actually needs.** It probes one platform
  fact — whether `env` blocks union per key across scopes — not the installer's write. Zero coverage
  of idempotency, unknown-key preservation, atomic write, or uninstall restore. The name promises a
  gate on the env merge; the body tests a precondition of it.

Tested directly against #141's `settings.py` instead: idempotency ✓, unknown-key preservation ✓,
atomic (no stray tmp) ✓, exact uninstall ✓ — but **mode `0600` → `0640` ✗, a symlinked settings.json
replaced by a regular file ✗, and a read-only `0444` file silently modified ✗.** The mode widening is
the one that matters: that file holds `ANTHROPIC_AUTH_TOKEN`.

## 9. Blocking — nothing runs either gate

```
$ grep -rn "gate-a\|gate-b\|scripts/gate" .github/
(no match)          # on main and in #141
```

The only two references in the entire repository are this spec's own appendix (`:330`, `:345`).
Combined with finding 1 and the item above — neither gate is capable of failing, and neither is
invoked — the acceptance criteria are unenforced prose. Wire them into `ci.yaml` (gate A's local form
runs in ~1.6s once it stops containerizing away the module cache; gate B needs to be made hermetic
first), or state plainly that they are one-off manual checks and record their results in the spec.

## 10. Blocking — the spec's stated distribution decision is the opposite of what shipped

`:116` — *"Decision: tap first… the primary path for macOS and Linux."*

#141 ships **no Homebrew tap**, and `install.sh:19-21` records why: the tap repository does not exist
and signing ownership is unresolved. So merging this spec as written publishes a decision that the
implementation already reversed, and the next reader will treat "tap first" as the plan of record.

Either update `:116` to the curl-installer-first reality with the tap as future work, or state the
tap as a blocked prerequisite and say what unblocks it. This is the clearest instance of the general
problem in finding 4: the spec and its implementation have diverged and neither document says so.

PR #138, opened 2026-08-31
fix(collapse): cut a head/tail window whenever the output is oversized — including single-line and huge-line payloads

amiddavid:
## Headline

**17 of 75 runs errored** in a live 128k-band arm. Every one of them was an upstream 400
`prompt is too long` on an outgoing body of 2.6–14.8 MB, and no component in the pipeline
touched those bodies. `collapse` is documented as the last resort for an oversized tool
output, so when it declines, nothing caps the output — and it declined this entire shape by
construction.

**The payoff here is avoided hard failures, not a savings line, and a reader primed by the
four-denominator convention should not go looking for one.** `collapse` is tail-gated and
freezes its decision, so a new cut happens only on the turn the output arrives — the cold
tail. That is exactly where a multi-megabyte `tool_result` lands and exactly where the 400
fires. A savings series aggregated over turns will barely move; what moves is the error
count. (Independently, a dashboard agent driving the live UI measured 113,975 → 1,824 bytes
on a novel 200 KB single-line blob where `main` reduced it by zero, so the cut is real and
visible in `/api/components` with no `dash/` changes.)

## The defect

`collapse` was line-oriented only:

```go
lines := strings.Split(content, "\n")
if len(lines) <= cl.headLines+cl.tailLines {
    rep.Gate("too_few_lines") // few long lines; head/tail wouldn't help
    continue
}
```

Defaults are `head_lines: 20`, `tail_lines: 20`, so **any output of 40 lines or fewer was
skipped**. For the single largest shape of tool output in practice the reasoning is inverted: a
database or HTTP API result serialised as JSON is very often **ONE line**, so a single-line
multi-megabyte payload fell through `collapse` entirely.

Review then found the same defect one level up: the fix's own gate was still in **line-count**
units, so an output with *more* than 40 lines took the line path however big those lines were.
Both are fixed here; see *The gate is in bytes* below.

### Nothing else caught it (measured, live 128k-band arm)

| component | acted | why not |
|---|---|---|
| `extract` | **0** | `no_obvious_noise` 16,891 — it only strips known noise patterns |
| `cmdfilter` | **0** | `no_filter_match` — only known command families |
| `toon` | 21 | `not_uniform_object_array` 233,501 |
| `dedup` | 774 | exact duplicates only |
| `summarize` | 2,349 | the only real compactor — and it protects `keep_last`, where a FRESH oversized output sits |
| `extract_llm` | — | declines by design: an output exceeding the compaction model's context is left verbatim (`over_model_context`) |
| `linecap` | **0** | `no_long_or_repeated_lines` — `neverTruncate` exempts the line; see below |

**`linecap` is not a partial mitigation for this shape — it is void for it.** On paper its
500-char per-line cap should rescue a huge line. It does not: `neverTruncate` exempts any line
matching `^\S+:\d+`, and a one-line JSON body (`{"items":[{"id":1,...`) matches that **by
accident**. Measured directly by the reviewer: `in=2097191B out=2097191B acted=False
gates=map[no_long_or_repeated_lines:1] neverTruncate(line)=true`. Earlier revisions of this PR
described `linecap` as capping huge kept lines "unless `neverTruncate` exempts them", which
implied a conditional mitigation; for the exact payload shape this PR exists to handle the
exemption always fires. That claim has been removed from the code comment, the commit message
and this body. Filed as #151.

### Consequence

- **17 of 75 benchmark runs errored** in that arm
- **17 upstream 400s**, all `"prompt is too long"`, in 3,347 requests
- outgoing bodies of **2.6 MB to 14.8 MB**

A note on those counts, because an earlier version of this body said 16. The capture digest for
the arm (`capfail-i21A.jsonl`) holds **19** records: 17 × `400 prompt is too long` and 2
transport-level failures (`IncompleteRead`, read timeout) that are not provider rejections. I
cannot reconstruct the filter that produced 16, so the number the record actually supports is
17, and that is what this body now uses.

### Correction: the evidence does NOT show the fix would have prevented those failures

Earlier revisions argued: *"a digest of one failing 14.8 MB body held only 4 messages, so a
single `tool_result` was multiple megabytes."* Review is right that this is a **message** count,
not a **line** count, and that nothing in it establishes those bodies had ≤ 40 lines — i.e.
nothing established the old `too_few_lines` gate was what let any of them through.

**The line-count histogram asked for cannot be produced, and not because the data is
inconvenient to reach — it was never recorded.** I went back to the experiment record on the
box that still holds it (`/tmp/cg-loca/capfail-i21A.jsonl`, read-only). The digest schema is:

```
status, body_bytes, rig_seq, n_messages, system_present, error,
digest[] = { i, role, types[], tool_use[], tool_result[] }
```

Per message it records **structure only** — index, role, block-type names, tool-use ids. No
content, not even a truncated head (the writer's `data[:600]` truncation applies to the *error*
string, not the request). No per-message or per-block size of any kind, so the largest
`tool_result` cannot even be identified, let alone measured in lines. The only size datum is
whole-request `body_bytes`, which includes JSON escaping, the system prompt and 94–149 tool
definitions, and cannot be converted to a line count of anything. `rig_seq` — the field added
so a record could be joined back to the request the shim handed the proxy — is `null` in all 19
records, and no raw copy of the bodies survives anywhere on the box (`proxy-i21A.log` and
`flap-i21A.jsonl` record bytes, never lines). Verified against the writer,
`/tmp/cg-loca/capture_hop.py`.

So: **the question "did those 17 failing bodies have ≤ 40 lines" cannot be answered from this
experiment record, in either direction.** What the record establishes is the class — 17
oversize-prompt 400s on 2.6–14.8 MB bodies with 4–57 messages — not that this specific gate was
the one that let them through. The case for the change therefore rests on the *class* being
real and on `collapse` being the only component contracted to catch it, plus the reproduced
mechanism in tests and the reviewer's live measurements — not on a demonstrated one-to-one with
those 17. Answering it properly needs a new capture that records newline counts per block;
that is worth doing before the next arm and is not something this PR can retrofit.

## The design

`collapse` picks its window before the depth gate: **by lines** when a line window actually
shrinks the output, **by runes** when it does not. Runes rather than bytes, because the result
is spliced back into the request body verbatim and a split multibyte rune would emit invalid
UTF-8 upstream.

### The gate is in BYTES, not lines, and not tokens

Having more lines than the window says nothing about their size. With a line-count gate, an
output with a multi-megabyte line inside the kept head or tail survived untouched — and *worse
than being declined*, because dropping the small filler lines in the middle counts as acting,
which stamps a marker, which then makes `linecap` decline `marker_or_kept_verbatim`. Measured:

```
huge line in head: in=4195111B  after collapse=4194922B (events=map[])
                   after linecap=4194922B (gates=map[marker_or_kept_verbatim:1])  tokens=1570220
```

**4 MB in, 189 bytes saved, 1.57 M tokens forwarded.** The line window is now taken only when
it at least **halves** the output; anything else falls through to the character window:

```go
if cl.headLines+cl.tailLines > 0 && len(lines) > cl.headLines+cl.tailLines {
    asm := /* the existing line closure, unchanged */
    if len(asm(""))*2 <= len(content) {
        return asm, false, true // the line window really cut it: byte-identical, as before
    }
    // it kept most of the bytes (a giant kept line) -> fall through to the rune window
}
```

**Bytes deliberately, not tokens.** `maxTokens` comes from `resolveBudget(..., c.CtxWindow)`,
`CtxWindow` can resolve differently mid-session (model swap, refreshed `modelinfo`), and
`repairLostFreeze` re-derives at depth. A token gate could therefore classify the same content
line-window on one turn and character-window on the next and emit **different bytes** — trading
a savings bug for a cache bug, in a component whose collapsed form is frozen and replayed
inside the cached prefix. The byte form depends only on the content and
`head_lines`/`tail_lines`, so it is replay-safe unconditionally. Both forms classify every
fixture identically, including all four in `TestCollapseLineWindowIsByteIdentical`, which is
what preserves byte-identity for everything the line window handles today.

### Where the character budget comes from — no new knob

It is the component's own token threshold expressed in characters:

- `max_tokens`, resolved through `resolveBudget` (so `max_frac` × the model window still wins),
  × **4** — the same chars/token ratio `internal/tokens.Count` itself falls back to when the
  BPE codec is unavailable, so the two places in the tree that relate bytes to tokens agree;
- **never seeded above the content**: `min(maxTokens*collapseCharsPerToken, len(r)-1)`. chars/4
  *overestimates* on dense JSON (~2.5 chars/token), and the bail-out was tested against that
  **un-measured** seed, so content genuinely oversized in tokens was still shorter in runes
  than its char budget and was declined outright as `too_few_lines_and_chars`. That hole was 55%
  wide, and widest in a shipped preset — `codesafe` sets `max_tokens: 3000`, so one-line JSON
  from **3,006 to 4,491 tokens was oversized and left entirely uncut**, with no test at 3000.
  `marker_no_win` already protects the downside of seeding low;
- split between head and tail in the **`head_lines`:`tail_lines` ratio**, so those knobs keep
  meaning what they say on both paths;
- then **measured and tightened**. On the 3 MB JSON fixture, unmeasured chars/4 lands at
  **3,056 tokens against a 2,000-token threshold**. The fit loop is bounded at 3 passes and only
  ever tightens;
- floored at **200 characters**, so a `max_tokens: 0` cannot collapse an output to a bare marker
  with no cue about what to expand. `head_lines: 0, tail_lines: 0` now routes here for the same
  reason — it used to reach the *line* path (`1 > 0+0`) and emit a bare 90-byte marker.

The 200-char floor and the 3-pass bound are now **documented** (they were not), because they
mean a very small `max_tokens` does not deliver what it says: `max_tokens: 100` measures ~118
tokens out. Two causes, both now stated in the code rather than implied away — the floor returns
before any measurement, and `got` measures `asm("")`, an *empty* marker, while the shipped text
carries the ~15 tokens of `<<cg:HASH>> [full output: ...]`. `tryMark`'s never-worse guard
measures the real marker and declines rather than growing a message, which is what makes the
residue harmless.

**Legality is inherited, not re-implemented.** The character path goes through exactly the same
shared helpers as the line path — `tryMark`/`commitMark` (reversible stash + `<<cg:HASH>>`
marker, and the marker-inclusive never-worse guard), `skipReduce` (idempotence), `freeze`
(replay at depth). The only new code is the window arithmetic.

### What I rejected

- **Making `linecap` own it** by loosening `neverTruncate`. Its exemptions exist because a
  600-char stack frame or compiler diagnostic is exactly the line a model needs whole, and the
  file carries measured evidence against narrowing the pattern. `collapse` is the component
  whose *stated contract* is "the last resort for anything still oversized" — a hole in the last
  resort belongs to the last resort. (#151/#140.)
- **A structure-aware JSON cut** (parse, keep the first N array elements). Strictly better
  output when it applies, but it is a new component's worth of work with a parse cost on every
  megabyte-scale payload, it fails on truncated/non-JSON bodies, and it is not what a
  *content-agnostic fallback* is for. A `jsoncut` component can be layered ahead of `collapse`
  later.
- **A token-based fall-through gate.** Rejected on cache-safety grounds; see above.
- **Grapheme-safety.** With decomposed content the kept tail can begin with an orphaned
  combining mark. Cosmetic, and the output stays valid UTF-8, which is the property that matters
  for splicing into a request body. Not worth a segmentation dependency here.

Earlier revisions rejected *"applying the character window as a second stage after the line
window"* on byte-identity grounds. That was the wrong call and review was right to block it: the
byte gate gets the same coverage **without** changing the bytes for anything the line window
already cut, because the halving test is passed by every shape it handles today by orders of
magnitude. The `ponytail` note recording that rejection is gone, along with the residual gap it
described.

### The counters

`too_few_lines` is **gone, not renamed**. Most of what it counted is now *handled*, and
exporting handled work under a decline's name yields a series that falls as the component works
better — the distinction `components/component.go`'s `Gate`/`Event` split exists for. Its
population is now:

- `char_window` **event** — the character path acted;
- `too_few_lines_and_chars` **gate** — the only genuine decline left: no window that shrinks the
  output, i.e. no usable line window *and* fewer characters than the character window's floor.

## Byte-identical layouts, both paths

`TestCollapseLineWindowIsByteIdentical` asserts the **exact expected string** for four
multi-line fixtures. `TestCollapseCharWindowIsByteIdentical` is the equivalent for the character
path, which review correctly identified as missing: its output is frozen and replayed at depth
in exactly the same way, so a changed layout re-anchors every cached position after it in every
live session mid-flight. Four fixtures (one-line JSON, multibyte, huge line in the kept head,
huge line in the kept tail), expectation written out as a literal format string rather than
captured from `window()`, with the head/tail split pinned to 100/100 by construction (at
`max_tokens: 50` the budget floors, so `runeWindow` returns at pass 0 and there is no
budget-dependent arithmetic in the expectation). Each case also asserts `char_window` fired, so
the comparison cannot pass vacuously via the line path.

`TestCollapseCharWindowAtExactlyTheLineWindowSize` pins the boundary this PR moves from the
other side: at exactly `head_lines+tail_lines` lines the line window would omit zero lines, so
it must not be taken.

## Mutation results

Standing rule: for each new test, revert the code it covers and confirm the test FAILS.
Review's separate point is taken — vacuity of an assertion is weaker than catching a plausible
wrong implementation — so this round is stated as mutation results, with the mutants review
found surviving.

**8 mutations, 7 caught, 1 equivalent.**

| mutation | outcome | failure |
|---|---|---|
| remove the byte gate (revert to the line-COUNT gate) | **CAUGHT** | `huge line in head: character window is not byte-identical` |
| revert the budget clamp to `maxTokens*4` | **CAUGHT** | `max_tokens 2000: 2061-token single-line output left uncut; gates map[too_few_lines_and_chars:1]` |
| flip the too-few-lines gate to `>=` | **EQUIVALENT** | see below |
| off-by-one head slice `r[:h+1]` | **CAUGHT** | `one line json: character window is not byte-identical` |
| off-by-one omitted count | **CAUGHT** | `huge line in tail: character window is not byte-identical` |
| byte slicing instead of rune slicing | **CAUGHT** | `character window split a multibyte rune` |
| remove the zero-line-window routing | **CAUGHT** | `a 72000-byte output collapsed to 90 bytes — a bare marker with no cue` |
| revert the `newCollapse` negative clamp | **CAUGHT** | `negative line knobs must be clamped to 0, got -10/-5` |

**On the `>=` survivor** — review's worst one, and it survives now for a different reason than
before. Under the byte gate it is an **equivalent mutant**, not a wrong implementation: at
exactly `head_lines+tail_lines` lines the mutant does take the line path and compute
`omitted = 0`, but `asm("")` is then the whole content plus ~24 characters, which fails the
halving test and falls through to the character window anyway. The behaviour the mutation used
to produce — whole content plus a marker, `marker_no_win`, a 40-line multi-megabyte output
sailing through a green suite — can no longer be expressed by that edit.
`TestCollapseCharWindowAtExactlyTheLineWindowSize` pins the boundary and is non-vacuous (the
byte-gate mutation fails it), but no test can distinguish equivalent code, and killing the
behaviour is a stronger outcome than catching the mutant. The other two survivors are caught.

The earlier round's 11 vacuity checks stand as reported; nothing in them is retracted.

## Follow-ups fixed here

- `head_lines: 0, tail_lines: 0` reached the line path and produced a bare marker with no
  content cue — 2,001 lines became 90 bytes, precisely what `collapseMinWindowChars` exists to
  prevent. Routed to the character window.
- Negative `head_lines`/`tail_lines` **panicked** on the line path (`1 > -10` sends even a
  one-liner there, then `lines[:-10]` slices out of range), which also made `splitWindow`'s
  `max(...)` guard unreachable for the config it looked like it defended. Clamped once in
  `newCollapse`, covering both paths; `splitWindow` now only handles zero.
- The `runeWindow` doc comment's claim that "an output above `max_tokens` comes back at or under
  it" is softened to the truth, with both causes named.
- The 200-char floor, the 3-pass bound, the seed clamp and the negative-knob clamp are
  documented in `docs/components/collapse.md`; `docs/components.md`'s "too small for either
  window to shrink it" now describes the gate rather than the intent.
- `[]rune` on a 14.8 MB payload allocates ~59 MB per oversized message per turn, on top of the
  `strings.Split` above it. Recorded as a `ponytail` with the measured rate that makes it
  tolerable for now (17 bodies in 3,347 requests).
- **Dashboard, from the follow-up review:** `char_window` rendered under a column header reading
  *"Why declined"* on a row that acted. The merged cell is deliberate (`dash/event.go` folded
  events in to fix an empty-cell bug) and is left merged; the header is now *"Why / what
  happened"* in `dash/ui/index.html`, `dash/ui/app.js` and `docs/dashboard.md`. `collapse` is the
  first component to surface an event there, which is why it was never revisited.

## Not fixed here, filed instead

- **#158 — the first savings bar renders 158% in a 100%-wide track.** The arithmetic is
  intended and documented (gross over attempted, re-counted every turn); the presentation is
  not. This PR pushes it from 115% to 158% precisely because `collapse` now catches enormous
  blobs, so it gets worse as the product gets better. Left out because it is a change to the
  headline savings display affecting all four denominator rows, and belongs with its own
  screenshot verification rather than bolted to a `components/offload` fix.
- **#151 / #140 — `linecap`'s `neverTruncate` exempts one-line JSON.** Pre-existing; the claim
  that it mitigated this shape is removed from this PR rather than the exemption being loosened.
- **#139 — `internal/tokens.Count` is superlinear on a single unbroken word.** 600 KB of
  separator-free CJK does not finish tokenizing inside a ten-minute `go test` timeout.
  Pre-existing (the size check has always tokenized full content), but note the population
  change: single-line payloads used to be skipped by `too_few_lines`, so `collapse` is now the
  component most likely to meet such an input. It is why the UTF-8 fixture keeps ASCII
  separators between the CJK runs.
- **#157 — `dash/TestSpendSurvivesRowEviction` is red for the first 10 hours of every month.**
  Found running the full suite for this PR; its fixture backdates 10 hours and today is the 1st,
  so 3 of 10 sessions land in the previous month. Reproduced 3/3 on unmodified `main`
  (`e88f5d8`), and this PR touches no `dash` Go code. Distinct from the two load-dependent
  flakes in #91.

## One review item that does not apply

Review's follow-up asks to fix a vacuous `combining` case using `"éà"` (which compiles to
precomposed runes in Go source). There is no `combining` case in this suite — the rune-safety
test is `TestCollapseCharWindowKeepsValidUTF8`, which uses CJK, where the multibyte property is
real and the byte-slicing mutation is caught. Nothing to fix; flagging it rather than silently
skipping it in case the note came from a different branch.

## Build and test

Eval box, Go 1.26.4, `CGO_ENABLED=1`: `gofmt` clean on every touched file, `go build ./...`
clean, `mkdocs build --strict` clean, full `go test ./...` green **except**
`dash/TestSpendSurvivesRowEviction`, which fails identically on unmodified `main` (#157).

## One correction to the brief

The brief quoted `collapse.go` without the depth gate, the freeze/replay logic, or the
`rep.Gate("non_text_blocks")` / `below_max_tokens` counters. `origin/main` has all of them, so
the fix had to slot into `TailOnlyCold` + `freeze` + `repairLostFreeze`. The window choice is
placed **before** the depth gate on purpose, so a message with no usable window is not
miscounted as a cache decline.

Separately, `docs/components/collapse.md` documents `cold_cache` as defaulting to `false` while
`coldCacheField()` defaults it to `true` (and `docs/components.md` says **true**). Pre-existing
inconsistency, unrelated, left alone.

### DIFF
--- a/components/offload/collapse.go
+++ b/components/offload/collapse.go
@@ -18,8 +18,8 @@ func init() { components.Register("collapse", newCollapse) }
 // cmdfilter/format), and skips anything already carrying a marker so it never
 // double-collapses.
 //
-// The window is cut by LINES when there are enough of them, and by CHARACTERS when
-// there are not. The character path is not a nicety. Being the last-resort catch-all,
+// The window is cut by LINES when that actually shrinks the output, and by CHARACTERS when
+// it does not. The character path is not a nicety. Being the last-resort catch-all,
 // collapse declining a payload means nothing caps it, and the line-only version
 // declined every output of head_lines+tail_lines lines or fewer — which includes the
 // single most common shape of an enormous tool result: a database or HTTP API response
@@ -35,11 +35,21 @@ func init() { components.Register("collapse", newCollapse) }
 // allow-list exempts any line matching `^\S+:\d+`, and a one-line JSON body
 // (`{"items":[{"id":1,...`) matches that by accident.
 //
-// ponytail: an output with MORE than head_lines+tail_lines lines still takes the line
-// window and can stay multi-megabyte if the kept lines are themselves huge (linecap,
-// which runs after, caps those lines unless neverTruncate exempts them). Deliberately
-// out of scope here so the line path stays byte-identical; the character window is a
-// natural second stage for it if that shape shows up in production.
+// The choice between the two windows is made on BYTES, not on the line count: an output
+// with more than head_lines+tail_lines lines whose kept head or tail is itself
+// multi-megabyte takes the CHARACTER window too. Gating on the line count alone left the
+// motivating shape alive above 40 lines — measured at 4,195,111 B in, 4,194,922 B out, and
+// worse than declining, because dropping the small filler lines in the middle counts as
+// acting and stamps a marker, which then makes linecap decline (marker_or_kept_verbatim)
+// and forwards 1.57 M tokens. linecap could not have capped it even without the marker:
+// its neverTruncate allow-list exempts any line matching `^\S+:\d+`, which one-line JSON
+// matches by accident (issue #151), so the mitigation this comment used to name is void
+// rather than conditional.
+//
+// ponytail: []rune(content) on the character path copies the whole output — ~59 MB for the
+// 14.8 MB payloads that motivated this component — once per oversized message per turn, on
+// top of the strings.Split above it. Fine at the measured rate (16 bodies in 3,347
+// requests) but it is the obvious thing to make streaming if that rate rises.
 type Collapse struct {
 	maxTokens int
 	maxFrac   float64
@@ -66,8 +76,12 @@ func newCollapse(raw []byte) (components.Component, error) {
 	if err := components.Decode(raw, &cfg); err != nil {
 		return nil, err
 	}
-	return &Collapse{maxTokens: cfg.MaxTokens, maxFrac: cfg.MaxFrac, headLines: cfg.HeadLines,
-		tailLines: cfg.TailLines, mode: parseMarkerMode(cfg.MarkerMode), coldCache: coldCacheDefault(cfg.ColdCache)}, nil
+	// Clamp here rather than at each use: a negative head_lines used to panic on the line
+	// path (`1 > -10` sends even a one-liner there, then lines[:-10] slices out of range),
+	// which fail-open contained but which also made splitWindow's own guard unreachable for
+	// the config it looked like it defended.
+	return &Collapse{maxTokens: cfg.MaxTokens, maxFrac: cfg.MaxFrac, headLines: max(cfg.HeadLines, 0),
+		tailLines: max(cfg.TailLines, 0), mode: parseMarkerMode(cfg.MarkerMode), coldCache: coldCacheDefault(cfg.ColdCache)}, nil
 }
 
 func (Collapse) Name() string                 { return "collapse" }
@@ -180,20 +194,35 @@ const collapseMinWindowChars = 200
 // after it in every live session mid-flight.
 func (cl *Collapse) window(content string, maxTokens int, rep *components.Report) (assemble func(string) string, byChars, ok bool) {
 	lines := strings.Split(content, "\n")
-	if len(lines) > cl.headLines+cl.tailLines {
+	if cl.headLines+cl.tailLines > 0 && len(lines) > cl.headLines+cl.tailLines {
 		omitted := len(lines) - cl.headLines - cl.tailLines
 		head := strings.Join(lines[:cl.headLines], "\n")
 		tail := strings.Join(lines[len(lines)-cl.tailLines:], "\n")
-		return func(tok string) string {
+		asm := func(tok string) string {
 			return fmt.Sprintf("%s\n... (%d lines omitted) %s\n%s", head, omitted, tok, tail)
-		}, false, true
+		}
+		// Take the line window only if it actually CUT: having more lines than the window
+		// says nothing about their size, and a multi-megabyte line inside the kept head or
+		// tail survives it untouched. Halving is the test because the line window is the
+		// byte-identical path and every shape it handles today clears it by orders of
+		// magnitude; the shapes that do not are the ones this fall-through exists for.
+		//
+		// BYTES, deliberately, and not a token comparison against maxTokens: maxTokens comes
+		// from resolveBudget(..., c.CtxWindow), CtxWindow can resolve differently mid-session,
+		// and repairLostFreeze re-derives at depth (see Offload). A token gate could therefore
+		// classify the same content line-window on one turn and character-window on the next
+		// and emit different bytes, trading a savings bug for a cache bug. This form depends
+		// only on content, head_lines and tail_lines, so it is replay-safe unconditionally.
+		if len(asm(""))*2 <= len(content) {
+			return asm, false, true
+		}
 	}
 	asm, ok := cl.runeWindow(content, maxTokens)
 	if !ok {
-		// The only genuine decline left: too few lines for a line window AND too few
-		// characters for a character one. `too_few_lines` is deliberately NOT reused —
-		// most of what it counted is now HANDLED, and exporting handled work under a
-		// decline's name yields a series that falls as the component works better.
+		// The only genuine decline left: no usable line window AND fewer characters than the
+		// smallest character one. `too_few_lines` is deliberately NOT reused — most of what
+		// it counted is now HANDLED, and exporting handled work under a decline's name yields
+		// a series that falls as the component works better.
 		rep.Gate("too_few_lines_and_chars")
 		return nil, false, false
 	}
@@ -207,12 +236,23 @@ func (cl *Collapse) window(content string, maxTokens int, rep *components.Report
 // The budget is the component's own token threshold expressed in characters, split
 // between head and tail in the head_lines:tail_lines ratio, so the existing knobs keep
 // meaning what they say ("how much of the start / the end is kept") without a second set
-// of char knobs to keep in sync. The assembled window is then measured against maxTokens
-// and the budget scaled down if it overshoots, which is what keeps the promise the
-// threshold makes: an output above max_tokens comes back at or under it.
+// of char knobs to keep in sync. The assembled window is then measured against maxTokens and
+// the budget scaled down if it overshoots, which brings an output above max_tokens back to
+// roughly it — not exactly it, on two counts worth naming rather than implying otherwise.
+// The collapseMinWindowChars floor returns before any measurement, so a very small
+// max_tokens buys a 200-character window regardless (max_tokens: 100 measures 118 tokens
+// out). And `got` measures asm("") — an EMPTY marker — while the shipped text carries the
+// ~15 tokens of `<<cg:HASH>> [full output: ...]`, which only the 10% margin below absorbs.
+// tryMark's never-worse guard is what makes the residue harmless: it measures the real
+// marker and declines rather than growing the message.
 func (cl *Collapse) runeWindow(content string, maxTokens int) (func(string) string, bool) {
 	r := []rune(content)
-	budget := maxTokens * collapseCharsPerToken
+	// Never seed above the content. chars/4 OVERESTIMATES on dense JSON (~2.5 chars/token),
+	// so bailing out below on the un-measured seed declined content the fit loop would have
+	// cut: at max_tokens 3000 — what the shipped codesafe preset sets — one-line JSON from
+	// 3,006 to 4,491 tokens was oversized and left entirely uncut. marker_no_win already
+	// protects the downside of seeding low.
+	budget := min(maxTokens*collapseCharsPerToken, len(r)-1)
 	for pass := 0; ; pass++ {
 		if budget < collapseMinWindowChars {
 			budget = collapseMinWindowChars
@@ -245,10 +285,13 @@ func (cl *Collapse) runeWindow(content string, maxTokens int) (func(string) stri
 
 // splitWindow divides a character budget between head and tail in the head_lines:tail_lines
 // ratio, so `head_lines: 40, tail_lines: 10` keeps four fifths of the window at the start on
-// the character path too. Both zero (or negative) means an even split rather than an empty
-// window, since a component configured with no line window still has to cap the output.
+// the character path too. Both zero means an even split rather than an empty window, since a
+// component configured with no line window still has to cap the output — and with both zero
+// window() routes here rather than to the line path, which would otherwise emit a bare
+// marker with no cue about what to expand (a 64 KB output became 87 bytes). Negatives are
+// clamped in newCollapse, so this only has to handle zero.
 func (cl *Collapse) splitWindow(budget int) (head, tail int) {
-	hl, tl := max(cl.headLines, 0), max(cl.tailLines, 0)
+	hl, tl := cl.headLines, cl.tailLines
 	if hl+tl == 0 {
 		hl, tl = 1, 1
 	}
@@ -263,9 +306,9 @@ func init() {
 		{Key: "max_frac", Type: components.FieldFloat,
 			Hint: "The same threshold as a fraction of the model's context window; wins when the window is known. 0 = unset."},
 		{Key: "head_lines", Type: components.FieldInt, Default: 20,
-			Hint: "Lines kept from the start of a collapsed output. On an output with too few lines to cut (a one-line JSON payload), the same ratio splits a CHARACTER window sized from max_tokens."},
+			Hint: "Lines kept from the start of a collapsed output. When a line window would not shrink it (a one-line JSON payload, or a huge line inside the kept window), the same ratio splits a CHARACTER window sized from max_tokens."},
 		{Key: "tail_lines", Type: components.FieldInt, Default: 20,
-			Hint: "Lines kept from the end of a collapsed output. Also the tail share of the character window used when there are too few lines."},
+			Hint: "Lines kept from the end of a collapsed output. Also the tail share of the character window used when a line window would not shrink the output."},
 		markerModeField(),
 		coldCacheField(),
 	})
--- a/components/offload/collapse.go
+++ b/components/offload/collapse.go
@@ -17,6 +17,29 @@ func init() { components.Register("collapse", newCollapse) }
 // full original, and leaves an expand marker. Runs late in the pipeline (after
 // cmdfilter/format), and skips anything already carrying a marker so it never
 // double-collapses.
+//
+// The window is cut by LINES when there are enough of them, and by CHARACTERS when
+// there are not. The character path is not a nicety. Being the last-resort catch-all,
+// collapse declining a payload means nothing caps it, and the line-only version
+// declined every output of head_lines+tail_lines lines or fewer — which includes the
+// single most common shape of an enormous tool result: a database or HTTP API response
+// serialised as ONE line of JSON. Measured on a live 128k-band arm: 16 upstream 400s
+// ("prompt is too long") in 3,347 requests on outgoing bodies of 2.6 MB to 14.8 MB,
+// and 17 of 75 runs errored. Nothing else caught them either — extract acted 0 times
+// (no_obvious_noise 16,891, it only strips KNOWN noise patterns), cmdfilter 0
+// (no_filter_match, only known command families), toon 21 (not_uniform_object_array
+// 233,501), dedup 774 (exact duplicates only), extract_llm declines by design when the
+// output exceeds the compaction model's own window (over_model_context), and
+// summarize protects keep_last, which is exactly where a FRESH oversized output sits.
+// linecap's per-line cap does not rescue this shape either: its neverTruncate
+// allow-list exempts any line matching `^\S+:\d+`, and a one-line JSON body
+// (`{"items":[{"id":1,...`) matches that by accident.
+//
+// ponytail: an output with MORE than head_lines+tail_lines lines still takes the line
+// window and can stay multi-megabyte if the kept lines are themselves huge (linecap,
+// which runs after, caps those lines unless neverTruncate exempts them). Deliberately
+// out of scope here so the line path stays byte-identical; the character window is a
+// natural second stage for it if that shape shows up in production.
 type Collapse struct {
 	maxTokens int
 	maxFrac   float64
@@ -86,10 +109,12 @@ func (cl *Collapse) Offload(req *schemas.BifrostChatRequest, rep *components.Rep
 			rep.Gate("marker_or_kept_verbatim") // already offloaded, or expanded by the agent
 			continue
 		}
-		lines := strings.Split(content, "\n")
-		if len(lines) <= cl.headLines+cl.tailLines {
-			rep.Gate("too_few_lines") // few long lines; head/tail wouldn't help
-			continue
+		// Pick the window BEFORE the depth gate so a message that has no usable window is
+		// not charged as a cache decline. assemble places the marker token; it is a pure
+		// function of (content, config, maxTokens), which is what lets freeze replay it.
+		assemble, byChars, ok := cl.window(content, maxTokens, rep)
+		if !ok {
+			continue // window() named the reason
 		}
 		// A NEW collapse only in the uncached tail. This component used to carry no depth
 		// restriction at all, contradicting the contract in components/component.go: it
@@ -102,13 +127,7 @@ func (cl *Collapse) Offload(req *schemas.BifrostChatRequest, rep *components.Rep
 			rep.Gate("cached_prefix")
 			continue
 		}
-		omitted := len(lines) - cl.headLines - cl.tailLines
-		head := strings.Join(lines[:cl.headLines], "\n")
-		tail := strings.Join(lines[len(lines)-cl.tailLines:], "\n")
-		newText, key, eff, ok := tryMark(c, cl.mode, content, " [full output: call "+expand.ToolName+"]",
-			func(tok string) string {
-				return fmt.Sprintf("%s\n... (%d lines omitted) %s\n%s", head, omitted, tok, tail)
-			})
+		newText, key, eff, ok := tryMark(c, cl.mode, content, " [full output: call "+expand.ToolName+"]", assemble)
 		if !ok {
 			rep.Gate("marker_no_win") // head/tail window+marker wouldn't shrink this output
 			continue
@@ -116,6 +135,12 @@ func (cl *Collapse) Offload(req *schemas.BifrostChatRequest, rep *components.Rep
 		commitMark(c, rep, eff, key, content)
 		schema.SetMessageText(m, newText)
 		freeze(c, cl.Name(), content, newText) // freeze so later turns replay it (no churn)
+		if byChars {
+			// Counted as an EVENT, not a gate: this population used to be the
+			// `too_few_lines` DECLINE, and exporting "handled" under the old decline name
+			// would make the series fall as the component works better.
+			rep.Event("char_window")
+		}
 		changed++
 		if key != "" {
 			keys = append(keys, key)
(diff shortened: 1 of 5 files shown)


OsherElhadad:
Reviewed against main `e88f5d8` in a clean worktree, including a live proxy under
`--preset codesafe`. `go build ./...` clean, `go test ./...` → 27 packages ok / zero FAIL,
`mkdocs build --strict` clean. Docs are accurate against the code — every claim in
`docs/components/collapse.md` checked against `collapse.go` (`too_few_lines_and_chars:197`,
`char_window:142`, `collapseCharsPerToken=4:162`, rune-based `:214`, tightening-only `:238`),
and retiring the old gate name rather than reusing it was the right instinct.

**The character window itself is correct on the properties most likely to be wrong.** Verified
live, not read:

- **Slices runes, not bytes** (`r := []rune(content)`). Emoji, CJK, 4-byte and mixed content all
  stay valid UTF-8, including a 6-offset sweep walking the cut through every byte of a 3-byte
  rune. Mutating it to byte slicing is caught by the new suite.
- **Reversible.** `GET /expand?id=` byte-matched both fixtures (204891/204891, 180000/180000),
  and cross-session access correctly 404s.
- **Deterministic.** Byte-identical across 3 turns in one session and 2 fresh sessions (same md5).
- **Never-worse holds in tokens.** `tryMark` refuses before any side effect; 201–380 B inputs are
  correctly declined `marker_no_win`, so the smallest input actually shrunk is ~400 B.
- **Skip-if-already-marked is honoured structurally** — `skipReduce` at `:108` precedes
  `window()`, so one call site covers both paths.

9 of 12 mutations I wrote are caught. This is careful work. The problem is that the gate is in
the wrong units.

## 1. Blocking — the fix is gated on LINE COUNT, not size, so the motivating shape survives above 40 lines

`collapse.go:183` takes the line path whenever there are *more* than `head_lines+tail_lines`
lines, regardless of how big those lines are. A multi-megabyte line inside the kept head or tail
window survives untouched — and because `collapse` still "acts" (it drops the small filler lines
in the middle) it stamps a marker, which then makes `linecap` decline:

```
huge line in head: in=4195111B  after collapse=4194922B (events=map[])
                   after linecap=4194922B (gates=map[marker_or_kept_verbatim:1])  tokens=1570220
huge line in tail: in=4195110B  after collapse=4194921B (events=map[])
```

**4 MB in, 189 bytes saved, 1.57 M tokens forwarded.** `events` is empty, so the char window never
engaged.

And `linecap` could not have rescued it even without the marker. The commit's ponytail note says
*"linecap, which runs after, caps those lines unless neverTruncate exempts them"* — but the commit
also correctly observes that `^\S+:\d+` matches `{"items":[{"id":1,...` by accident, so for the
exact payload shape this PR exists to handle, `neverTruncate` **always** exempts it:

```
linecap alone: in=2097191B out=2097191B acted=False gates=map[no_long_or_repeated_lines:1] neverTruncate(line)=true
```

The named mitigation is void, not conditional. (Filed separately as #151, since it is
pre-existing.)

**Compounding this: the evidence does not establish the fix covers the measured failures.** The
commit cites *"A digest of one failing 14.8 MB body held only 4 messages, so a single tool_result
was multiple megabytes"* — that is a **message** count, not a **line** count. Nothing in the PR
shows the 16 failing bodies had ≤40 lines, so nothing shows this change would have prevented any
of them. A line-count histogram of those 16 bodies converts this from plausible to demonstrated,
and it looks like data you already have.

**Fix — move the gate to size.** Use the **byte-based** form, not a token-based one:

```go
if len(lines) > cl.headLines+cl.tailLines {
    asm := /* existing line closure, unchanged */
    if len(asm(""))*2 <= len(content) {
        return asm, false, true   // line window really cut it: byte-identical, as today
    }
    // line window kept most of the bytes (a giant kept line) -> fall through to runeWindow
}
```

A token-based gate reading `maxTokens` would **not** be replay-safe: `maxTokens` comes from
`resolveBudget(..., c.CtxWindow)`, `CtxWindow` can change mid-session, and `repairLostFreeze`
(`:126`) re-derives at depth — which `collapse.go:119-125` already warns about. Under `max_frac`
that could flip line→char for the same content and emit different bytes, trading a savings bug for
a cache bug. The byte form depends only on content plus `head_lines`/`tail_lines`, so it is
unconditionally replay-safe. Both forms classify all six cases identically, including the four
fixtures from `TestCollapseLineWindowIsByteIdentical`, which is what preserves byte-identity for
everything the line window handles today.

If you would rather not change the gate, the alternative is to amend the PR body: state the
residual hole plainly, drop the `linecap` claim, and show the histogram.

## 2. Wanted before merge — a units mismatch declines a 55%-wide oversize band, widest in a shipped preset

`collapse.go:215` seeds `budget := maxTokens * collapseCharsPerToken` (×4) and `:221` tests the
bail-out against that **un-measured** seed, before the fit loop runs. Dense JSON tokenizes at
~2.5 chars/token, so content genuinely oversized in tokens is still shorter in runes than a
4-chars/token budget and is declined outright:

```
runes=6015 tokens=2304 oversized=true acted=false gates=map[too_few_lines_and_chars:1]
runes=7958 tokens=3033 oversized=true acted=false gates=map[too_few_lines_and_chars:1]
runes=8180 tokens=3114 oversized=true acted=true  events=map[char_window:1]

max_tokens=2000: one-line JSON from 2007 to 3033 tokens is OVERSIZED and still UNCUT
max_tokens=3000: one-line JSON from 3006 to 4491 tokens is OVERSIZED and still UNCUT
```

`codesafe` sets `collapse.max_tokens: 3000`, so the **shipped deterministic preset has the wider
3006–4491-token hole**, and no test covers 3000. Since `collapse` is the last resort, a decline
means nothing caps the output — and the gate name makes it look on `/stats` like a legitimately
unhandleable payload when a 200-char window would have cut it 96%.

One line; `marker_no_win` already protects the downside:

```go
-	budget := maxTokens * collapseCharsPerToken
+	// Never seed above the content: chars/4 overestimates on dense JSON, and bailing out on
+	// the un-measured seed declines content the fit loop would have cut.
+	budget := min(maxTokens*collapseCharsPerToken, len(r)-1)
```

## 3. Wanted before merge — no byte-exact assertion on the char-window layout

Three of my twelve mutations survived all 8 new tests, and they are the same gap:

```
MUT flip too-few-lines gate to >=:   *** SURVIVED ***
MUT off-by-one head slice (r[:h+1]): *** SURVIVED ***
MUT off-by-one omitted count:        *** SURVIVED ***
```

The char path's output is frozen and replayed at depth exactly like the line path, and the PR
argues at `:177-180` why that matters — a changed layout *"re-anchors every cached position after
it in every live session mid-flight"*. The line path got
`TestCollapseLineWindowIsByteIdentical` on that basis; the char path got no equivalent.

The `>` → `>=` survivor is the worst: at exactly 40 lines the mutant takes the line path, computes
`omitted = 0`, emits the whole content plus a marker, fails `tryMark` and declines
`marker_no_win`. A 40-line multi-MB output would silently sail through — the precise boundary this
PR moves — with a green suite.

`TestCollapseCharWindowIsByteIdentical` mirroring the line-path test's construction (spell the
expected string out rather than capture it), plus one `len(lines) == headLines+tailLines` case
asserting `char_window` fires, pins all three.

*(Note: the PR body's "11 mutations, 11 required failures, 0 vacuous" is about vacuity of the new
assertions — a weaker and different property than catching a plausible wrong implementation. Both
can be true; I am not disputing that table.)*

## Follow-ups

- `collapse.go:210-212` claims *"an output above max_tokens comes back at or under it"*. False for
  small budgets: `max_tokens=100` → 118 tokens out. Two causes — the 200-char floor returns at
  `:229` before any measurement, and `got := TextTokens(asm(""))` measures an **empty** marker
  while the shipped text carries ~15 tokens of `<<cg:HASH>> [full output: ...]`. Only the
  incidental 10% margin absorbs that. Soften the comment or measure `asm(token)`.
- `head_lines: 0, tail_lines: 0` → a 64 KB output becomes 87 bytes, a bare marker with no content
  cue, which is exactly what `collapseMinWindowChars` exists to prevent — reached via the line
  path (`1 > 0+0`), which the new floor does not guard.
- Negative `head_lines`/`tail_lines` panic on the line path (`1 > -10` sends even a one-liner
  there), so `splitWindow`'s new `max(cl.headLines, 0)` at `:251` is unreachable for the config it
  appears to defend. Pre-existing and fail-open contains it, but clamping once in `newCollapse`
  covers both paths and makes the new guard mean what it reads as.
- `docs/components.md:284` — "too small for either window to shrink it" describes intent, not the
  gate (see item 2). The 200-char floor and 3-pass bound are also undocumented, so `max_tokens: 50`
  silently yields a window the user did not ask for.
- Rune-safe but not grapheme-safe: with decomposed content the kept tail can start with an orphaned
  combining mark. Cosmetic (still valid UTF-8). Related: the new test's `combining` case is
  vacuous — Go source `"éà"` compiles to precomposed single runes; use `"é"` to cover it.
- `collapse.go:214` — `[]rune` on a 14.8 MB payload allocates ~59 MB per oversized message per
  turn, on top of the `strings.Split` at `:182`. Worth a `ponytail:` note given the PR's own
  motivating sizes.

## Two framing suggestions for the PR body

Neither is stated, and both make the case stronger:

1. **Say plainly that the payoff is avoided hard failures, not a savings line.** `collapse` is
   tail-gated and freezes its decision, so a new char-window cut only happens on the turn the
   output arrives — the cold tail, which is exactly where a multi-MB `tool_result` lands and where
   the 400 fires. A reader primed by the four-denominator convention will look for a savings
   number and read its absence as a weak result.
2. **Lead with 17 of 75 errored runs.** That is the honest headline and it is stronger than any
   token count.

## Generalization

The class is evidenced, not speculative (16 upstream 400 `prompt is too long` in 3,347 requests,
bodies 2.6–14.8 MB). It generalizes cleanly across agents and dialects because it is
content-shaped rather than agent-shaped — confirmed by driving the **OpenAI** dialect
(`role:"tool"`) through `/compact` and watching the char window fire (`char_window: 2`). No sibling
component shares the too-few-lines hole (`mask`/`failed_run` are whole-message, `smartcrush` works
on `[]json.RawMessage`, `extract` has no window, the DSL's `truncate_lines_at` is rune-safe), so
`collapse` is the right file — the fix is just bolted to the cutting mechanism rather than to the
gate that causes the problem.

## Verdict

**Needs changes.** Item 1 is the blocker and it is a scope/evidence problem rather than a defect
in the code that was written; items 2 and 3 are each ~1 line and ~1 test. The character window is
correct where it counts.

OsherElhadad:
Follow-up: a dashboard agent drove the live UI (headless Chromium, real proxy, real traffic) against
this branch and `main`. **This PR is a real measured win, and it surfaces in the dashboard with zero
`dash/` changes** — but two rendering details are worth a line each.

## First, the confirmation

Same novel high-entropy 200KB single-line blob through both builds:

```
main       before=113975  after=113975  saved_unique=0        (nothing acted — linecap's neverTruncate didn't catch it either)
integrated before=113975  after=1824    saved_unique=112151
```

So the PR's "nothing else catches this shape" claim **holds** on independent traffic, and the
`linecap`-cannot-help half of my earlier finding 1 is confirmed from the other direction. The
component table picked it up automatically — `runs 2 acted 2 100.0% unique 170k gross 170k overcount
1.0× 24ms` — and the new `char_window` event reached both `/api/components` and the request drawer
with no dashboard code changes at all. That is a good sign about the event/gate split you chose.

The inline diff also renders and wraps correctly: longest text node 4,016 chars, no page overflow, 0
JS console errors, all four view modes, and the `…(196831 characters omitted)` marker present. I went
looking for a silent-truncation bug and there isn't one — `before` is honestly capped with
`…[+196009 bytes]`.

## 1. LOW — `char_window` renders under a column header that says "Why declined", on a row that acted

`dash/ui/index.html:481` and `dash/ui/app.js:3297` (`el('th',{text:'Why declined'})`).

Your data model is right: `/api/components` keeps `events:{char_window:1}` separate from `gates:{}`,
and the commit message is explicit about why — *"Counted as an EVENT, not a gate… exporting 'handled'
under the old decline name would make the series fall as the component works better."* The UI then
renders both into one cell under a decline-only header:

```
request drawer, row 8:  collapse  offload  112k  112k  15ms  [acted]   char_window 1
                                                                        ^ column header: "WHY DECLINED"
```

`dash/event.go:417` records that events were folded into the gates cell to fix an empty-cell bug — so
the merge is deliberate and correct; the header was just never revisited. Fix is one string in two
files: **"Why / what happened"**. Do not un-merge the cell.

Not your PR's defect, but your PR is what makes it visible, so it seems worth raising here rather
than in isolation.

## 2. MEDIUM (pre-existing, and this PR makes it worse) — the first savings bar now reads 158%

```
main       pct=115.42  available=True
integrated pct=158.17  available=True
```

Rendered as a bar labelled `158.17%` with subtitle `409k ÷ 259k tokens`, in a track that is 100% by
construction — so the bar is pinned full while the number contradicts it, as the **first** figure on
the page.

The arithmetic is intended (`docs/dashboard.md:80`: gross-over-attempted, because `attempted_tokens`
is re-counted every turn). The presentation is not: a >100% figure under a heading "Savings
percentage" reads as a bug to anyone who has not read that paragraph. **Your PR pushes it from 115% to
158%** precisely because collapse now catches enormous blobs — it gets worse as the product gets
better.

It is a ratio, not a share. Render `1.58×` with the track at `min(1, …)`, or drop the track and show
the bare figure with its divisor. Don't touch the arithmetic. Raising it here because this PR is what
moves it past the point where a reader can charitably ignore it.

Neither item changes my verdict — the blocking items remain the line-count gate (finding 1), the
`max_tokens × 4` seed (finding 2), and the missing byte-exact char-window assertion (finding 3).

amiddavid:
Thanks — the gate/cutting distinction was the right diagnosis, and finding a second copy of the
same units mistake one level up from the one I fixed is the review I needed. All three blocking
items are addressed in `4ce7c33`, plus every follow-up except two I filed instead. Details, the
mutation results, and the histogram outcome below.

## 1. Blocking — the gate is now in BYTES

Taken as proposed, in the byte form and for the reason you gave:

```go
if cl.headLines+cl.tailLines > 0 && len(lines) > cl.headLines+cl.tailLines {
    asm := /* the existing line closure, unchanged */
    if len(asm(""))*2 <= len(content) {
        return asm, false, true // the line window really cut it: byte-identical, as before
    }
    // it kept most of the bytes (a giant kept line) -> fall through to the rune window
}
```

The cache argument against a token gate is decisive and I have written it into the code rather
than just the PR, so the next person to reach for `maxTokens` here sees why not: `resolveBudget`
+ a mid-session `CtxWindow` change + `repairLostFreeze` re-deriving at depth means a token gate
could classify the same content differently on a later turn and emit different bytes. Trading a
savings bug for a cache bug is a bad trade and the byte form costs nothing to get right.

The `ponytail` note that declared this shape out of scope is **gone**, and so is its `linecap`
claim. You are right that it was void rather than conditional — `neverTruncate(line)=true` on
exactly the payload shape the component exists for. Removed from the code comment, the commit
message and the PR body, with #151 referenced where the mechanism is described.

Two things I did beyond the proposal, both small:

- I gated the line branch on `headLines+tailLines > 0` as well, which folds in the
  `head_lines: 0, tail_lines: 0` follow-up: that config used to reach the *line* path via
  `1 > 0+0` and emit a bare 90-byte marker.
- `TestCollapseCutsAHugeLineInsideTheLineWindow` reproduces both of your fixtures (huge line in
  head, huge line in tail) and asserts the output at least halves, stays under `max_tokens`,
  fires `char_window`, and is still byte-recoverable from the stash.

## The histogram — it cannot be produced, and the reason is worse than inconvenience

You were right that the "4 messages" line is a message count doing a line count's job, and right
that it looked like data I already had. It is not. I went back to `capfail-i21A.jsonl` on the old
box (read-only; nothing touched). The digest schema is:

```
status, body_bytes, rig_seq, n_messages, system_present, error,
digest[] = { i, role, types[], tool_use[], tool_result[] }
```

Per message that is **structure only** — index, role, block-type names, tool-use ids. No content,
not even a truncated head (`capture_hop.py`'s `data[:600]` truncation applies to the *error*
string, not the request body, which is discarded after `len(raw)`). No per-message or per-block
size of any kind, so the largest `tool_result` cannot even be identified, let alone counted in
lines. The only size datum is whole-request `body_bytes`, which includes JSON escaping, the
system prompt and 94–149 tool definitions. `rig_seq` — the field that exists precisely so a
record can be joined back to the request the shim handed the proxy — is `null` in all records,
and no raw copy survives (`proxy-i21A.log` and `flap-i21A.jsonl` record bytes, never lines).

So I have taken your alternative for the evidence half and stated it plainly in the PR body
rather than leaving the inference standing: **the question cannot be answered from this
experiment record, in either direction.** The case rests on the class being real and `collapse`
being the only component contracted to catch it, plus the reproduced mechanism in tests and your
live measurements — not on a demonstrated one-to-one with those failures.

One correction that came out of the same dig, in your favour and against my own body: the digest
holds **19** records — 17 × `400 prompt is too long` plus 2 transport-level failures
(`IncompleteRead`, read timeout) that are not provider rejections. Both of us have been writing
16. I cannot reconstruct that filter, so the PR now says 17 and names the 2 non-400s. And the
capture needs a newline count per block before the next arm; that is a rig change, not something
this PR can retrofit.

## 2. The oversize band — taken as written

```go
budget := min(maxTokens*collapseCharsPerToken, len(r)-1)
```

`TestCollapseCutsDenseJSONInTheOversizeBand` covers `max_tokens` 2000 **and 3000**, sizing each
fixture into the band programmatically and asserting the preconditions (over the threshold in
tokens, under it in chars/4 runes) so it cannot pass vacuously if the band ever moves. Reverting
the clamp fails it with `2061-token single-line output left uncut; gates
map[too_few_lines_and_chars:1]`.

Your point about the gate *name* making a decline look legitimate on `/stats` is the part that
stuck with me — that is the failure mode where nobody ever finds out.

## 3. Byte-exact char-window assertion — added, and two of your three survivors are caught

`TestCollapseCharWindowIsByteIdentical`, mirroring the line-path test: expectation spelled out as
a literal format string, not captured from `window()`, over four fixtures (one-line JSON,
multibyte, huge line in the kept head, huge line in the kept tail). The head/tail split is
pinned to 100/100 by construction — at `max_tokens: 50` the budget floors at
`collapseMinWindowChars`, so `runeWindow` returns at pass 0 and there is no budget-dependent
arithmetic hiding in the expectation. Every case also asserts `char_window` fired, so the
comparison cannot pass vacuously through the line path.

**8 mutations, 7 caught, 1 equivalent:**

| mutation | outcome | failure |
|---|---|---|
| remove the byte gate (revert to the line-COUNT gate) | **CAUGHT** | `huge line in head: character window is not byte-identical` |
| revert the budget clamp to `maxTokens*4` | **CAUGHT** | `max_tokens 2000: 2061-token single-line output left uncut` |
| flip the too-few-lines gate to `>=` | **EQUIVALENT** | see below |
| off-by-one head slice `r[:h+1]` | **CAUGHT** | `one line json: character window is not byte-identical` |
| off-by-one omitted count | **CAUGHT** | `huge line in tail: character window is not byte-identical` |
| byte slicing instead of rune slicing | **CAUGHT** | `character window split a multibyte rune` |
| remove the zero-line-window routing | **CAUGHT** | `a 72000-byte output collapsed to 90 bytes — a bare marker with no cue` |
| revert the `newCollapse` negative clamp | **CAUGHT** | `negative line knobs must be clamped to 0, got -10/-5` |

**On the `>=` mutant — your worst survivor — it still survives, and I want to be precise about
why rather than claim a kill.** Under the byte gate it is an *equivalent mutant*, not a wrong
implementation. At exactly `head_lines+tail_lines` lines the mutant does take the line path and
compute `omitted = 0`, but `asm("")` is then the whole content plus ~24 characters, which fails
the halving test and falls through to the character window anyway. I verified this rather than
argued it: with the mutation applied, `TestCollapseCharWindowAtExactlyTheLineWindowSize` passes
*because `char_window` fires* — the exact behaviour you described (whole content plus a marker,
`marker_no_win`, a 40-line multi-megabyte output through a green suite) can no longer be
expressed by that edit. The boundary test is there and is non-vacuous (the byte-gate mutation
fails it), but no test distinguishes equivalent code. I think killing the behaviour is the
stronger outcome; if you would rather have a test that also kills the *edit*, say so and I will
add a direct `window()` unit assertion on `byChars` at the boundary — it will be green under both
versions, so it would be a documentation test, not a mutation kill.

Your note about the earlier "11 mutations / 0 vacuous" table being a claim about vacuity rather
than about catching plausible wrong implementations is fair and I have restated this round as
mutation results.

## Follow-ups — all fixed except two

- **`:210-212` comment false for small budgets.** Softened, with both causes named in the code
  (the floor returns before measurement; `got` measures `asm("")`, an empty marker, while the
  shipped text carries ~15 tokens) and `tryMark`'s guard named as what makes the residue
  harmless. I chose the comment over measuring `asm(token)` because `window()` does not yet know
  the hash — measuring a same-length placeholder is doable but changes the emitted bytes of a
  path that is now frozen and replayed, for a residue the never-worse guard already contains.
  Happy to revisit if you disagree.
- **`head_lines: 0, tail_lines: 0` → bare marker.** Fixed, routed to the character window.
- **Negative knobs panic / unreachable `max(...)`.** Clamped once in `newCollapse`, as you
  suggested — it covers both paths, and `splitWindow` now only has to handle zero, so its guard
  means what it reads as.
- **Undocumented 200-char floor and 3-pass bound.** Documented, with the seed clamp and the
  negative clamp, and `docs/components.md`'s "too small for either window to shrink it" now
  describes the gate.
- **`[]rune` on 14.8 MB.** Recorded as a `ponytail` with the rate that makes it tolerable
  (17 bodies in 3,347 requests) and the note that streaming is the obvious fix if that rises.
- **Grapheme-safety.** Declined, as cosmetic-and-still-valid-UTF-8. Recorded in the PR's rejected
  list rather than left silent.
- **The `combining` case.** This one I cannot action: there is no `combining` case in the suite.
  The rune-safety test is `TestCollapseCharWindowKeepsValidUTF8`, which uses CJK — the multibyte
  property is real there and the byte-slicing mutation is caught. Possible the note came from a
  different branch; flagging rather than skipping it.

## Dashboard follow-up

- **Header.** Fixed as you specified — one string in `dash/ui/index.html` and `dash/ui/app.js`,
  now *"Why / what happened"*, plus the paragraph in `docs/dashboard.md` that described the
  column as gate counts only. Cell left merged; `dash/event.go`'s reason for merging is sound and
  `collapse` being the first component to surface an event there is why the header was never
  revisited.
- **The 158% bar.** Filed as **#158** rather than fixed here, with your measurements, the
  `renderDenominators` `max: 100` mechanism, and three fix options (`1.58×` with the track at
  `min(1, …)` recommended, and *do not touch the arithmetic* stated explicitly). It changes the
  headline savings display and affects all four denominator rows, so it wants its own screenshot
  verification rather than riding along on a `components/offload` fix. You are right that this PR
  is what pushes it past charitable-ignoring, and I have said so in the issue.

## Framing

Both taken. The body now leads with **17 of 75 errored runs** and says explicitly that the payoff
is avoided hard failures rather than a savings line, with the tail-gate/freeze reasoning for why
no savings series will show it — plus your dashboard agent's 113,975 → 1,824 measurement as the
independent confirmation that the cut is real when it fires.

## Verification

Eval box, Go 1.26.4, `CGO_ENABLED=1`: `gofmt` clean, `go build ./...` clean,
`mkdocs build --strict` clean, full `go test ./...` green **except**
`dash/TestSpendSurvivesRowEviction` — which fails 3/3 on unmodified `main` (`e88f5d8`) too. Its
fixture backdates 10 hours and today is the 1st, so 3 of its 10 sessions land in the previous
month and `MonthToDateUSD` correctly returns 7 instead of 10. Filed as **#157** with fix options;
distinct from the two load-dependent flakes in #91, and it would have been red for you a few
hours earlier in the day. This PR touches no `dash` Go code.

Not merging — back to you.

amiddavid:
Rebased onto `main` (`3f272cd`) — force-pushed `4ce7c33` → `5f16ffa`, so review anchors on the old SHA are stale.

**The conflict was not in the collapse code.** It came entirely from a two-word dashboard label edit
this PR had picked up as a follow-up (`"Why declined"` → `"Why / what happened"` in `dash/ui/app.js`
and `dash/ui/index.html`). Two findings while resolving it:

- **`main` already made that change**, and went further: for the table this PR touched it *removed*
  the column altogether (`Errors | Verdict`) rather than renaming it. So the edit was redundant, and
  it is **dropped** — no `dash/` files remain in this PR. Changed files: 7 → 5.
- **The `docs/dashboard.md` half is kept, because it is not redundant.** `main`'s UI now says
  `Why / what happened` while `main`'s own docs still say **"Why declined"** — the UI change left the
  documentation stale. This PR's doc edit corrects that, and additionally explains the thing that
  made the rename necessary in the first place: the column now carries *events*, not only declines,
  and `collapse`'s `char_window` is the first one to appear there on a row that **acted**.

Verified on the rebased tree: `gofmt` clean, `go build ./...` clean, `Collapse` tests pass, full
`go test ./...` green — including `dash/TestSpendSurvivesRowEviction`, which is genuinely fixed on
current `main` (see #157: every earlier report of it, mine included, was reproduced against a tree
predating `1f1ac2b`).

No functional change from the review response; this is purely the rebase and the scope trim.

PR #172, opened 2026-09-01
fix(dashboard): re-order the Inventory tab to the reader's five questions

OsherElhadad:
The owner's verdict on this tab was *"it's too loaded and users cannot understand anything from it"* — against a build in which every feature they asked for already existed. So this adds no feature and removes none. It changes the order and the weight.

## Measured, on the seeded 200-session corpus

`seed/v2`: 16,444 requests, 69,062 declarations, 132 distinct MCP tools, 170/200 sessions captured, 17 unpriced, $247.20 of never-called tokens. Both sides measured by one browser against two proxies on identical copies, `MODEL_PRICES` set and the janitor disabled.

| | before | after |
|---|--:|--:|
| page height @1440 | 14,920 px / 16.6 screens | **8,101 px / 9.0** |
| words | 3,653 | **2,207** |
| **first named thing on the page** | **2,569 px** | **823 px** |
| **named things with figures in the first screen** | **0** | **4** |
| **prompt regions rendered at load** | **0** | **96** |
| atoms (`inv0901-value.mjs`) | 356 | 358 |
| …counting a region/part/lead row | 356 | **459** |
| px in zero-atom blocks (extended) | 2,171 | **892** |
| body `scrollWidth` @390 | **549 (+159)** | **390** |
| elements past the 390 px edge | 2,151 | **0** |

The sharpest one: the panel that merely *summarised* the prefix into five legend numbers had **968 px** while the prefix itself had **254 px and rendered nothing until clicked**. The summary was four times the size of the thing it summarised. It is deleted and its bar now lives in the panel it was summarising.

## What changed, in the reader's order

1. **The decision comes second**, right after the headline: the named groups, each worth *X across every session that carried it* and *Y per session*, in that order, with the command that removes it. A "Do this" block leads with the ordered names and one combined snippet per mechanism, built by merging the server's own `removal` documents — never by composing syntax here.
2. **Every card renders benefit, cost and a signed net as one component** (DESIGN §3.7). The cost — one prompt-prefix rebuild at cache-creation rates — previously existed only in the third paragraph of a collapsed note, so eleven cards each led with a dollar to save and none showed the dollar it costs to save it. On this corpus the net is negative over the first session for every group.
3. **"Realized by your removals" moves next to the projection.** Honesty rule 2 was right and was executed as *size* (155 px at 66 % depth against a projection in 100 pt type), which hides a claim rather than distinguishing it. Adjacency distinguishes it.
4. **The prompt panel renders all 96 regions and the system prompt's parts without a click**, grouped by owner, each share drawn, long runs folding their tail.
5. **The four re-answers fold** (flat table, per-server rollup, skills, self-removals, per-token rates), each stating its row count on the closed summary. **7,662 px → 171 px.** Nothing deleted.

## The price of ask 3, stated as a price

The user asked to see their system prompt without hunting for it. That costs **+193,776 bytes and one extra API call on every Inventory load**, because `/api/prompt` is now fetched at render rather than on a click. DOM nodes 2,900 → 3,850 and `warmSwitchMs` 65 → 116 ms. Pulling the other way, `longTasks > 50 ms` went **62 ms → 0** — folding four tables removed the big synchronous render. Assets: `tools.js` +30.3 KB raw / +10.3 KB gzipped, `tools.css` +9.2 KB / +2.7 KB, mostly comments. `/api/prompt` returns full text for every region when level 1 needs only names and weights; a `?meta=1` shape would remove nearly all of this, and that is a server change.

## Gated, deferred, and not mine

- **The per-skill opt-out switch is inert**, `SKILL_SWITCH_SAFE = false`, reason stated visibly once above the list. `internal/skills/skills.go` folds name-only listing entries into the previous surviving row, so `POST /api/toolfilter` would remove a run of skills and report one. The `skillOverrides` snippet never touches that parser and is the route the page leads with. **Depends on #168** (`fix/dash-acts-on-wrong-data-0901`).
- **The `listing_tokens` double-count workaround stays.** #168 is not in `main` as of `3697f3c` — `declTok`/`unusedTok` are still double-added at `dash/toolapi.go:573/598/617`. I own its deletion in a follow-up once #168 merges; until then the headline tile still reads ~15 % above what the bar's segments sum to, by construction and not by accident.
- **`prompt.sessions` counts rows not sessions** (`toolapi.go:561`) and three of Claude Code's own MCP tools are missing from `builtinTools`. impl-safety confirmed neither is in its scope. **These have no owner.**

## Verification

`go test ./dash/` passes; `dash/uiinventory_test.go`'s nine static checks are green and every existing `data-testid` is kept. Contrast: 0 failures over 16 pairs, both themes, 14 new pairs measured with the bundled validator (worst 3.13:1 on a non-text bar against a ≥3 gate). 20 `n/a` marks, every one with a reason in both `title` and `aria-label`; 313 `unpriced` pills where `bob/premium` has no rates; the 24 remaining `$0` cells are rows whose "read for nothing" is literally 0 — identical on both builds, a true zero. Focus: 40 stops, 0 without an indicator.

**Two instrument corrections that changed my own numbers.** `shots.mjs` waits for the nav rather than the view, so on this corpus it audits the loading skeleton — all 8 runs returned `0/3/15` off four grey bars, and its numbers are `n/a` on both sides. And its `textClip` check cannot tell a deliberate ellipsis from a broken one; with the correct predicate (`textOverflow === 'ellipsis' && el.title`) the count is **0 → 0**, so there was never a clipping defect here. I had "fixed" it and reverted that. `shortName()` — dropping the redundant `mcp__<server>__` prefix where the row already shows its server — stays on usability grounds: those characters are identical across every row of a server and the distinguishing part is at the end. Both reported to infra-harness.

### DIFF
--- a/dash/ui/tools.css
+++ b/dash/ui/tools.css
@@ -334,15 +334,6 @@ details.why > .inv-prompt-regions { max-width: none; }
   #view-tools .tiles { grid-template-columns: repeat(2, minmax(0, 1fr)); }
 }
 
-/* The identifier wraps rather than truncating. `.trunc` ends-truncates at 34ch, and an
-   `mcp__server__tool` name differs at the END — so the ellipsis eats exactly the distinguishing
-   part, on the one column that decides whether a reader can tell two removable rows apart.
-   shortName() drops the shared prefix; this catches whatever is still long. Wrapping costs a row
-   of height in a folded table and hides no characters, which end-truncation cannot say. */
-#view-tools .comp-name.trunc {
-  max-width: 34ch; overflow-wrap: anywhere; white-space: normal; text-overflow: clip;
-}
-
 /* PRE-EXISTING on origin/main, measured on the 132-MCP-tool corpus: the body scrolls sideways
    159 px at 390 because the removal snippet asserts a min-width (260 px, and 22rem = 352 px for
    the folded one) inside a card that is already inside two levels of padding. The snippet's own
--- a/dash/ui/tools.css
+++ b/dash/ui/tools.css
@@ -333,3 +333,22 @@ details.why > .inv-prompt-regions { max-width: none; }
 @media (max-width: 560px) {
   #view-tools .tiles { grid-template-columns: repeat(2, minmax(0, 1fr)); }
 }
+
+/* The identifier wraps rather than truncating. `.trunc` ends-truncates at 34ch, and an
+   `mcp__server__tool` name differs at the END — so the ellipsis eats exactly the distinguishing
+   part, on the one column that decides whether a reader can tell two removable rows apart.
+   shortName() drops the shared prefix; this catches whatever is still long. Wrapping costs a row
+   of height in a folded table and hides no characters, which end-truncation cannot say. */
+#view-tools .comp-name.trunc {
+  max-width: 34ch; overflow-wrap: anywhere; white-space: normal; text-overflow: clip;
+}
+
+/* PRE-EXISTING on origin/main, measured on the 132-MCP-tool corpus: the body scrolls sideways
+   159 px at 390 because the removal snippet asserts a min-width (260 px, and 22rem = 352 px for
+   the folded one) inside a card that is already inside two levels of padding. The snippet's own
+   <pre> scrolls internally, so the min-width buys nothing on a phone and costs the whole page its
+   horizontal lock. Zero here; the desktop values are untouched. */
+@media (max-width: 560px) {
+  .inv-removal, .inv-removal-fold .inv-removal { min-width: 0; }
+  .inv-removal-fold > summary { white-space: normal; }
+}
--- a/dash/ui/tools.css
+++ b/dash/ui/tools.css
@@ -334,12 +334,3 @@ details.why > .inv-prompt-regions { max-width: none; }
   #view-tools .tiles { grid-template-columns: repeat(2, minmax(0, 1fr)); }
 }
 
-/* PRE-EXISTING on origin/main, measured on the 132-MCP-tool corpus: the body scrolls sideways
-   159 px at 390 because the removal snippet asserts a min-width (260 px, and 22rem = 352 px for
-   the folded one) inside a card that is already inside two levels of padding. The snippet's own
-   <pre> scrolls internally, so the min-width buys nothing on a phone and costs the whole page its
-   horizontal lock. Zero here; the desktop values are untouched. */
-@media (max-width: 560px) {
-  .inv-removal, .inv-removal-fold .inv-removal { min-width: 0; }
-  .inv-removal-fold > summary { white-space: normal; }
-}
--- a/dash/ui/tools.css
+++ b/dash/ui/tools.css
@@ -201,3 +201,135 @@ details.why > .inv-prompt-regions { max-width: none; }
 /* The per-row removal fold sits in a table cell, so it must not stretch the row. */
 .inv-removal-fold > summary { white-space: nowrap; }
 .inv-removal-fold .inv-removal { min-width: 22rem; }
+
+/* ── the call to action ─────────────────────────────────────────────────────
+ * The one thing to DO, at the top of the list rather than inferred from eleven cards. It gets
+ * the raised surface and the accent edge because it is the only block on the page that is an
+ * instruction rather than a measurement — and it is deliberately NOT a button: the action is a
+ * config edit the reader makes in their own file, so the affordance is a snippet they can read
+ * before they run it. A dashboard that offers a one-click "fix everything" for a change that
+ * rebuilds a cached prompt prefix is offering a click nobody can review.
+ */
+.inv-cta {
+  display: grid; gap: var(--sp-2); margin-top: var(--sp-3);
+  padding: var(--sp-3); border-radius: var(--r);
+  background: var(--accent-soft); border-left: 3px solid var(--accent);
+}
+.inv-cta-head { display: flex; flex-wrap: wrap; gap: var(--sp-1) var(--sp-3); align-items: baseline; }
+.inv-cta-head strong { font-size: var(--t-h3); }
+/* The snippet keeps the raised surface it has everywhere else on the page, so it reads as the
+   same component here as in a group card rather than as a differently-coloured one. */
+.inv-cta .inv-copy { background: var(--bg-raised); }
+
+/* ── benefit, cost, net ─────────────────────────────────────────────────────
+ * DESIGN §3.7 as one three-part grid. It is a grid rather than three lines of prose so the three
+ * figures share a column and can be compared by position, which is the whole reason the cost has
+ * to sit beside the benefit instead of below it. The sign is in the text, never in the colour
+ * alone (§2.5), so this survives greyscale and a printout.
+ */
+.inv-pb-label { font-size: var(--t-small); font-weight: 500; color: var(--fg-muted); }
+.inv-pb-val {
+  font-family: var(--mono); font-variant-numeric: tabular-nums; font-size: var(--t-body);
+  color: var(--fg);
+}
+.inv-pb-val.good { color: var(--good); }
+.inv-pb-val.bad  { color: var(--bad); }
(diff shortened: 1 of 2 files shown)


OsherElhadad:
## Update — two corrections since opening, both in the honesty path

**1. `totals.priced` was the wrong gate, and this change had made it worse.** (`817df6b`)

That flag is false whenever *any* part of the population is unpriced, so on `seed/v2` — 17 unpriced sessions of 200 — it is false in steady state and never flips, while `totals.unused_usd` is a correct **$247.20**. Two tiles gated their dollar on it, so the headline rendered `645,662,264 tok / "no dollar: some models here are unpriced"`. A number that is right, labelled unknown, is its own kind of dishonesty, and it was worse here than before this PR because I had moved that figure into the first screen.

Both now gate on the **priced population** — `coverage.sessions − coverage.unpriced_sessions`, both already shipped — and name it:

> **$247.20** — if none of it had been carried, over the 183 of 200 sessions that ran on a priced model

Per-row and per-group dollars are untouched: they consult their own row's `priced`, which is the right flag at that scale, and still render `unpriced` through `money()` — 313 of them here.

The same change fixed a second one in the opposite direction: the skills-listing cost tile said `unpriced` where the honest answer is **`$0`**. `unused_listing_reads` is 0 because every session carrying a listing invoked at least one skill, so the listing was never waste. It now agrees with the "Listing read for nothing: 0" tile beside it instead of claiming ignorance we don't have.

**2. The 390 px constraint is closed — confirmed by `shots.mjs` itself**, now that it has a readiness gate, on a rendered view in all four theme × viewport combinations:

| | `clippedOff` | `reachOver` | `bodyScrollsX` | `textClip` | nodes |
|---|--:|--:|---|--:|--:|
| `origin/main` @390 | **48** | **4** | **YES +159 px** | 0 | 1,547 |
| this branch @390 | **0** | **0** | **no** | 0 | 342 |

`textClip` is **0 on both**, which independently confirms the `.trunc` ellipses were never a defect — I had "fixed" them and reverted that in `8e898a9`. `shortName()` stays on usability grounds: the `mcp__<server>__` prefix is identical across every row of a server and the distinguishing part is at the end, so two rows of one server read as two identical strings until you hover each.

**Provenance of the px/atom table**, since the readiness bug invalidated other numbers on this sweep: it came from `inv0901-geom.mjs` / `inv0901-value.mjs`, re-copied from source at 13:50, and both gate on `#view-tools [data-testid="inv-gauge"]` — created by `gauge()` inside `renderHeadline`, which runs only after `tools.report` is populated. So they wait on **content**, not on the nav, which is what made `shots.mjs` measure a skeleton. The totals also match an independent probe gating on `inv-coverage`: two different content gates, 14,920 → 8,101 px both times. Re-taken on the final build and unchanged.

`go test ./dash/` passes. The `listing_tokens` workaround and `SKILL_SWITCH_SAFE = false` both still stand — #168 is not in `main` as of `3697f3c`.

OsherElhadad:
## Update 2 — a retraction, two cherry-picks, two deletions

**I retract the `549 → 393 px` attribution from my first update.** Those two numbers came from two different builds — `origin/main` and this branch — and I credited the delta to the CSS rule I had just added instead of isolating it. Corrected in `bc37b80`.

Isolated properly (injecting each rule's pre-fix value back with `!important`, one page, both states):

```
default state,  all three rules disabled:  390 px, 0 offenders
expanded state, all three rules disabled:  397 px, same 11 offenders as with them enabled
```

What actually locks the default state at 390 is **the folding** — the 1,221 px flat table is no longer rendered at load — **plus `compact()`** on the unpriced tile value. My two CSS rules were no-ops, and so are `impl-perf`'s three on this branch.

**So: two deletions and two cherry-picks.**

- **Deleted** my `.inv-removal { min-width: 0 }` — `impl-perf` measured it a no-op on its corpus, I measured it a no-op on mine, so it is a no-op on both — and my `white-space: normal`, a duplicate of `impl-perf`'s rule without its source-order comment.
- **Cherry-picked** `4da905f` (← `9fc2835`) and `e1c3b34` (← `65e7b6d`) from #169, comments intact. Its three `minmax(0, 1fr)` tracks are load-bearing at 159 px on the **pre-fold** layout, which is what `main` still is, and `minmax(0, 1fr)` is strictly more correct than a bare `1fr` for a track that can hold an unbreakable tool name. Kept as a judgment call, not a measurement — flagged as such to `impl-perf`, and I will drop them if it would rather own them in #169. One of us should, not both.

**Final, `shots.mjs` run directly, all 17 tabs, both themes, both viewports, 342 text nodes:**

| `tools` @ | `clippedOff` | `reachOver` | `bodyScrollsX` | `textClip` | contrast | focus |
|---|--:|--:|---|--:|---|---|
| 1440 dark/light | 0 | 0 | no | 0 | 0/17 | 0 missing |
| 390 dark/light | 0 | 0 | **no** | 0 | 0/17 | 0 missing |

Two tabs still fail the constraint and are not this PR's: `keepalive` **+604 px**, `kvcache` **+16 px**.

**One residual documented rather than fixed:** with all 815 disclosures forced open — a state no reader reaches by clicking — the body reads **397 px** with 11 offenders rooted at `div.inv-group`. Not the tile, the snippet, or the summary. The default state is 390 with zero offenders, which is the state the constraint is about, so I am leaving it named rather than chased.

px/atom table re-taken on this build and unchanged: **14,741 → 7,936 px, 3,653 → 2,219 words, 356 → 358 atoms** (`inv0901-value.mjs`); **14,920 → 8,101 px** (`inv0901-geom.mjs`). `go test ./dash/` passes.

OsherElhadad:
## ⚠️ Update 3 — a retraction, and a merge-order dependency

**This PR must not merge before #169.** On its own it reports `clippedOff 48` and `bodyScrollsX YES +159 px` at 390 px on the Inventory tab, which violates the BRIEF's hard constraint. With #169 in, all four theme × viewport combinations report `0 / 0 / no`. Recorded in the branch history as `a8eb181`, not just here.

**I retract the "no-op" claim from Update 2.** I reported that `impl-perf`'s three `minmax(0, 1fr)` tracks were no-ops on this branch, on the strength of an isolation test that injected each rule's pre-fix value back with `!important`. That test is unsound. Rebuilding without the commits gives the opposite answer:

```
with 9fc2835 + 65e7b6d      tools @390:  clippedOff  0, reachOver 0, bodyScrollsX no
without them                tools @390:  clippedOff 48, reachOver 4, bodyScrollsX YES +159 px
```

`impl-perf`'s original measurement stands unamended — the folding does not mask its fix. The likely mechanism is that I injected `grid-template-columns: auto !important` to stand in for a rule that originally had *no* `grid-template-columns`, and an explicit `auto` track is not equivalent to the implicit one it replaced. **Simulating a revert is not reverting it.** Its two-builds method was the sound one.

**The cherry-picks are still dropped**, for the better reason: ownership. #169 is where those rules are measured against `main`, one rule wants one owner, and they are `impl-perf`'s to keep. That turns a masked dependency into a declared one, which is the outcome I want.

**My own two rules stay deleted** (`aa3c656`). Those genuinely were no-ops — measured independently by both of us — and dropping them is not what broke this.

**Struck from the residual list:** `keepalive +604 px` and `kvcache +16 px` are not this branch's. Verified by fingerprint rather than on trust — `grep -c scroll-x dash/ui/index.html` → 1 and `grep -c 'grid-2 > \*' dash/ui/style.css` → 0, i.e. this branch is missing `f3992e4`, which fixes both.

**Caveat on one figure I reported earlier:** the "397 px with 11 offenders in the 815-forced-open state" came from the same probe family as the unsound test above. Treat it as indicative, not measured. It is not a state a reader reaches by clicking, and it stays documented rather than chased.

Unchanged and re-verified on this build: **14,741 → 7,936 px, 3,653 → 2,219 words, 356 → 358 atoms** (`inv0901-value.mjs`), **14,920 → 8,101 px** (`inv0901-geom.mjs`), `textClip 0` both sides, `contrast 0/17`, `focusMissing 0`, `go test ./dash/` passing.

PR #178, opened 2026-09-02
fix(metrics): scope extraction accounting per component, and log every extraction call

### LINKED ISSUE #176: fix(metrics): /stats credits extract_llm with 101 calls while its own log shows cands=0 on every request
## Symptom

On a 45-run benchmark (iteration 023, 3 arms × 15 tasks), `/stats` and `extract_llm`'s own debug log disagree about whether the component did anything, and they cannot both be right.

For the same arm, same run:

| source | says |
|---|---|
| `/stats` `components.extract_llm` | `runs: 389`, **`acted: 239`**, `saved_tokens: 6,077,421` |
| `/stats` `extract` nested map | **`calls: 101`**, `avg_latency_ms: 59,009`, `extraction_cost_usd: 3.129`, `net_value_usd: −1.162` |
| `cg.extract_llm` debug records | **`cands: 0` on all 374 records**, `skip_tail: 7,487`, `skip_floor: 692` |

`cands` is `len(cands)` logged after the gate loop (`components/offload/extract_llm.go:1101-1104`), so `cands: 0` plausibly means *no candidate ever survived the gates* — which is consistent with `skip_tail` accounting for 7,487 of them. But a component with zero surviving candidates cannot make 101 model calls or act 239 times.

Across the run, `cands` was **0 in every one of 692 records** in the two arms checked (318 and 374), with no exceptions.

## The likely cause, and why it matters

The `extract` nested map is not obviously scoped to one component. `extract_llm` and `extract_llm_sweep` are both extraction paths and both make model calls, and the sweep in this arm made **96 asks** — close to the 101 `calls` reported. If the nested map aggregates both, then:

- the 59-second `avg_latency_ms` and the **−$1.162 `net_value_usd`** are the *sweep's* figures, not the tail pass's;
- `components.extract_llm.acted: 239` may likewise be counting the sweep's frozen-decision replays, which the same snapshot records as `reapplied_same_session: 2,291` / `calls_avoided: 2,291`.

This is not a cosmetic labelling problem. Those counters were used to attribute a 5,452 ms/request latency cost and a negative net value to a specific component, and an experiment's conclusions were drawn from that attribution before the contradiction was noticed. Any per-component cost or latency claim from a pipeline containing both components is currently unsafe.

`saved_tokens: 6,077,421` against `saved_tokens_unique: 597,764` — an `overcount_ratio` of **31.58** — is a further sign the accounting is counting replays repeatedly.

## What a fix needs to establish

1. Whether `extract` (the nested map) is scoped to `extract_llm`, to `extract_llm_sweep`, or to both. If both, split it — the two have different economics and one of them is priced per-output on a cheap model while the other is one call on the request's own model.
2. Whether `components.extract_llm.acted` counts frozen replays as acts. If it does, replays and fresh decisions want separate counters; a replay costs nothing and an act costs a call.
3. Why `cands` is 0 while the component is credited with calls — one of the two is wrong, and which one decides whether the tail pass ran at all on this workload.

## Reproduction

Any pipeline containing both `extract_llm` and `extract_llm_sweep` with traffic that makes the sweep fire — e.g. `deploy/harbor/cfg-iter023-B-merged.yaml` on this repo's branch `feat/coref-recut`. Compare `/stats` `components.extract_llm.acted` and `extract.calls` against `cands` in the `cg.extract_llm` debug records for the same run.

Found while tracing iteration 023; see `docs/experiments/loca/iter023/results.md`. Filed separately from that branch's work because it is a defect in `main` and unrelated to the co-reference mechanism being tested.

### LINKED ISSUE #177: fix(extract_llm): no log record when an extraction call is made, so its 59s latency and negative net value have no per-request trace
## Symptom

There is no log record when `extract_llm` makes a model call. Across a 45-run benchmark the only message the component emits is `cg.extract_llm`, one per request:

```
extract_llm msg types = {'cg.extract_llm': 318}   # arm A
extract_llm msg types = {'cg.extract_llm': 374}   # arm B
```

That record carries `tools`, `cands`, `reapplied`, `skip_placeholder`, `skip_tail`, `skip_floor`, `big_but_not_tail`, `floor`, `nInput`, `maxCachedIdx`, `cacheAware`, `max_output_tokens` — everything about the *decision*, and nothing about the *call*: no per-call latency, no token counts, no cost, no accept/reject outcome, no candidate identity.

## Why it matters

The call is the expensive part. On the run above, `/stats` attributes to this component 101 calls at **59,009 ms average latency** and a **net value of −$1.162** — the single most expensive behaviour in the pipeline, and the only one with no per-request trace. Consequences observed while investigating:

- The 101 calls could not be tied to the requests that made them, so it was impossible to test why they happened in one arm and not the other two. The investigation stopped there.
- A 59-second average could be 101 calls at 59s or a handful of multi-minute outliers dragging the mean. Nothing in the logs distinguishes those, and they call for different fixes.
- The negative net value cannot be attributed to particular candidates, so there is no way to see *which* extractions lost money — which is exactly what #120's break-even argument needs in order to be checked against a second workload.

`extract_llm_sweep` already logs its call: `cg.sweep.ask` carries `offered`, `verdicts`, `dropped`, `candidate_tokens`, `removed_tokens`, `cache_read`, `fresh`, and (recently) `session`, `max_cached_idx`, `req_tokens`, `messages`. A run's sweep economics are fully reconstructable from the logs; the tail pass's are not reconstructable at all.

## Suggested shape

One DEBUG record per extraction call, at the call site, carrying at minimum: `session`, the candidate's content key, its size in tokens, the model used, latency, input/output tokens, whether the result was accepted or rejected by the never-worse check, and the resulting saving. That is enough to compute per-call net value and to answer "which calls, on which requests, and did they pay."

Guard it behind `logging.Debugging()` if the payload costs anything to build, per this repo's stated rule.

Found while tracing iteration 023; see `docs/experiments/loca/iter023/results.md`. Related to the attribution conflict filed alongside this — that issue is about counters disagreeing, this one is about there being no record to arbitrate them with. Filed separately from the co-reference branch because it is a `main` defect and unrelated to the mechanism being tested there.

amiddavid:
Closes #176. Closes #177.

Both issues are one subsystem and one fix: #176 is counters disagreeing, #177 is that there was no record to arbitrate them with.

> **Round 2 (`e95e90c`)** addresses all four findings from @amiddavid's first review. **Round 3 (`6e01798`)** fixes the last remaining gap in-PR and files the rest, so every open item has a number — #179, #180, #181, #182. **Round 4 (`7fed0d0`)** fixes the re-review's ledger-fold finding, including the vanishing row at the no-asker site. See [Review round 2](#review-round-2) onward.

## What the counters were actually scoped to

Established from the code that writes them, not from the names.

**The `extract` nested map was process-global.** `metrics/extract.go` held one set of package-level atomics — `xCalls`, `xLatencyMs`, `xCacheHits`, `xSuppressed`, `xGrossSaved`, `xLookups`, `xValueNano`, `xReasons` — and *both* `components/offload/extract_llm.go` and `components/offload/extract_sweep.go` wrote every one of them. So `calls`, `avg_latency_ms`, `calls_avoided`, `calls_suppressed`, `gross_saved_tokens`, `reasons` and the entire net-value block were the **sum of extract_llm and extract_llm_sweep**, presented under a name that reads as one of them.

The file said so as a design statement ("These are process-global counters, matching cheapmodel.Usage's existing scope"), and `internal/cheapmodel/usage.go` states the premise it rested on: *"per-component attribution would need the Model interface to carry a label, a deferred refinement — today the LLM component in a config is extract, so the global total is that component's cost."* That premise stopped being true when the cold-transcript sweep became its own component, and nothing re-examined the comment. #176's hypothesis was right.

**`extraction_cost_usd` was scoped wider still — not to extraction at all.** The proxy derived it from `cheapmodel.Usage()` / `CacheUsage()`, the process-global token totals, priced through one rate card. Every cheap-model call in the process lands in those totals, including `summarize` and `agentdiet` (both call `model.Complete`). The same figure also *mispriced* the half of extraction it did cover: the sweep's asks go to the request's own frontier model while the rate card is haiku's. So `−$1.162` was two components' savings against three components' token spend, priced at one component's rates.

**`components.extract_llm.acted` does count frozen replays as acts.** It is correctly scoped to the component (`metrics.Aggregator.Component` keys on `Report.Component`), and it is `Saved() > 0 && !Reverted && !Skipped`. A same-session replay splices previously frozen bytes and therefore saves tokens, landing in the same counter as the call that derived them. `acted: 239` beside `reapplied_same_session: 2,291` was read as 239 paid extractions.

**`cands: 0` was not the wrong side.** A replay `continue`s before its candidate is appended to `cands`, so `cands: 0` with a nonzero `reapplied` is exactly what a request that only replayed looks like. `cands: 0` on all 692 records and `acted: 239` are consistent and *both true*: extract_llm acted 239 times without making a single fresh call in that arm. The number that was wrong is `calls`, and it was wrong because it was pooled — the 101 are very nearly the sweep's 96 asks, and the 59,009 ms mean is the sweep's frontier-model ask latency.

**`saved_tokens` 6,077,421 vs `saved_tokens_unique` 597,764 is not a defect.** `Saved` is documented as cumulative-per-turn and `SavedUnique` dedups by content key; `overcount_ratio: 31.58` is the designed signal that this workload re-sends history verbatim, and `reapplied_same_session: 2,291` is its cause. Deliberately left alone — changing it would break the one honest pair /stats already had.

## What changed

- **`metrics/extract.go`**: the counters are keyed by component. Every recorder takes the name; the aggregate is *derived* by summing rather than maintained alongside, because two totals kept in parallel is how the second drifts. `ExtractStats` gains `by_component`; the enclosing block keeps its keys and its meaning as the sum.
- **`RecordExtractionSpend`**: each component records what its own call cost, priced with the rates of the model it actually addressed (`ModelCall.CostUSD`, which both components already computed and neither reported to /stats). Per-component net value is now that component's own arithmetic.
- **The latency accessors are keyed too, and that is a behaviour change.** `offload.tooSlowToExplore` read the *global* p50, so the sweep's ~59-second frontier asks braked extract_llm's cheap-model exploration on evidence from a different component and a different model. Called out because it is a decision path, not a display.
- **`components.Report` gains `Replays`**, set through `Report.Replay(name)` so the count and the descriptive Event (`reapplied_same_session` / `reapplied_cross_session`) cannot drift apart when a third replay path is added. `metrics` splits `acted` into `acted_fresh` and `acted_replay` via one `compStat.act` called from both the enforced and observe loops. `acted` keeps its name, its meaning and its value; a deterministic component reports all-fresh.
- **`cg.extract_llm.call`** (#177): one DEBUG record per call at the call site, carrying `session`, `content_key`, `candidate_tokens`, `model`, `latency_ms`, `input_tokens`, `output_tokens`, `cache_read`, `cache_write`, `cost_usd`, `accepted`, `saved_tokens`, `rejection`, `gate`, `strategy`, `timed_out`. Guarded by the `dbg` flag already resolved once per request. `accepted` is read off the same `ModelCall` the splice used, so the log cannot claim an acceptance the request did not get.

## Vacuity — round 1

Six single-point mutations, applied on the eval box, test run, mutation restored. **Five killed, one survived.**

| # | Mutation | Result |
|---|---|---|
| M1 | `xFor` returns one shared bucket for every name — i.e. **introducing** the original bug, since the fix makes it unrepresentable | **FAILED.** `TestExtractStatsAreScopedPerComponent` reports "extract_llm credited with 2 calls; it made none", "avg_latency_ms = 59009", "net = −1.162"; `TestLatencyAccessorsAreScopedPerComponent` fails; the rendered /stats test fails with "credits extract_llm with 1 calls (baseline 0)" and "charged $1.2 it did not spend" |
| M2 | spend always taken from the host's process-global figure | **FAILED.** `extraction_cost_usd = 9.99, want 0.32` |
| M3 | `ByComponent` never attached to the snapshot | **FAILED** in both the JSON test and the rendered /stats body |
| M4 | `act()` ignores `Replays` | **FAILED.** `acted_fresh 4 / acted_replay 0` against 1/3, in the package test *and* in the rendered /stats body |
| M5 | the per-call record neutralized (`if dbg` → `if false`) | **FAILED.** "got 0 cg.extract_llm.call records for 1 reported calls" |
| M6 | `Debug(` → `Info(` on the per-call record | **STILL PASSED — see below** |

M1 is the **equivalent-mutant case**: keying the counters makes the pooled bug unrepresentable, so reverting proves nothing and the bug had to be *introduced* instead. Stated because that is the only honest way to read that row.

M6 survived, and the reason is a genuinely redundant defence rather than a weak test: the property "no record at INFO" is held by *both* the `if dbg` guard (which exists so the payload is not built) and the slog level on `Debug()` (which exists so it is not printed). Remove either and the other still holds. The **combined** mutant — `if dbg` → `if true` *and* `Debug` → `Info` — does fail it (verified; the INFO line appears in the buffer). `TestExtractLLMCallRecordIsDebugGated` therefore pins the outcome, not either mechanism, and that is written into the test's own doc comment so nobody reads its passing as evidence the guard is present. `TestExtractLLMLogsOneRecordPerCall` (M5) is the one that dies when the record itself goes away.

Both new tests in `components/offload` also assert their **preconditions** before their subject (`model.calls != 0`, `len(rep.Calls) != 0`, and the fixture logger actually being at DEBUG), because an extract_llm fixture that silently declines is how this kind of test passes vacuously — `extract_llm_timeout_test.go` carries the same warning from the last time it happened.

Assertions are on **rendered output** wherever the handler could drop a field: `proxy/extract_attribution_test.go` drives `h.stats(w, req)` and unmarshals the body, and `metrics.TestByComponentSurvivesJSON` goes through `json.Marshal`. The proxy test states its assertions against a **baseline read** of the counters rather than against zero, since they are process-global by construction.

---

### 1. MAJOR (introduced by this PR) — the sweep's fallback-ask spend and latency reported as $0

Correct, and the mechanism is exactly as described. `fallbackAsk` returned only `(string, error)`; `r.rec` was built from the *prefix ask's* usage and assigned once; `RecordExtractionCall` ran before two of the three fallback points. So on all three live paths — no prefix asker (any non-Anthropic route, where the fallback is the **only** call), `ErrNoPrefix` (every session's first turn), and `usage.CacheRead == 0` (a mistimed window) — a real frontier-model call on a full sampled transcript was booked at $0.00, and on the latter two its seconds were dropped from `avg_latency_ms` too. And the reviewer is right that this PR is what made it visible: before, the tokens still reached /stats through cheapmodel's process totals via the `cost` argument, and making components price their own spend removed that accident.

Fixed:

- `fallbackAsk` returns `(string, components.PrefixUsage, float64, error)`. `model.Complete` reports no usage, so it is read from a per-call `cheapmodel.WithCallSink` nested inside whatever scope already wraps ctx — the same construction `extract_llm` uses, and it still reaches every ancestor so the request's own bill keeps the tokens. The sink is read **whether or not the call errored**: a completion that failed after the provider billed it still cost money.
- Each **leg** books itself through one `recordLeg` closure, reached from all three fallback sites via `runFallback`, so a fourth site cannot be added without the accounting coming with it. Two model calls now count as two.
- `foldFallback` grows the ledger row to cover both legs (tokens, dollars, wall time) and names the strategy for what actually ran — `"fallback"` on the no-asker route, `"prefix_ask+fallback"` otherwise, rather than claiming a prefix-ask leg that never happened.
- One deliberate exception: an `ErrNoPrefix` refusal is **not** booked as a call. It is refused locally, before any request leaves the process, so counting it would inflate `calls` with work that by definition did not happen and put a 0 ms sample into the mean the exploration brake reads.

### 2. MAJOR (introduced by this PR) — the host-fallback cost was aggregate-only

Also correct, on both halves, and `TestRecordedSpendBeatsTheHostsGlobalFigure` did pin the bad shape. The host's `cost` cannot be split across rows — it is one process-global number that also carries `summarize` and `agentdiet` — and splitting it by call count would be an invented number published at the same precision as a measured one. So the fix is to make the unknown *visible* rather than to guess:

- **`net_value_usd` is now `*float64`**, rendered as JSON `null` when the spend behind it is not known. A row that made calls and priced none of them no longer publishes `+grossValue` on no cost evidence. The **aggregate is never null** — it always has a determined spend — so `/metrics` is unaffected in practice, though `promexport` now reads it through a `Net() (float64, bool)` accessor and omits the series rather than publishing an unknown as `0`, which on that gauge reads as exactly break-even.
- **`cost_source`** names the case, because `$0 spent` and `$0 of evidence` are the same number and the opposite claim: `component` (every call priced itself — trust it), `host_total` (the host's superset figure), `partial` (some calls unpriced, so the total is a **floor**), `unpriced` (this row made calls and priced none — nothing is known), `none` (no calls; `0` is true).
- **Partial recording no longer under-reports.** `unpricedCalls` replaces the single `anySpend` boolean: when any calls are unpriced the aggregate publishes the larger of the recorded floor and the host's superset figure, and labels which one it chose. A total that quietly omits real dollars is the defect, not the loud one.

### 3. The docs described the old scoping

- `docs/reference/routes.md` — `acted_fresh` / `acted_replay` added to the per-component table, with `acted` now flagged as including free replays; a new **Extraction economics** section documents `by_component`, `cost_source` and the nullable net, and carries a warning that the enclosing block is a sum.
- `docs/components/extract_llm.md` — the table that presented `calls`, `extraction_cost_usd`, `avg_latency_ms` and `gross_saved_tokens` as this component's own figures now opens with a danger admonition stating they are sums across components, with the measured numbers (101 calls / 59,009 ms / −$1.162 against `cands: 0` on 374 requests) as the worked example, and directs every per-component claim to `by_component`. The `acted`-counts-replays note is there too.

### 4. The false claim in this body

Verified and removed. Grepping `deploy/` for `gross_saved_tokens`, `calls_avoided`, `extraction_cost_usd` and `avg_latency_ms` returns nothing — no harbor script reads any key in this block. What harbor *does* read is `runs`, `acted` and `saved_tokens` off the per-**component** objects, and `measure.py` reads them by direct index (`v["acted"]`), so removing one is a `KeyError` rather than a silent zero. That part of the rationale was sound; it was attached to the wrong table.

The real reason the extract block's names are frozen is `/metrics`: `proxy/promexport.go` publishes them as `cg_extract_calls_total`, `cg_extract_cost_usd`, `cg_extract_net_value_usd` and `cg_extract_latency_ms`, and `dash/metrics_export.go` documents the last as the endpoint's only dollar figure. Those are queried by off-repo, unversioned alert rules and dashboards, where a rename breaks monitoring *silently* — a worse failure than breaking the harness, which would at least crash. Corrected in `metrics/extract.go`, `docs/reference/routes.md` and `docs/components/extract_llm.md`, with the distinction between the two consumers stated rather than blurred.

### The dashboard finding → #179

`dash/event.go:527` does keep the exact reading #176 describes (`row.Acted = row.Mutated && row.SavedGross > 0`, with `r.Replays` available and unread), and `/api/components`' `acted`, `acted_tokens`, `act_rate` and `act_rate_structural` all inherit it. Pre-existing on `main` rather than introduced here, so per the standing rule it is filed as its own issue with the evidence — **#179** — and left out of this PR so it can ship without waiting. It is in the known gaps below. The `act_rate_structural` derivation (`mutated − acted`) will want re-checking against whatever shape the split takes there.

### Vacuity — round 2

Six new single-point mutations. **All six killed** — but one only after the tests were strengthened, and that is worth reading.

| # | Mutation | Result |
|---|---|---|
| M8 | `fallbackAsk` reports no usage | **FAILED.** Ledger row prices the fallback at $0; row tokens 0/0 against 31000/420; `/stats` delta $0; `cost_source` `"unpriced"` where `"component"` is due |
| M9 | the fallback books no leg | **FAILED.** 0 calls recorded where 1 is due; 1 where 2 are due |
| M10 | the ledger fold removed | **FAILED.** Row shows prompt 40 / completion 90 of 31040/510; `Strategy` `"prefix_ask"` where `"fallback"` is due |
| M11 | an unpriced row claims a known net | **FAILED.** `net_value_usd = 0.05` where `null` is due |
| M12 | back to the `anySpend` boolean | **FAILED.** Total $0.02 against a host figure of $4.00, and `"component"` where `"partial"` is due |
| M13 | `cost_source` tagged `json:"-"` | **SURVIVED at first**, then **FAILED** after the fix — see below |

**M13 is the one to read.** It survived the first run, because the `cost_source` assertions I had written read the Go struct field rather than the encoded payload — precisely the "a field can be computed correctly and dropped by the handler" failure this repo's rendered-output rule exists for, committed in the same change that cites the rule. `TestByComponentSurvivesJSON` and `TestStatsRendersPerComponentExtractionAttribution` now assert `cost_source` and the `null` net **on the marshalled payload and on the `/stats` body**, and M13 then fails on both surfaces (`the encoded by_component row is missing "cost_source"`, `/stats serves no extract.cost_source`). Recorded here rather than quietly fixed, because a mutant that survives and then dies after a test change is the only evidence that the test change did anything.

M6's redundant-guard situation is unchanged and still documented in the test.

### Round 3 — every remaining gap is now either fixed here or tracked

Per the rule that anything doable inside this PR is done inside it and everything else gets a number, the four items left as prose gaps after round 2 are resolved. Commit `6e01798`.

**Fixed here: `cost_source: partial` now names what the total is short of.** The label said the total was a **floor** without saying what it was missing, so finding the culprit meant scanning every `by_component` row for `cost_source: unpriced` — re-deriving an answer the snapshot had already computed, since it is the same test that sets the aggregate's label. A reader who has to re-derive it will read the label and stop, and then the floor gets quoted as the bill. `extract.unpriced_components` names them, sorted, on the aggregate only, omitted when every call priced itself. Published under **both** incomplete regimes: under `partial` it names what the floor is short of, under `host_total` it names what forced the fallback. Documented in `routes.md` and `extract_llm.md`.

**Tracked, with the reasoning in the issue rather than only here:**

| Issue | What | Why not here |
|---|---|---|
| **#179** | `/api/components` counts free frozen replays as acts (`dash/event.go:527`), so `acted` / `acted_tokens` / `act_rate` keep the exact conflation #176 reports — on the surface an operator actually looks at | Pre-existing on `main`, unrelated to what this PR tests. Own issue, own PR, so it can ship without waiting |
| **#180** | `cg_extract_*` has no `component` label — and after round 2 it *inherits* the `cost_source` regime without being able to express it, so a gauge reading `−0.71` may be priced arithmetic, a **floor**, or the host's superset figure, identically | A label changes the shape of series that off-repo, unversioned alert rules are written against. That is a monitoring break needing its own rollout, not a code change |
| **#181** | `cheapmodel.Usage()` is still process-global, and its comment still asserts the premise that produced #176 ("today the LLM component in a config is extract") | The "Model interface carries a label" refinement the comment itself defers — a wider refactor. The issue records *why deferring is safe*: /stats depends on it only as a **labelled** fallback (`cost_source: host_total`, documented as a superset), so the wrongness is declared at the point of use rather than silently inherited |
| **#182** | Whether a locally-refused `ErrNoPrefix` ask should count as a call. **Assigned to @OsherElhadad** | A judgement call that silently steps down a published counter, so it should not rest on one person's read |

On **#182**: I took the correct accounting — keep it uncounted. An `ErrNoPrefix` is refused inside `proxy/prefixask.go` before any request leaves the process, so it is not a call; it contributes a ~0 ms sample to the p50 the exploration brake reads, and it inflates the denominator of `avg_latency_ms` and cost-per-call. A counter named `calls` that includes calls never made is how #176 happened. But the continuity cost is real and named in the issue: `extract.calls` and `cg_extract_calls_total{outcome="made"}` will read lower for the same traffic, roughly one per session, and "the component stopped working" and "the counter got honest" look identical from outside. The issue gives the concrete alternative — count it under a distinct name like `refusals`, which I think is *better* than what I shipped and which I skipped only to avoid widening this PR further — and says explicitly what would change my mind.

**Dropped from the gaps:** `docs/experiments/loca/iter023/results.md` lives on `feat/coref-recut`, not `main`, so this PR cannot touch it; @amiddavid is correcting it. The general point stands below without implying a fix was owed here.

### Vacuity — round 3

Three mutations on the new field, **all three killed, each on both surfaces**:

| # | Mutation | Result |
|---|---|---|
| M14 | the list is never populated (revert to `partial` naming nothing) | **FAILED** in both — "the rendered block does not name the unpriced component, so `partial` still leaves the reader to scan every row"; "/stats serves an incomplete total with no unpriced_components" |
| M15 | computed but never attached to the snapshot | **FAILED** in both |
| M16 | dropped by the encoder (`json:"-"`) | **FAILED** in both |

M16 is the M13 shape, and it is why both assertions read the **marshalled payload** rather than the Go field: `TestTheAggregateNamesWhichComponentIsUnpriced` goes through `json.Marshal`, `TestStatsNamesTheUnpricedComponent` reads the `/stats` HTTP body. Both assert the **precondition** that the total is actually incomplete before asserting the list, so neither can pass vacuously on a snapshot with nothing to report, and both assert the field is **omitted** when everything priced itself — an empty list on a complete total would read as a warning where there is none.

The proxy test uses a component name of its own rather than a real one: these counters are process-global for the package run, and `TestStatsRendersPerComponentExtractionAttribution` asserts *absolute* values on `extract_llm` and `extract_llm_sweep`. Making either unpriced there would break that test through shared state for a reason unrelated to either test's subject.

Running total across three rounds: **15 mutations, 15 killed** — two of them (M6, M13) only after being reported as survivors first.

### Round 4 — a failed fallback's wall time never reached the ledger row

The one item from re-review, fixed in `7fed0d0`. Non-blocking as filed, and correct as diagnosed.

All three fallback sites are `if reply, err = runFallback(); err != nil { return nil, r }`, and every `foldFallback()` call sat *after* those early returns — so a fallback that failed was never folded. `runFallback` has already booked the leg into metrics via `recordLeg`, so `/stats` counted the call and its seconds while `r.rec` kept only the prefix ask's `LatencyMs` and a `Strategy` naming a leg that was no longer the only one that ran.

**Latency, not dollars**, and your reasoning for that holds: on an error the sink is empty, because `recordUsageCache` is reached only after a successful decode on both backends and neither errors after billing. Same class as the round-2 finding, not a repeat of its severity — but the fallback is the slow leg by construction, so a failed one puts tens of seconds into `avg_latency_ms` against a per-call row showing milliseconds.

**Site 681's vanishing row is covered.** `foldFallback` now establishes the identity fields — `Component`, `Model`, `CandidateTokens`, `GateReason` — when the row does not exist yet. Without that, deferring the fold alone would have populated latency and tokens on a row whose `Component` was still `""`, so `Offload` would have dropped it exactly as before: a fold that ran and changed nothing observable. That was the half most available to leave undone, and it has its own mutant (M18).

**I took `defer foldFallback()`, with one correction that turns out to be the whole fix.** It does not work against a local: every early return here is `return nil, r` on `var r sweepResult`, and a deferred mutation of a local happens *after* the return value has been copied — so the fold would compile, run, and be silently discarded on exactly the error paths it exists to cover. The `sweepResult` is a **named result** now; the other result stays blank-named, because only this one needs the treatment. One arming point covers all four exits that can carry a fallback. The explicit call after `r.rec` is assigned stays: that assignment overwrites the deferred fold's work-in-progress, and re-folding is free since `foldFallback` reads the accumulators rather than adding to them.

#### Vacuity — round 4

| # | Mutation | Result |
|---|---|---|
| M17 | revert to the per-site fold (the pre-fix shape) | **FAILED.** No-asker: "the ledger carries 0 rows" against a call counted in `/stats`. Ask-then-fail: `Strategy "prefix_ask"` and `LatencyMs 0` against ~8 ms |
| M18 | revert the identity establishment | **FAILED.** The no-asker row vanishes on the `Component != ""` guard |
| M19 | revert the **named result** | **FAILED**, identically to M17 — and this is the one that matters: `defer foldFallback()` is still present and still runs, and does nothing. A fix that looks present and is inert |

**On the assertion surface**, since the instruction was to assert on the rendered one: there isn't a marshalled surface for a `ModelCall` inside `/stats` — ledger rows travel to the dash `Event`, not this payload — so a JSON-tag assertion would test `dash`'s converter rather than this fix. What actually broke is that two surfaces **disagreed**, so that is what is pinned: if `/stats` counts the call, the ledger must carry a row for it. The assertion reads `rep.Calls` *after* `Offload`, i.e. past the guard that was doing the dropping, which is the same discipline by a different route. Both tests assert the precondition that the fallback genuinely **errored** (`sweep_fallback_failed`) before asserting the row, so neither can pass on a fixture that silently declined.

Running total across four rounds: **18 mutations, 18 killed** — three of them (M6, M13, and M19's named-result trap) only after the shape was corrected or reported as a survivor first.

## Checks

`gofmt -l .` — clean.
`CGO_ENABLED=1 go vet ./...` — clean, exit 0.
`CGO_ENABLED=1 go test ./...` — green, exit 0, 28 packages ok. Run on the eval box (Go 1.26.4) against the committed tree.

No benchmark was run.

## Open, and tracked

- **#179** — the dashboard keeps the conflation. Pre-existing; own PR.
- **#180** — `/metrics` per-component labels, and the cost gauges' inability to express `cost_source`.
- **#181** — `cheapmodel.Usage()` scoping and its stale comment.
- **#182** — the `ErrNoPrefix` accounting decision, awaiting @OsherElhadad.

Two things that are not fixable by code and so have no issue:

- **Why the 101 calls happened in one arm and not the other two, and whether the 59,009 ms mean was uniform or tail-driven.** That is what #177 says the missing record blocked, and the record unblocks it *going forward* — it cannot recover the answer for a run whose logs have no per-call trace. Both need a fresh measurement.
- **Per-component cost and latency claims made before this change need re-measuring.** They were drawn from the pooled block, and they do not become correct because the counter is fixed.

Assisted-By: Claude Opus 5 (1M context) &lt;noreply@anthropic.com&gt;

### DIFF
--- a/components/offload/extract_sweep.go
+++ b/components/offload/extract_sweep.go
@@ -305,16 +305,16 @@ func (e *ExtractSweep) Offload(req *bschemas.BifrostChatRequest, rep *components
 		// came from a model, but the REPLACEMENT is a pure function of (content, config), so a replay
 		// can never emit different bytes than the turn that decided it.
 		if cached, hit := getResult(c, id); hit {
-			metrics.RecordExtractionCacheLookup(true)
+			metrics.RecordExtractionCacheLookup(rep.Component, true)
 			if saved := schema.TextTokens(content) - schema.TextTokens(cached.Projected); saved > 0 {
-				metrics.RecordExtractionValue(float64(saved) * val.repeatPerToken)
+				metrics.RecordExtractionValue(rep.Component, float64(saved)*val.repeatPerToken)
 			}
 			if k, ok := applySweepDrop(c, rep, e.mode, msg, content); ok {
 				changed++
 				if k != "" {
 					keys = append(keys, k)
 				}
-				rep.Event("reapplied_same_session")
+				rep.Replay("reapplied_same_session")
 			}
 			continue
 		}
@@ -323,7 +323,7 @@ func (e *ExtractSweep) Offload(req *bschemas.BifrostChatRequest, rep *components
 			// all such a turn has to do.
 			continue
 		}
-		metrics.RecordExtractionCacheLookup(false)
+		metrics.RecordExtractionCacheLookup(rep.Component, false)
 		if schema.TextTokens(content) < e.minTokens {
 			rep.Gate("below_output_floor")
 			continue
@@ -610,7 +610,7 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 		reply, usage, err = c.PrefixAsk.Ask(ctx, c.Session, extract.BuildPrefixAsk(items))
 	}
 	latency := float64(time.Since(start).Milliseconds())
-	metrics.RecordExtractionCall(latency)
+	metrics.RecordExtractionCall(rep.Component, latency)
 	// COST, which this record carried as $0.00 forever. It never set CostUSD at all, so the per-call
 	// ledger the dashboard shows for this component reported zero on every firing — measured live at
 	// $0.00 against real cache reads of 449,304 and 449,376 tokens and real completion tokens, while
@@ -636,6 +636,11 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 			int64(usage.CacheWrite), int64(usage.CacheRead)),
 		GateReason: "pre-expiry window: the cache still exists and is nearly worthless",
 	}
+	// This component's own spend, priced at the REQUEST's model — the one it actually calls.
+	// /stats derived extraction cost from cheapmodel's process totals through the cheap-model
+	// rate card, so this component's frontier-model asks were priced at haiku rates and pooled
+	// with extract_llm's, summarize's and agentdiet's. See metrics.RecordExtractionSpend.
+	metrics.RecordExtractionSpend(rep.Component, r.rec.CostUSD)
 	if ctx.Err() != nil {
 		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
 			atomic.AddInt64(&llmTimeouts, 1)
@@ -814,8 +819,8 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 		r.event("sweep_dropped")
 		drop = append(drop, v.Label)
 		removed += sz - after
-		metrics.RecordExtractionSaving(sz - after)
-		metrics.RecordExtractionValue(float64(sz-after) * savedTokenValue(c).perToken)
+		metrics.RecordExtractionSaving(rep.Component, sz-after)
+		metrics.RecordExtractionValue(rep.Component, float64(sz-after)*savedTokenValue(c).perToken)
 	}
 	// An output named in the inventory that no verdict mentioned is UNJUDGED, and it must not look
 	// like a keep: 4ca1f13 found a live arm where the model silently omitted labels and the missing
--- a/components/offload/extract_sweep.go
+++ b/components/offload/extract_sweep.go
@@ -544,10 +544,14 @@ func (r *sweepResult) event(name string) { r.events = append(r.events, name) }
 // obligation, a verdict for something we did not offer. A wrong keep costs tokens on one turn; a wrong
 // drop is a silent permanent loss the agent does not notice and cannot ask about. The two errors are
 // not comparable, so this does not treat them symmetrically.
+// The sweepResult is a NAMED result purely so `defer foldFallback()` below can reach it. It has to
+// be: every early return here is `return nil, r` on a LOCAL, and a deferred mutation of a local
+// happens after the return value has already been copied, so the fold would be silently lost on
+// exactly the error paths it exists to cover. The first result stays blank-named — only this one
+// needs the treatment, and saying so beats leaving a reader to wonder what `dropped` is for.
 func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *components.Ctx,
-	rep *components.Report, cands []sweepCand) ([]int, sweepResult) {
+	rep *components.Report, cands []sweepCand) (_ []int, r sweepResult) {
 
-	var r sweepResult
 	// A SINGLE-CANDIDATE ASK IS THE REFUTED SHAPE WEARING A NEW NAME, so it is counted rather than
 	// silently accepted. Shown one output, a model simply drops it: 6% live-kept on haiku and 14% on
 	// sonnet, both inside the drop-everything null model's error bar. The ask still proceeds — a
@@ -648,12 +652,35 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 		return out, err
 	}
 	// foldFallback adds the fallback leg's tokens, dollars and wall time into the ask's ledger row.
-	// Idempotent by construction — at most one fallback runs per adjudication and this reads the
-	// accumulators rather than adding to them — so calling it on every path that can reach a
-	// fallback is safe.
+	//
+	// Idempotent by construction — at most one fallback runs per adjudication and this READS the
+	// accumulators rather than adding to them — which is what lets it be deferred and also called
+	// explicitly on the happy path, where the row is built after the no-asker fallback has already
+	// run and would otherwise overwrite the fold.
+	//
+	// DEFERRED, so it also covers the three fallback ERROR paths. Each of those is
+	// `if reply, err = runFallback(); err != nil { return nil, r }`, and `runFallback` has already
+	// booked the leg into metrics via recordLeg — so /stats counted the call and its seconds while
+	// the ledger row kept only the prefix ask's LatencyMs and a Strategy naming a leg that was no
+	// longer the only one that ran. Latency rather than dollars, because on an error the sink is
+	// empty (recordUsageCache is reached only after a successful decode on both backends, and
+	// neither returns an error after billing) — but the fallback is the SLOW leg by construction, so
+	// a failed one contributes tens of seconds to avg_latency_ms against a row showing milliseconds.
 	foldFallback := func() {
 		if fbMs == 0 && fbCost == 0 && fbUsage == (components.PrefixUsage{}) {
 			return
+		}
+		// IDENTITY FIRST, and this is the half of the fix that is easy to miss. On the no-asker
+		// path the fallback runs BEFORE r.rec exists, so if it errors the row is never built at
+		// all — Component stays "" and the caller drops the whole row on the
+		// `call.rec.Component != ""` guard. /stats then reports a call the ledger has no row for.
+		// That path never built a row before either, so the row is not a regression; the recorded
+		// spend and latency are new, so the DIVERGENCE is.
+		if r.rec.Component == "" {
+			r.rec.Component = rep.Component
+			r.rec.Model = c.ModelName
+			r.rec.CandidateTokens = before
+			r.rec.GateReason = "pre-expiry window: the cache still exists and is nearly worthless"
 		}
 		r.rec.LatencyMs = askMs + fbMs
 		r.rec.PromptTokens = int64(usage.Fresh + fbUsage.Fresh)
@@ -669,6 +696,10 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 			r.rec.Strategy = "prefix_ask+fallback"
 		}
 	}
+	// One arming point for all four exits that can carry a fallback — the three error returns and
+	// the happy path. A per-site call was what left the error paths uncovered, and a fifth site
+	// added later would have been missed the same way.
+	defer foldFallback()
 	if c.PrefixAsk == nil {
 		// No asker at all: a non-Anthropic route, or no incoming client. Not a failure of the ask —
 		// there was nothing to ask through — so it takes the same fork as a missed read.
@@ -702,7 +733,11 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 		CostUSD:    askCost,
 		GateReason: "pre-expiry window: the cache still exists and is nearly worthless",
 	}
-	foldFallback() // covers the no-asker path, whose fallback ran before this row existed
+	// The no-asker path's fallback ran BEFORE this row existed, and the assignment above just
+	// overwrote the deferred fold's work-in-progress. Re-folded here rather than relying on the
+	// defer alone because the defer's ordering relative to this assignment is what makes that
+	// reliance fragile — and folding twice is free, since foldFallback reads the accumulators.
+	foldFallback()
 	if ctx.Err() != nil {
 		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
 			atomic.AddInt64(&llmTimeouts, 1)
@@ -728,7 +763,6 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 			return nil, r
 		}
 		fellBack = true
-		foldFallback()
 	}
 	// THE CACHE READ IS THE MECHANISM'S WHOLE JUSTIFICATION, so a read that did not happen is always
 	// COUNTED — a silent miss is what hid this class of problem before, and it looks identical to a
@@ -759,10 +793,6 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 			return nil, r
 		}
 		fellBack = true
-		// The ledger row must cover EVERY leg. Folded here rather than at construction because two
-		// of the three fallback points fire after r.rec is built, and a row showing only the prefix
-		// ask is what made the dashboard's per-call view agree with the $0 figure.
-		foldFallback()
 	} else if !fellBack {
 		r.event("sweep_prefix_cache_read_ok")
 	}
--- a/components/offload/extract_sweep.go
+++ b/components/offload/extract_sweep.go
@@ -583,34 +583,6 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 	for _, it := range items {
 		before += it.SizeTokens
 	}
-	start := time.Now()
-	var (
-		reply string
-		usage components.PrefixUsage
-		err   error
-		// fellBack records that the expensive path has already run, so the cache-read check below does
-		// not fire a second time for the same call. Without it a failed ask both fell back AND then
-		// reported a zero cache read, double-counting one event as two.
-		fellBack bool
-	)
-	if c.PrefixAsk == nil {
-		// No asker at all: a non-Anthropic route, or no incoming client. Not a failure of the ask —
-		// there was nothing to ask through — so it takes the same fork as a missed read.
-		r.gate("sweep_no_asker")
-		if e.blockFallback {
-			r.gate("sweep_fallback_blocked")
-			r.rec.Rejection = "no prefix asker on this route and block_fallback is set"
-			return nil, r
-		}
-		if reply, err = e.fallbackAsk(ctx, req, c, &r, items, cands); err != nil {
-			return nil, r
-		}
-		fellBack = true
-	} else {
-		reply, usage, err = c.PrefixAsk.Ask(ctx, c.Session, extract.BuildPrefixAsk(items))
-	}
-	latency := float64(time.Since(start).Milliseconds())
-	metrics.RecordExtractionCall(rep.Component, latency)
 	// COST, which this record carried as $0.00 forever. It never set CostUSD at all, so the per-call
 	// ledger the dashboard shows for this component reported zero on every firing — measured live at
 	// $0.00 against real cache reads of 449,304 and 449,376 tokens and real completion tokens, while
@@ -623,24 +595,114 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 	// rather than being two independent guesses. c.SelfRates is the model the request came in on;
 	// falling back to the env card keeps a figure when the host supplies no rates, the same
 	// convention extract_llm.pricingFor uses.
+	//
+	// Resolved BEFORE the ask because both legs below price themselves with it. The fallback goes to
+	// the same model the prefix ask addresses (c.Model.For("incoming")), so one rate card is right
+	// for both.
 	pricing := cheapmodel.PricingFromEnv()
 	if !c.SelfRates.Zero() {
 		pricing = ratesPricing(c.SelfRates)
 	}
+	var (
+		reply string
+		usage components.PrefixUsage
+		err   error
+		// fellBack records that the expensive path has already run, so the cache-read check below does
+		// not fire a second time for the same call. Without it a failed ask both fell back AND then
+		// reported a zero cache read, double-counting one event as two.
+		fellBack bool
+		// The ask's own totals, accumulated PER LEG. An adjudication is not one model call: the
+		// prefix ask and the fallback are two, on two prompts, and either can be the only one that
+		// happens. See recordLeg.
+		askMs, askCost float64
+		fbUsage        components.PrefixUsage
+		fbMs, fbCost   float64
+	)
+	// recordLeg books ONE model call this component made — its wall time and its own priced spend.
+	//
+	// PER LEG, not once per adjudication, because `fallbackAsk` is a SECOND model call on a full
+	// sampled transcript and it used to be accounted at $0.00 and 0 ms. The single record was built
+	// from the PREFIX ask's usage and assigned once, while two of the three fallback points fire
+	// after that assignment — so on every route without an Anthropic prefix asker (where the prefix
+	// ask never happens at all), on every session's first turn (ErrNoPrefix) and on every mistimed
+	// window (CacheRead == 0), the component's real frontier-model spend was reported as free.
+	//
+	// That was survivable before this PR only by accident: the fallback's tokens still reached
+	// /stats through cheapmodel's process totals, which the host passed in as the `cost` argument.
+	// Making the components price their own spend removed that accident, so the leg has to book
+	// itself. This is the same defect class as #176, in the component this PR re-scoped.
+	recordLeg := func(ms float64, u components.PrefixUsage) float64 {
+		metrics.RecordExtractionCall(rep.Component, ms)
+		cost := pricing.Cost(int64(u.Fresh), int64(u.Output),
+			int64(u.CacheWrite), int64(u.CacheRead))
+		metrics.RecordExtractionSpend(rep.Component, cost)
+		return cost
+	}
+	// runFallback runs the expensive path and books it as its own leg, at all three call sites, so a
+	// fourth one cannot be added without the accounting coming with it. It records even when the
+	// call ERRORS: a failed completion still burned wall time, and on some failures tokens.
+	runFallback := func() (string, error) {
+		out, u, ms, err := e.fallbackAsk(ctx, req, c, &r, items, cands)
+		fbUsage, fbMs = u, ms
+		fbCost = recordLeg(ms, u)
+		return out, err
+	}
+	// foldFallback adds the fallback leg's tokens, dollars and wall time into the ask's ledger row.
+	// Idempotent by construction — at most one fallback runs per adjudication and this reads the
+	// accumulators rather than adding to them — so calling it on every path that can reach a
+	// fallback is safe.
+	foldFallback := func() {
+		if fbMs == 0 && fbCost == 0 && fbUsage == (components.PrefixUsage{}) {
+			return
+		}
+		r.rec.LatencyMs = askMs + fbMs
+		r.rec.PromptTokens = int64(usage.Fresh + fbUsage.Fresh)
+		r.rec.CompletionTokens = int64(usage.Output + fbUsage.Output)
+		r.rec.CacheRead = int64(usage.CacheRead + fbUsage.CacheRead)
+		r.rec.CacheWrite = int64(usage.CacheWrite + fbUsage.CacheWrite)
+		r.rec.CostUSD = askCost + fbCost
+		// Name what actually ran. On the no-asker route the prefix ask never happens, so calling
+		// the row "prefix_ask+fallback" would report a leg that did not exist.
+		if c.PrefixAsk == nil {
+			r.rec.Strategy = "fallback"
+		} else {
+			r.rec.Strategy = "prefix_ask+fallback"
+		}
+	}
+	if c.PrefixAsk == nil {
+		// No asker at all: a non-Anthropic route, or no incoming client. Not a failure of the ask —
+		// there was nothing to ask through — so it takes the same fork as a missed read.
+		r.gate("sweep_no_asker")
+		if e.blockFallback {
+			r.gate("sweep_fallback_blocked")
+			r.rec.Rejection = "no prefix asker on this route and block_fallback is set"
+			return nil, r
+		}
+		if reply, err = runFallback(); err != nil {
+			return nil, r
+		}
+		fellBack = true
+	} else {
+		askStart := time.Now()
+		reply, usage, err = c.PrefixAsk.Ask(ctx, c.Session, extract.BuildPrefixAsk(items))
+		askMs = float64(time.Since(askStart).Milliseconds())
+		// ErrNoPrefix is refused LOCALLY — there is no stashed body to append to, so no request
+		// leaves the process. Booking it as a call would put a 0 ms, $0 sample into the mean that
+		// the exploration brake reads, and inflate `calls` with work that by definition did not
+		// happen. Every other outcome, transport failure included, went to the provider.
+		if !errors.Is(err, components.ErrNoPrefix) {
+			askCost = recordLeg(askMs, usage)
+		}
+	}
 	r.rec = components.ModelCall{
 		Component: rep.Component, Model: c.ModelName, Strategy: "prefix_ask",
-		CandidateTokens: before, LatencyMs: latency,
+		CandidateTokens: before, LatencyMs: askMs,
 		PromptTokens: int64(usage.Fresh), CompletionTokens: int64(usage.Output),
 		CacheRead: int64(usage.CacheRead), CacheWrite: int64(usage.CacheWrite),
-		CostUSD: pricing.Cost(int64(usage.Fresh), int64(usage.Output),
-			int64(usage.CacheWrite), int64(usage.CacheRead)),
+		CostUSD:    askCost,
 		GateReason: "pre-expiry window: the cache still exists and is nearly worthless",
 	}
-	// This component's own spend, priced at the REQUEST's model — the one it actually calls.
-	// /stats derived extraction cost from cheapmodel's process totals through the cheap-model
-	// rate card, so this component's frontier-model asks were priced at haiku rates and pooled
-	// with extract_llm's, summarize's and agentdiet's. See metrics.RecordExtractionSpend.
-	metrics.RecordExtractionSpend(rep.Component, r.rec.CostUSD)
+	foldFallback() // covers the no-asker path, whose fallback ran before this row existed
 	if ctx.Err() != nil {
 		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
 			atomic.AddInt64(&llmTimeouts, 1)
@@ -662,10 +724,11 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 			r.rec.Rejection = "prefix ask failed and block_fallback is set: " + err.Error()
 			return nil, r
 		}
-		if reply, err = e.fallbackAsk(ctx, req, c, &r, items, cands); err != nil {
+		if reply, err = runFallback(); err != nil {
 			return nil, r
 		}
 		fellBack = true
+		foldFallback()
 	}
 	// THE CACHE READ IS THE MECHANISM'S WHOLE JUSTIFICATION, so a read that did not happen is always
 	// COUNTED — a silent miss is what hid this class of problem before, and it looks identical to a
@@ -692,10 +755,14 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 				"declining rather than paying again for a full-price transcript read"
 			return nil, r
 		}
-		if reply, err = e.fallbackAsk(ctx, req, c, &r, items, cands); err != nil {
+		if reply, err = runFallback(); err != nil {
 			return nil, r
 		}
 		fellBack = true
+		// The ledger row must cover EVERY leg. Folded here rather than at construction because two
+		// of the three fallback points fire after r.rec is built, and a row showing only the prefix
+		// ask is what made the dashboard's per-call view agree with the $0 figure.
+		foldFallback()
 	} else if !fellBack {
 		r.event("sweep_prefix_cache_read_ok")
 	}
@@ -856,14 +923,18 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 //
 // The reply budget is raised through components.Budgeter where the client supports it, for the same
 // reason the prefix ask raises it: one reply carries a verdict for every candidate.
+// It returns its OWN usage and wall time so the caller can price it. `model.Complete` reports
+// neither, so the usage is read from a per-call cheapmodel sink nested inside whatever scope already
+// wraps ctx — the same construction extract_llm uses to attribute one call. The sink also reaches
+// every ancestor, so the request's own bill does not lose these tokens.
 func (e *ExtractSweep) fallbackAsk(ctx context.Context, req *bschemas.BifrostChatRequest,
 	c *components.Ctx, r *sweepResult, items []extract.AdjudicationItem,
-	cands []sweepCand) (string, error) {
+	cands []sweepCand) (string, components.PrefixUsage, float64, error) {
 	model := c.Model.For("incoming")
 	if model == nil {
 		r.gate("sweep_fallback_no_model")
 		r.rec.Rejection = "the prefix ask could not read the cache and no request model is available"
-		return "", errNoFallbackModel
+		return "", components.PrefixUsage{}, 0, errNoFallbackModel
 	}
 	if b, ok := model.(components.Budgeter); ok {
 		if m := b.WithMaxTokens(cheapmodel.PrefixAskMaxTokens); m != nil {
@@ -878,13 +949,22 @@ func (e *ExtractSweep) fallbackAsk(ctx context.Context, req *bschemas.BifrostCha
 		withSamples[i] = it
 	}
 	r.event("sweep_fallback_used")
+	ctx, sink := cheapmodel.WithCallSink(ctx)
+	start := time.Now()
 	reply, err := model.Complete(ctx, extract.BuildFallbackAsk(sweepIntent(req), withSamples))
+	ms := float64(time.Since(start).Milliseconds())
+	// Read the sink whatever happened: a completion that failed after the provider billed it still
+	// cost money, and returning zeros there is how spend goes missing.
+	_, inTok, outTok := sink.Totals()
+	cw, cr := sink.CacheTotals()
+	u := components.PrefixUsage{Fresh: int(inTok), Output: int(outTok),
+		CacheWrite: int(cw), CacheRead: int(cr)}
 	if err != nil {
 		r.gate("sweep_fallback_failed")
 		r.rec.Rejection = "fallback completion failed: " + err.Error()
-		return "", err
+		return "", u, ms, err
 	}
-	return reply, nil
+	return reply, u, ms, nil
 }
 
 // sweepIntent renders the conversation's intent for a SPENT-NESS judgement, which wants it ordered
--- a/metrics/extract.go
+++ b/metrics/extract.go
@@ -39,6 +39,16 @@ import (
 // ~59-second asks braked extract_llm's exploration on evidence from a different component
 // and a different model. A brake must read the latency of the calls it is braking.
 
+// CostSource values for ExtractStats.CostSource. Constants because they are read off /stats by
+// operators and by promexport; a typo in one of two string literals is a silently wrong label.
+const (
+	costSourceComponent = "component"
+	costSourceHost      = "host_total"
+	costSourcePartial   = "partial"
+	costSourceUnpriced  = "unpriced"
+	costSourceNone      = "none"
+)
+
 // xCounters is one component's extraction accounting.
 type xCounters struct {
 	calls      atomic.Int64 // extraction LLM calls actually made
@@ -250,9 +260,21 @@ func RecordExtractionSpend(component string, usd float64) {
 	}
 }
 
-// ExtractStats is the extraction economics block served inside /stats. It is ADDITIVE:
-// every pre-existing /stats field keeps its name and meaning, because deploy/harbor/*.py
-// parses them.
+// ExtractStats is the extraction economics block served inside /stats. It is ADDITIVE: every
+// pre-existing field keeps its name and meaning.
+//
+// NOT because the benchmark harness parses it — that reason was carried here and in
+// docs/components/extract_llm.md and it is FALSE. `deploy/` reads `runs`, `acted` and
+// `saved_tokens` off the per-COMPONENT objects (measure.py does so by direct index, so removing
+// one is a KeyError), and reads nothing at all out of this block; grepping deploy/ for
+// gross_saved_tokens, calls_avoided, extraction_cost_usd or avg_latency_ms returns nothing.
+//
+// The real reason these names are frozen is /metrics. proxy/promexport.go publishes them as
+// cg_extract_calls_total, cg_extract_cost_usd, cg_extract_net_value_usd and cg_extract_latency_ms,
+// and dash/metrics_export.go documents cg_extract_net_value_usd as the endpoint's only dollar
+// figure. Those are what an operator's alert rule and dashboard query are written against, off-repo
+// and unversioned, so a rename breaks monitoring silently — which is a worse failure than breaking
+// the harness, because the harness would at least crash.
 type ExtractStats struct {
 	Calls            int64 `json:"calls"`            // extraction LLM calls made
 	CallsAvoided     int64 `json:"calls_avoided"`    // global result-cache hits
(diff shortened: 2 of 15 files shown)


components/offload/extract_sweep.go
```diff
@@ -636,6 +636,11 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 			int64(usage.CacheWrite), int64(usage.CacheRead)),
 		GateReason: "pre-expiry window: the cache still exists and is nearly worthless",
 	}
+	// This component's own spend, priced at the REQUEST's model — the one it actually calls.
+	// /stats derived extraction cost from cheapmodel's process totals through the cheap-model
+	// rate card, so this component's frontier-model asks were priced at haiku rates and pooled
+	// with extract_llm's, summarize's and agentdiet's. See metrics.RecordExtractionSpend.
+	metrics.RecordExtractionSpend(rep.Component, r.rec.CostUSD)
```
amiddavid: **MAJOR: the sweep's fallback-ask spend and latency are dropped from `extraction_cost_usd` entirely.**

`RecordExtractionSpend(rep.Component, r.rec.CostUSD)` sits immediately after `r.rec` is built from the *prefix ask's* `usage`, and `r.rec` is assigned exactly once (line 630) — never recomputed. But `fallbackAsk` makes a real second model call on `c.Model.For("incoming")`, the request's own frontier model, on three live paths:

- `c.PrefixAsk == nil` (any non-Anthropic route) — `usage` is never assigned here at all, so `CostUSD == 0` and `RecordExtractionSpend` is skipped by its `usd > 0` guard;
- `ErrNoPrefix` — every session's first turn;
- `usage.CacheRead == 0` — a mistimed window.

`fallbackAsk` returns `(string, error)`: no usage, no cost. Before this PR that spend still reached `/stats`, because `model.Complete` lands in `cheapmodel`'s process totals (`internal/cheapmodel/anthropic.go:161` calls `recordUsageCache`) and was priced into the `cost` argument. Now, as soon as *any* component records spend (`anySpend == true`, `metrics/extract.go:346`), the host fallback is discarded and the fallback ask's cost is reported as **$0**.

Concrete scenario: `extract_llm` + `extract_llm_sweep` behind an OpenAI upstream. The sweep always falls back, pays frontier rates for a full sampled transcript on every firing, and `/stats` shows `extract_llm_sweep.extraction_cost_usd: 0` with a purely positive `net_value_usd` — a component whose entire justification is cost, reported as free. That is the same class of defect as #176, in the component this PR just re-scoped.

The ordering also excludes the fallback's wall time from `RecordExtractionCall` (line 613, before the fallback) on the latter two paths, so `avg_latency_ms` under-reports as well.

metrics/extract.go
```diff
@@ -200,50 +308,124 @@ type ExtractStats struct {
...
+		spend += sp
+		anySpend = anySpend || recorded
+		totLatMs += x.latencyMs.Load()
+		for k, v := range reasons {
+			totReason[k] += v
+		}
+	}
+	// The components' own priced spend where any of them recorded some; the host's
+	// process-global figure only as a fallback. See xCounters.spendNano.
+	if !anySpend {
```
amiddavid: **The host-fallback cost is applied only to the aggregate, so `by_component` rows can contradict the block that encloses them.**

`spend = cost` is all-or-nothing on the total; `xCounters.snapshot` never sees `cost`, so a component that recorded no spend gets `ExtractionCostUSD: 0` and `NetValueUSD: +grossValue`.

This PR's own test pins that shape: `TestRecordedSpendBeatsTheHostsGlobalFigure` — after `RecordExtractionCall("extract_llm", 100)` with no spend, top-level `extraction_cost_usd == 9.99` while the only `by_component` row reads `0`. Scenario: a library embedding or `/compact`, where `ModelCall.CostUSD` is never filled — the operator reads `by_component.extract_llm.net_value_usd` as comfortably positive while the block directly above says the component is underwater. Two figures for the same quantity disagreeing by the whole spend is precisely the failure mode this change exists to remove.

**Separately, partial recording silently under-reports the aggregate.** `anySpend` is one boolean across all components. If `extract_llm` prices its calls (Haiku card, always non-zero) but `extract_llm_sweep` records nothing — via the fallback-ask path above, or `c.SelfRates` zero with zero usage — the total becomes `extract_llm`'s spend alone and the sweep's real dollars vanish rather than falling back to the host figure.

Both point the same way: the fallback wants to be per component, not per snapshot. Where a component genuinely has no priced spend, marking the row "spend not recorded" would beat printing `0`, which is indistinguishable from "spent nothing".

amiddavid (commented):
## Review

Reviewed the 11-file diff against `origin/main`, and independently re-ran the checks on the eval box (Go 1.26.4, `CGO_ENABLED=1`): `go vet ./...` clean, `go test ./...` fully green, `gofmt -l .` clean. Commit trailers are correct — DCO `Signed-off-by`, the exact `Assisted-By` line, conventional title, and the mutation table in the commit body rather than only in the PR description.

The core of the change holds up. Every `metrics.RecordExtraction*` / `Extraction*LatencyMs` call site is converted; `rep.Component` is set by `components/pipeline.go:81` before `Offload` runs, so no recorder sees an empty key on the proxy path. Both `reapplied_*` sites in each component go through `Report.Replay`, with no third path missed. `compStat.act` is called on the same `*compStat` from both the enforced and observe loops and `reverseDiscarded` does not touch `Acted`, so `acted == acted_fresh + acted_replay` holds. In the per-call DEBUG record, every referenced variable is assigned before the log, and the single-flight follower's early return correctly emits nothing. `xCounters` locking, the per-component latency ring, and `sortedReasons`' nil-in/nil-out contract are sound, and `ByComponent` cannot recurse.

Four things below. Two are inline on the diff; two are not reachable inline.

### The dashboard keeps the exact reading #176 describes

`dash/event.go:527` still computes `row.Acted = row.Mutated && row.SavedGross > 0`, so a free frozen replay still counts as an act. `components.Report.Replays` is available on `r` at that point and is not read. `/api/components`'s `acted`, `acted_tokens` and `act_rate` (`dash/query.go:776-782`) therefore keep conflating the 2,291 free replays with paid extractions — on the surface the dashboard actually renders. `/stats` is fixed; the dashboard is not, and this is not listed among the PR's known gaps. Whether it belongs in this PR or a follow-up is a scope call, but it should be stated either way rather than left to be rediscovered.

### Docs still describe the old scoping

- `docs/reference/routes.md:84` — the per-component field table is exhaustive by convention and now omits `acted_fresh` and `acted_replay`. `extract.by_component` is documented nowhere.
- `docs/components/extract_llm.md:678` — this table still presents `calls`, `extraction_cost_usd`, `avg_latency_ms` and `gross_saved_tokens` as extract_llm's own figures. That is exactly the misreading the PR exists to correct: post-change those keys are the *sum* across components, and the doc is now the artifact that produces the #176 misread. `statsGoldenComponent` was updated; the prose contract was not.

### One claim in the PR body does not check out

The body and `docs/components/extract_llm.md:677` both justify key stability with "`deploy/harbor/*.py` parses it". I grepped `deploy/` for `gross_saved_tokens`, `calls_avoided`, `extraction_cost_usd` and `avg_latency_ms` — no harbor script reads any of them. Keeping the keys stable is still the right call for operators and for `/metrics`, but the stated reason is not the real one, and it is load-bearing in the design rationale.

Also noting, since it bears on finding 1: `/metrics` staying aggregate is called out as deliberate, but `cg_extract_cost_usd` will now inherit whatever `extraction_cost_usd` computes, including the dropped fallback spend.

components/offload/extract_sweep.go
amiddavid: Confirmed and fixed in e95e90c — the diagnosis was exactly right, including that this PR is what made it reachable: before, the fallback's tokens still landed in /stats through cheapmodel's process totals via the `cost` argument, and making components price their own spend removed that accident.

What changed:

- `fallbackAsk` now returns `(string, components.PrefixUsage, float64, error)`. `model.Complete` reports no usage, so it is read from a per-call `cheapmodel.WithCallSink` nested inside whatever scope already wraps ctx — the construction extract_llm already uses, and it still reaches every ancestor so the request's own bill keeps the tokens. The sink is read **whether or not the call errored**: a completion that failed after the provider billed it still cost money, and returning zeros there is how spend goes missing.
- Each **leg** books itself, through one `recordLeg` closure reached from all three fallback sites via `runFallback`, so a fourth site cannot be added without the accounting coming with it. Two model calls now count as two, and the fallback's wall time is its own sample rather than being folded into a figure measured before it ran.
- `foldFallback` grows the ledger row to cover both legs — tokens, dollars, wall time — and names the strategy for what actually ran: `"fallback"` on the no-asker route, `"prefix_ask+fallback"` otherwise, rather than a row claiming a prefix-ask leg that never happened.

One deliberate exception, which is a small behaviour change: an `ErrNoPrefix` refusal is **not** booked as a call. It is refused locally before any request leaves the process, so counting it would inflate `calls` with work that by definition did not happen and put a 0 ms sample into the mean `tooSlowToExplore` reads. Say the word if you would rather keep it counted for continuity with the old `calls` series.

Two new tests drive this through the real recording path — a `cheapmodel.Anthropic` against an `httptest` server that reports usage, because `recordingModel` bills nothing and a fake that bills nothing would let the assertion pass against a fix that plumbs nothing:

- `TestSweepFallbackAskPricesItsOwnCall` — the no-asker path, your OpenAI-upstream scenario.
- `TestSweepCountsBothLegsWhenItFallsBackAfterAsking` — the zero-cache-read path, where the fold matters because the row already exists.

Vacuity: three mutations, all killed. Zeroing the returned usage → ledger row $0, row tokens 0/0 against 31000/420, /stats delta $0, `cost_source: "unpriced"`. Removing the leg booking → 0 calls where 1 is due, 1 where 2 are due. Removing the fold → row shows prompt 40 / completion 90 of 31040/510 and the wrong Strategy.

metrics/extract.go
amiddavid: Both halves confirmed and fixed in e95e90c. And yes — `TestRecordedSpendBeatsTheHostsGlobalFigure` pinned the bad shape, which is a fair hit: the test asserted the behaviour rather than questioning it.

I did not make the host figure per-component, because I do not think it can be made per-component honestly: `cost` is one process-global number that also carries `summarize` and `agentdiet`, and splitting it by call count would be an invented number published at the same precision as a measured one. So the fix is to make the unknown *visible* instead of guessing at it:

- **`net_value_usd` is now `*float64`**, rendered as JSON `null` when the spend behind it is not known. A row that made calls and priced none of them no longer publishes `+grossValue`. The **aggregate is never null** — it always has a determined spend — so the `cg_extract_net_value_usd` gauge is unaffected in practice; `promexport` reads it through a `Net() (float64, bool)` accessor and omits the series rather than publishing an unknown as `0`, which on that gauge reads as exactly break-even.
- **`cost_source`** names the case, which is your "spend not recorded" made machine-readable: `component` (every call priced itself), `host_total` (the host's superset figure), `partial` (some calls unpriced, so the total is a **floor**), `unpriced` (this row made calls and priced none — nothing is known), `none` (no calls, so `0` is the true figure).
- **Partial recording no longer under-reports.** `unpricedCalls` replaces the single `anySpend` boolean: when any calls are unpriced the aggregate publishes the larger of the recorded floor and the host's superset figure, and labels which one it chose. A total that quietly omits real dollars is the defect; a loud one is not.

Note the two findings interlock — with finding 1 fixed, the sweep prices every leg, so `unpriced` is no longer reachable through the fallback path. It stays reachable for a host that never fills `ModelCall.CostUSD` at all (a library embedding, `/compact`), which is the case the label is for.

Two new tests, `TestAnUnpricedRowSaysSoInsteadOfClaimingZero` and `TestPartialPricingDoesNotSilentlyUnderReportTheTotal`, and the old one now asserts the corrected contract instead of the shape you flagged.

**One thing worth flagging about my own vacuity run.** The `cost_source` mutant (`json:"-"`) **survived** the first pass, because my assertions read the Go struct field rather than the encoded payload — the exact "computed correctly and dropped by the handler" failure the rendered-output rule exists for, committed in the same change that cites the rule. `TestByComponentSurvivesJSON` and `TestStatsRendersPerComponentExtractionAttribution` now assert `cost_source` and the `null` net on the marshalled payload and on the `/stats` body, and the mutant then fails on both surfaces. Recorded in the PR body rather than quietly fixed, since a mutant that survives and then dies is the only evidence the test change did anything.

One limitation I did not solve: `cost_source: partial` cannot say *which* component is unpriced. Finding the culprit still means scanning the rows for `cost_source: unpriced`. Adequate, not ideal — noted in the PR's known gaps.

amiddavid:
Thanks — all four findings hold up, and two of them were mine to answer for. Fixed in e95e90c; the PR body is rewritten with a **Review round 2** section. The two inline threads have detailed replies; this covers the rest.

### Docs (finding 3)

You are right that the doc had become the artifact producing the misread, which is worse than it being merely stale.

- `docs/reference/routes.md` — `acted_fresh` / `acted_replay` added to the per-component table, with `acted` itself now flagged as including free replays and a worked contrast ("`acted: 239` with `acted_replay: 239` made no calls; the same `acted` with `acted_replay: 0` paid for every one"). New **Extraction economics** section documents `by_component`, `cost_source` and the nullable net, with a warning that the enclosing block is a sum.
- `docs/components/extract_llm.md` — the table now opens with a danger admonition stating those keys are the sum across extraction components, using the measured numbers as the worked example (101 calls / 59,009 ms / −$1.162 against `cands: 0` on all 374 requests), and directs every per-component claim to `by_component`. The `acted`-counts-replays note is there too.

### The harbor claim (finding 4)

Verified, and you are right. Grepping `deploy/` for `gross_saved_tokens`, `calls_avoided`, `extraction_cost_usd` and `avg_latency_ms` returns nothing — no harbor script reads any key in that block.

What harbor *does* read is `runs`, `acted` and `saved_tokens` off the per-**component** objects, and `measure.py` reads them by direct index (`v["acted"]`), so a removal there is a `KeyError` rather than a silent zero. So the reasoning was sound and attached to the wrong table — which is the more embarrassing failure of the two, because it is the kind that propagates by being quoted.

The real reason those names are frozen is `/metrics`: `promexport` publishes them as `cg_extract_calls_total`, `cg_extract_cost_usd`, `cg_extract_net_value_usd`, `cg_extract_latency_ms`, and `dash/metrics_export.go` documents the last as the endpoint's only dollar figure. Those are queried by off-repo, unversioned alert rules, where a rename breaks monitoring *silently* — worse than breaking the harness, which would at least crash. Corrected in `metrics/extract.go`, `docs/reference/routes.md` and `docs/components/extract_llm.md`, with the two consumers distinguished rather than blurred, and in the PR body.

### The dashboard (`dash/event.go:527`) → #179

Taking the scope call the other way, per the standing rule that a `main` defect unrelated to the change under test gets its own issue and its own PR so it can ship without waiting: filed as **#179** with the evidence — the `row.Acted` computation, `r.Replays` being available and unread at that line, and the downstream `acted` / `acted_tokens` / `act_rate` / `act_rate_structural` in `dash/query.go:776-782`. Listed in this PR's known gaps.

Your real objection — that it was unstated — is fair and taken; leaving it to be rediscovered was the mistake, not leaving it unfixed. I also noted in #179 that `act_rate_structural` is derived as `mutated − acted`, so that derivation wants re-checking against whatever shape the split takes there rather than being left to compose by accident.

And your note that `cg_extract_cost_usd` inherits whatever `extraction_cost_usd` computes is now sharper than when you wrote it: after round 2 it inherits the `cost_source` regime without being able to express it. That is one more reason the labelled series is the right follow-up, and it is in the known gaps.

### One thing about my own vacuity run

The `cost_source` mutant **survived** the first pass. My assertions read the Go struct field rather than the encoded payload — the exact "a field can be computed and dropped by the handler" failure this repo's rendered-output rule exists for, committed in the same change that cites the rule. Both tests now assert on the marshalled payload and on the `/stats` body, and the mutant fails on both. Recorded in the PR body rather than quietly fixed, because a mutant that survives and then dies after a test change is the only evidence the test change did anything.

Round 2 totals: six new mutations, all six killed. `gofmt -l` clean, `CGO_ENABLED=1 go vet ./...` clean, `CGO_ENABLED=1 go test ./...` green at 28 packages, on the eval box against the committed tree. No benchmark.

amiddavid:
Round 3 (`6e01798`): the last of the four items I'd left as prose is fixed in-PR, and the other three now have numbers. Nothing is tracked as narrative any more.

### Moved into the PR

**`cost_source: partial` now names what the total is short of.** You were right that this was reachable here — the aggregate already had everything it needed. `extract.unpriced_components` names them, sorted, aggregate only, omitted when every call priced itself.

Two details worth flagging:

- Published under **both** incomplete regimes, not just `partial`. They are incomplete for the same reason, and `host_total` needs it more if anything: under `partial` the list says what the floor is short of, under `host_total` it says what *forced* the fallback in the first place.
- The argument for doing it at all is that the snapshot had already computed the answer — `!recorded && s.Calls > 0` is the same test that sets the aggregate's label — so making the reader re-derive it from the rows was asking them to redo work already done. And a reader who has to redo work will read the label and stop, at which point the floor gets quoted as the bill.

Documented in `docs/reference/routes.md` and `docs/components/extract_llm.md`.

**Vacuity: three mutations, all three killed, each on both surfaces.** M14 the list never populated; M15 computed but never attached to the snapshot; M16 dropped by the encoder (`json:"-"`). M16 is the M13 shape, which is exactly why both assertions read the marshalled payload rather than the Go field — `TestTheAggregateNamesWhichComponentIsUnpriced` through `json.Marshal`, `TestStatsNamesTheUnpricedComponent` on the `/stats` body. Both check the precondition that the total is genuinely incomplete before asserting the list, and both check the field is *omitted* when everything priced itself, since an empty list on a complete total reads as a warning where there is none.

### Became issues

| | | |
|---|---|---|
| **#180** | `cg_extract_*` per-component labels | Can't ship here — a label changes series shape that off-repo alert rules are written against. Includes the round-2 sharpening you asked for: the cost gauges now *inherit* the `cost_source` regime with no way to express it, so a gauge reading `−0.71` may be priced arithmetic, a floor, or the host's superset figure, identically. That makes the unlabelled series actively misleading rather than merely coarse |
| **#181** | `cheapmodel.Usage()` scoping | The "Model interface carries a label" refactor its own comment defers. Records what this PR established as the reason deferring is *safe*: /stats depends on it only as a **labelled** fallback now (`cost_source: host_total`, documented as a superset carrying `summarize` and `agentdiet`), so the wrongness is declared at the point of use rather than silently inherited. Also notes the comment should be corrected as part of the fix or ahead of it — a stale premise stated as a design rationale is how #176 survived review in the first place |
| **#182** | the `ErrNoPrefix` accounting decision, **assigned to @OsherElhadad** | Position taken: keep it uncounted, because a refusal that never leaves the process is not a call and a counter named `calls` that includes calls never made is how #176 happened. Continuity cost named explicitly — `extract.calls` and `cg_extract_calls_total{outcome="made"}` step down by roughly one per session, and "the component stopped working" and "the counter got honest" are indistinguishable from outside. Concrete alternative given (count it under `refusals`, which I think is *better* than what I shipped and which I skipped only to avoid widening this PR further), plus what would change my mind and what wouldn't |

**#179** was already filed after round 2 and is referenced in the gaps table.

### Dropped

`docs/experiments/loca/iter023/results.md` — on `feat/coref-recut`, not `main`, so this PR can't touch it. Removed from the gaps; the general point that pre-change per-component claims need re-measuring stays, without implying a fix was owed here. Thanks for taking it.

### State

`gofmt -l .` clean, `CGO_ENABLED=1 go vet ./...` clean, `CGO_ENABLED=1 go test ./...` green at 28 packages, on the eval box against the committed tree. No benchmark.

Running total across the three rounds: 15 mutations, 15 killed — two of them (M6's redundant DEBUG guard, M13's `cost_source` reading the struct instead of the payload) only after I'd reported them as survivors.

metrics/extract.go
amiddavid: Following up on the one limitation I flagged when closing this thread — that `cost_source: partial` could not say *which* component was unpriced, so finding the culprit meant scanning the rows. That is fixed in `6e01798` rather than left as "adequate, not ideal".

`ExtractStats.UnpricedComponents` (`unpriced_components`) names them, sorted, on the aggregate only, omitted when every call priced itself. The argument for doing it: `!recorded && s.Calls > 0` is the *same test* that sets the aggregate's label, so the snapshot already knew the answer and was making the reader re-derive it — and a reader who has to re-derive will read the label and stop, at which point the floor gets quoted as the bill.

Published under **both** incomplete regimes rather than only `partial`. Under `partial` the list says what the floor is short of; under `host_total` it says what forced the fallback. Same reason in both cases.

Vacuity: three mutations, all killed on both surfaces — the list never populated, computed-but-not-attached, and dropped by the encoder (`json:"-"`). That last one is the M13 shape, so both new assertions read the marshalled payload: `TestTheAggregateNamesWhichComponentIsUnpriced` through `json.Marshal` and `TestStatsNamesTheUnpricedComponent` on the `/stats` body. Each asserts the precondition that the total really is incomplete before asserting the list, and each asserts the field is omitted when nothing is unpriced.

The other two limitations from this area became issues rather than staying prose: **#180** (`cg_extract_cost_usd` inherits the `cost_source` regime with no way to express it) and **#181** (`cheapmodel.Usage()` scoping, plus its comment still asserting the premise that produced #176 — with the reason deferring is safe recorded there). Full summary in the PR body and in the top-level comment.

amiddavid (commented):
## Re-review

Both regressions are genuinely fixed, and I verified the mechanisms rather than the claims.

**Finding 1 (the sweep's fallback ask reported $0 and 0 ms) — fixed, and correctly.** `fallbackAsk` now returns its own usage and wall time from a nested `cheapmodel` sink, and `recordLeg` books each leg from a single closure reached by all three call sites. I checked the two things the design rests on: `WithCallSink` chains through `parent` (`internal/cheapmodel/usage.go:113`, `add` walking `cur = cur.parent`), so the request's own bill does not lose these tokens; and `recordUsageCache` feeds `SinkFrom(ctx)` on both backends, so the OpenAI route is covered too, which matters because that route is precisely where the fallback is the *only* call. The `!fellBack` guards mean exactly one fallback runs per adjudication, so `foldFallback`'s idempotency claim holds. Resolving `pricing` before the ask is right — both legs address `c.Model.For("incoming")`, so one card is correct for both.

Not booking a locally-refused `ErrNoPrefix` as a call is the right call for the reason given (no request leaves the process, and a 0 ms sample would pollute the mean the exploration brake reads), and filing #182 for a second opinion rather than deciding it unilaterally is the better handling of a published counter stepping down.

**Finding 2 (aggregate-only cost fallback) — fixed, and the fix is better than what I asked for.** I suggested a per-component fallback or an explicit marker; refusing to split the host's process-global figure across rows is the more honest answer, since apportioning it by call count would manufacture a number at the same precision as a measured one. `NetValueUSD *float64` rendering `null`, with `CostSource` naming which of the five cases a row is in, makes the unknown legible instead of arithmetically laundered. `unpricedCalls` replacing the `anySpend` boolean closes the silent under-report, and publishing `max(floor, host)` under a `partial`/`host_total` label is the right direction to fail in.

I checked for the hazard the pointer introduces: `NetValueUSD` has exactly one non-test consumer, `proxy/promexport.go:618`, and it reads through `Net()` rather than dereferencing, so an unknown omits the series instead of publishing `0` — which on that gauge would read as exactly break-even. No unguarded dereference anywhere in the tree.

**Findings 3 and 4 — handled.** The docs no longer produce the misread: `extract_llm.md` now leads with a danger admonition naming the sum, and `routes.md` documents `acted_fresh`/`acted_replay`, `by_component`, `cost_source` and `unpriced_components`. And the corrected rationale is *verifiably* the right one — I re-grepped: `measure.py:140` reads `runs`, `acted` and `saved_tokens` off the per-component objects by direct index, and nothing under `deploy/` touches any key in the extract block. Worth noting that the surviving harness claim on `cheapmodel.Usage()` (`usage.go:159`) is accurate — `measure.py:138` does index `llm_calls`, `llm_input_tokens` and `llm_output_tokens` — so that one is correctly left alone.

All four deferred items carry numbers (#179, #180, #181, #182) and I confirmed each is open and says what the commit claims it says.

### Checks, re-run independently on the eval box (Go 1.26.4, `CGO_ENABLED=1`)

`gofmt -l .` clean · `go vet ./...` clean · `go test ./...` green, no failures.

I also ran my own revert-verify of the headline fix rather than taking M8 on trust. Neutralizing the sink wiring (`ctx, sink :=` → `_, sink :=`, so the fallback's usage is never captured) and asserting the mutation landed, both new tests die with exactly the claimed output:

```
89:  the ledger row prices the fallback at $0 — the defect this test exists for
93:  the fallback's tokens did not reach the row: prompt=0 completion=0, want 31000/420
107: extract_llm_sweep.extraction_cost_usd moved by $0: the fallback's spend is still missing from /stats
111: cost_source = "unpriced", want "component"
151: Strategy = "prefix_ask", want "prefix_ask+fallback"
155: row tokens = prompt 40 / completion 90, want 31040/510 — the row still shows one leg
```

Restored, green again. The tests assert their preconditions (the no-asker path was taken, the fallback ran, a `ModelCall` exists) before their subject, so they cannot pass vacuously on a fixture that silently declined.

One minor item inline — a leftover of the same class as finding 1, latency-only. Nothing blocking.

components/offload/extract_sweep.go
```diff
@@ -592,7 +611,64 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
...
+		out, u, ms, err := e.fallbackAsk(ctx, req, c, &r, items, cands)
+		fbUsage, fbMs = u, ms
+		fbCost = recordLeg(ms, u)
+		return out, err
+	}
+	// foldFallback adds the fallback leg's tokens, dollars and wall time into the ask's ledger row.
+	// Idempotent by construction — at most one fallback runs per adjudication and this reads the
+	// accumulators rather than adding to them — so calling it on every path that can reach a
+	// fallback is safe.
+	foldFallback := func() {
```
amiddavid, line 669: **Minor, same class as the bug this closure fixes: `foldFallback()` is skipped on all three fallback *error* paths, so a failed fallback's wall time never reaches the ledger row.**

All three sites are `if reply, err = runFallback(); err != nil { return nil, r }` (lines 681, 727, 758) — the `foldFallback()` call sits after the early return. `runFallback` has already booked the leg into `metrics` via `recordLeg`, so `/stats` counts the call and its latency, while `r.rec` keeps only the prefix ask's `LatencyMs` and a `Strategy` of `"prefix_ask"` naming a leg that is no longer the only one that ran.

Scope, having checked both backends: on an error the sink is empty, because `recordUsageCache` is only reached after a successful decode on both paths (`anthropic.go:161`, `openai.go:120`) and neither returns an error after billing — an empty content list yields `"", nil`, not an error. So this is **latency-only, not dollars**, which is why it is minor rather than a repeat of the original finding. It still matters because the fallback is the slow leg by construction: a failed one contributes tens of seconds to `avg_latency_ms` while the per-call row the dashboard renders shows only the ask's milliseconds, and the row misnames what ran.

Site 681 is a slightly different shape — `r.rec` does not exist yet, so `rec.Component` is `""` and the row is dropped by the `call.rec.Component != ""` guard at line 473 entirely. `/stats` has a call the ledger has no row for. Not a regression (that path never built a row), but the spend recording is new, so the divergence is.

`defer foldFallback()` immediately after the accumulators are declared would cover all three: it is already documented as idempotent, it reads the accumulators rather than adding to them, and at site 681 it would populate the row that currently vanishes. The one thing to check if you take that route is that the deferred call must not run before `r.rec` is assigned on the happy path — it won't, since the defer fires at function exit.

amiddavid: Fixed in `7fed0d0`. Correct as diagnosed, including the latency-not-dollars reasoning — I re-checked and agree: the sink is empty on an error because `recordUsageCache` sits after a successful decode on both backends, and an empty content list yields `("", nil)` rather than an error, so there is no path that bills and then errors.

**I took `defer foldFallback()`, but it needed one correction that turns out to be the whole fix.**

It does not work against a local. Every early return in `adjudicate` is `return nil, r` on `var r sweepResult`, and a deferred mutation of a local happens *after* the return value has been copied — so the fold would compile, run, and be silently discarded on exactly the three error paths it exists to cover. The `sweepResult` is a **named result** now; the other result stays blank-named (`(_ []int, r sweepResult)`), because only this one needs the treatment and saying so beats leaving a reader to wonder what a second name is for.

Your check — "the deferred call must not run before `r.rec` is assigned on the happy path" — holds, and there is a second half to it worth stating: the happy path's `r.rec = components.ModelCall{…}` *overwrites* the deferred fold's work-in-progress, which is why the explicit call after that assignment stays. Re-folding is free, since `foldFallback` reads the accumulators rather than adding to them.

**Site 681's vanishing row is covered, and it needed more than the fold.** Deferring alone would have populated latency and tokens on a row whose `Component` was still `""` — so `Offload` would have dropped it on the `call.rec.Component != ""` guard exactly as before, and the fix would have run and changed nothing observable. `foldFallback` now establishes the identity fields (`Component`, `Model`, `CandidateTokens`, `GateReason`) when the row does not exist yet. That has its own mutant (M18) precisely because it was the half most available to leave half-done.

**Vacuity: three mutations, all three killed.**

- **M17** — revert to the per-site fold. No-asker: *"/stats counted a call and the ledger carries 0 rows"*. Ask-then-fail: `Strategy "prefix_ask"` and `LatencyMs 0` against ~8 ms.
- **M18** — revert the identity establishment. The no-asker row vanishes on the guard.
- **M19** — revert the named result. Fails identically to M17, with `defer foldFallback()` still sitting there running and doing nothing. This is the one I'd point at: had I applied the suggested shape literally, the diff would have looked like a fix, passed a casual read, and been inert on every path it targeted.

**On the assertion surface — I diverged, and here is why.** You asked for the rendered surface, and there isn't one for a `ModelCall` inside `/stats`: ledger rows travel to the dash `Event` (`dash/event.go` → `ExtractionRow`), not this payload, so a JSON-tag assertion would be testing `dash`'s converter rather than this fix. Instead I pinned what actually broke, which is that two surfaces **disagreed**: if `/stats` counts the call, the ledger must carry a row for it. Both tests assert `/stats` booked the leg *and* that `rep.Calls` has the row, so neither can pass while they diverge. And the assertion reads `rep.Calls` **after `Offload`** — past the guard that was doing the dropping — which is the same discipline as asserting on rendered output, by a different route. If you'd rather have the dash-side JSON assertion as well, say so and I'll add it in `dash`, but it would be pinning a different component's behaviour.

Both tests also assert the precondition that the fallback genuinely **errored** (`sweep_fallback_failed == 1`, plus the model was actually called) before asserting the row, so neither can pass on a fixture that silently declined. Fixturing the error was not awkward — an 8 ms sleep then `("", err)` gives a measurable failed leg.

`gofmt -l .` clean, `CGO_ENABLED=1 go vet ./...` clean, `CGO_ENABLED=1 go test ./...` green at 28 packages, on the eval box against the committed tree. No benchmark. Gaps table unchanged — no new issues; #179–#182 still the full set.

Issue #190, opened 2026-09-03
store: a saturated rewind reserve is released only by the TTL, so one busy period can zero out savings process-wide for ~2.8h


amiddavid:
## The remaining gap

`#188` bounded the rewind reserve properly — an entry cap, a byte budget (`stash_max_bytes`), and an evictable floor so the exemptions cannot consume the whole cache — but a slot is still released **only by the TTL**, and the store is a single process-wide instance (`cmd/context-guru-proxy/main.go` builds `cfg.NewStore()` once).

So a busy period can hold the reserve until entries age out. With the default sliding `ttl_seconds: 10000` that is up to ~2.8 h after the last touch, and while the reserve is saturated every new removal is refused across every session sharing the process. The failure mode is not broken markers — `#188` fixed that — it is **savings falling to zero mid-run**.

## What #188 deliberately did not do

Per-session partitioning was considered and declined. The reasoning, recorded so it can be revisited rather than re-litigated: no session ever reads another's payload (`offload.OwnsKey` enforces that), so what is shared is the *budget*, not the data — and a fair-share cap needs a share number as arbitrary as the `max_entries` default it would replace, plus an ownership rule for a content-hash key two sessions can legitimately both stash. Sizing against the real resource (bytes) was judged the more honest fix for the same failure.

That narrows the window; it does not close it.

## How it was found

Raised in the `#188` review (finding on `store.go:307`) and answered there as partially addressed. Filed so the open half survives the PR.

## The options, with their trade-offs

1. **Age-based last-resort eviction.** Under saturation, allow a payload past some age to be evicted, counted as an explicitly reported broken promise (a new counter, or `cg_stash_missing_total`). Restores liveness at the cost of reintroducing *bounded and measured* irreversibility. This is the direct opposite of `#188`'s thesis, which is why it needs a decision rather than a patch — but "bounded and reported" is a different thing from `#187`'s "unbounded and silent".
2. **Per-session share cap.** No session may hold more than a fraction of the reserve; a runaway session refuses for itself while others keep offloading. Bounds the blast radius without any irreversibility. Costs the two design problems named above.
3. **A shorter TTL for payloads specifically.** A payload's life is only needed while a marker referencing it can appear in a live request, which is a shorter horizon than a frozen decision's. Cheap, but shortens the window in which a long-idle session can still expand.
4. **Release on session end.** The honest version of (3), but there is no explicit session-end signal — sessions are derived per request — so it reduces to an idle timeout.

[...]

PR #188, opened 2026-09-03
fix(store): give rewind payloads their own reserve, and refuse a removal whose original cannot be stored

amiddavid:
Closes #187.

## The cause, established

**The issue's hypothesis is correct**, and the store's own comment said so before the fix did: *"The rewind stashes (bare content hashes, the large payloads the expand loop resolves) stay fully evictable."*

A stash payload's key **is the marker id** — a bare lowercase-hex content hash the model reads out of the request and hands back to `context_guru_expand` (`expand.Marker`, `offload.hashKey` / `extract.ContentKey`). It therefore carries none of the pinned `cg:*` prefixes, and it cannot be given one: marker bytes sit inside the provider's cached prefix, so prefixing the key would rewrite every marker. Meanwhile the **decision records that reference it** are pinned (`cg:res:`, `cg:xres:`, `cg:frz:`).

The arithmetic that turns that into 209 failed expands:

* **one store per process.** `cmd/context-guru-proxy` builds `cfg.NewStore()` once (`main.go:480`); every concurrent session shares it. The iteration 024 arm configs set no `store:` block, so: 1,000 entries, pin cap 500, `DefaultTTL` 10,000 s.
* **one reversible removal writes five entries** — the payload (unpinned), `cg:own:` (unpinned), `cg:xseen:` (unpinned), `cg:res:` (pinned), `cg:xres:` (pinned). The payload competed for ~1/5 of the evictable half while the two records naming it were exempt.

Measured directly against the shipped defaults of the day (600 removals writing that real key mix):

```
removals=600  stash_live=100  res_live=350  xres_live=350  own_live=100  xseen_live=100  frozen_dropped=500
```

**100 of 600 payloads survived against 350 of 600 of each pinned decision — the payload was 3.5x less durable than the record that pointed at it.** 500 markers had been stamped for content the store could no longer produce.

### The other two candidate causes are refuted

* **The stash expired.** `DefaultTTL` is 10,000 s and *sliding* (refreshed on `Get`, and components re-`Put` on every replay turn); the runs are shorter. The reproduction above evicts payloads with the clock stationary, so expiry is not needed to produce the symptom.
* **The store did not persist.** The process store is `*store.Memory` (`Persists() == true`), and `effectiveMode` already degrades a full marker to `off` when it is not — so a non-persisting store cannot leave a dangling marker. Arm A ran the same store and scored 0.
[...]

PR #194, opened 2026-09-03
fix(offload): let a replayed summary/off decision declare its loss, or the pipeline reverts it

### LINKED ISSUE #193: offload: a summary/off-mode offloader is reverted on every turn from the second onward, re-writing the provider's cached suffix each time
## The defect

Under `marker_mode: summary` or `off`, an offloader takes a deliberate **lossy** drop: nothing is stashed and no `<<cg:HASH>>` is written, so the component returns no cache keys. `components/pipeline.go:135` treats that combination — the request shrank, no cache keys, not `Skipped` — as a contract violation and **reverts the component**, unless `rep.Irreversible` marks the loss as chosen.

`commitMark`'s non-`full` branch sets that flag, so the turn that *takes* the decision is fine. Every later turn replays the frozen decision through `reapplyFrozen` (`components/offload/state.go`), **which never sets it**. Measured on a two-turn `mask` fixture at `marker_mode: summary`, same content both turns:

```
turn 1 (fresh, via commitMark):  keys=[] Irreversible=true   -> kept
turn 2 (replay, reapplyFrozen):  keys=[] Irreversible=false  -> REVERTED
```

So from turn 2 onward, for the rest of the session, the component is discarded and the transcript is forwarded verbatim.

## Why it is worse than a lost saving

Earlier turns sent the **reduced** bytes. Sending the original back re-writes the provider's whole cached suffix at ~11.5x the read price — and it repeats every turn, for every message the component had reduced. The direction is the cache-destructive one this repo goes to some length to avoid elsewhere (`reapplyFrozen` exists *because* flipping an already-cached message is expensive).

It is also invisible from the component's side: it acts, it computes a replacement, and the pipeline throws the work away afterwards. Nothing counts a revert per component, so a summary-mode deployment looks like one that is simply saving less than expected.

## Affected components

Every offloader that freezes and replays — `mask`, `collapse`, `failed_run`, `readlifecycle`, `skeleton`, `cmdfilter`, `agentdiet` — whenever configured with `marker_mode: summary` or `off`. The default is `full`, which is unaffected: a full-mode frozen replacement always carries a marker, so `keys` is non-empty and the revert precondition never holds.

## How it was found

Auditing replay paths while addressing a review on #188, which flagged the same omission in two *new* replay branches introduced there. This instance is independent of that work and predates it — at the merge base (`51fcd91`) `reapplyFrozen` takes no `rep` parameter at all — which is why it is filed and fixed separately rather than folded into #188.

## The fix

`reapplyFrozen` takes `rep` and sets `rep.Irreversible` when the replayed replacement carries no markers. `len(keys) == 0` is a sound test for "degraded mode": every `freeze()` site is downstream of a `tryMark`/`commitMark` pair, so a full-mode frozen replacement always carries a marker — meaning the flag cannot mask a full-mode bug.

Threading `rep` through the seven call sites is mechanical. PR to follow, off `main`.

## Alternatives considered

* **Have `pipeline.go` treat "no keys and the replacement is a pure function of the original" as exempt.** Rejected: the pipeline cannot know that, and widening its exemption weakens the guard that catches a component which genuinely forgot to stash.
* **Record the marker mode on the frozen entry and consult it on replay.** More explicit, but it changes the store's value format for every frozen key and buys nothing over the marker test above.

amiddavid:
Closes #193.

## The defect

Under `marker_mode: summary` or `off`, an offloader takes a deliberate **lossy** drop: nothing is stashed and no `<<cg:HASH>>` is written, so it returns no cache keys. `components/pipeline.go:135` treats that combination — the request shrank, no cache keys, not `Skipped` — as a contract violation and **reverts the component**, unless `rep.Irreversible` says the loss was chosen.

`commitMark`'s non-`full` branch sets that flag, so the turn that *takes* the decision is fine. Every later turn replays it through `reapplyFrozen`, which never did. Measured on a two-turn `mask` fixture at `marker_mode: summary`, same content both turns:

```
turn 1 (fresh, via commitMark):  keys=[] Irreversible=true   -> kept
turn 2 (replay, reapplyFrozen):  keys=[] Irreversible=false  -> REVERTED
```

So from turn 2 onward, for the rest of the session, the component is discarded and the transcript is forwarded verbatim.

## Why it is worse than a lost saving

Earlier turns sent the **reduced** bytes. Sending the original back re-writes the provider's whole cached suffix at ~11.5x the read price — every turn, for every message the component had reduced. That is the cache-destructive direction `reapplyFrozen` exists to avoid, reached *through* `reapplyFrozen`.

It is also invisible from the component's side: it acts, it computes a replacement, and the pipeline discards the work afterwards. Nothing counts reverts per component, so a summary-mode deployment reads as one that is simply saving less than hoped.

## Scope

Affects `mask`, `collapse`, `failed_run`, `readlifecycle`, `skeleton`, `cmdfilter` and `agentdiet` whenever configured with `marker_mode: summary` or `off`. **The default (`full`) is unaffected** — a full-mode frozen replacement always carries a marker, so `keys` is non-empty and the precondition never holds.

## The fix

`reapplyFrozen` takes `rep` and sets `rep.Irreversible` when the replayed replacement carries no markers. Threading `rep` through the seven call sites is mechanical.

`len(keys) == 0` is a sound test for "degraded mode" rather than a proxy for it: every `freeze()` site is downstream of a `tryMark`/`commitMark` pair, so a full-mode frozen replacement always carries a marker — which means the blanket flag **cannot mask a full-mode bug**. That was the property worth checking before choosing a blanket flag over threading the mode through the store.

### Alternatives considered

* **Widen `pipeline.go`'s exemption** to "no keys, but the replacement is a pure function of the original". Rejected: the pipeline cannot know that, and it weakens the guard that catches a component which genuinely forgot to stash.
* **Record the marker mode on the frozen entry** and consult it on replay. More explicit, but it changes the store's value format for every frozen key and buys nothing over the marker test.

## Why this is off `main` and not part of #188

Found while auditing replay paths for the #188 review, which flagged the same omission in two **new** replay branches introduced there. This instance is independent and predates that work — at the merge base (`51fcd91`) `reapplyFrozen` takes no `rep` parameter at all.

#188 still carries the open reserve-lifetime question in #190, so folding this in would mean a bug that is live on `main` today ships only when that lands. #188 contains the same three lines and will be rebased onto this once it merges; the reviewer on #188 independently reached the same conclusion about separating it.

## Verification

`gofmt -l .` clean · `CGO_ENABLED=1 go vet ./...` clean · `CGO_ENABLED=1 go test ./...` all packages pass (Go 1.26.4, eval box).

`TestASummaryModeReplayIsNotRevertedFromTurnTwoOnward` asserts turn 2's exact conjunction. Turn 1 is asserted **first**, so a fixture that stopped reaching summary mode fails there rather than passing vacuously below — and the test also requires that turn 2 actually rewrote the message and did not report `Skipped`, since either would make the revert precondition unreachable.

Verified by reverting its subject:

```
a replayed summary-mode decision rewrote the message, returned NO cache key and did not
set rep.Irreversible. That is exactly what components/pipeline.go reverts as "offload
dropped content without stashing a cache_key" ...
```

### DIFF
--- a/components/offload/agentdiet.go
+++ b/components/offload/agentdiet.go
@@ -399,7 +399,7 @@ func (d *AgentDiet) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 			if !schema.Rewritable(*msg) {
 				continue
 			}
-			if fk, saved, ok := reapplyFrozen(c, d.Name(), msg); ok {
+			if fk, saved, ok := reapplyFrozen(c, rep, d.Name(), msg); ok {
 				rep.TokensBefore += saved // best-effort; the pipeline recomputes exactly
 				keys = append(keys, fk...)
 				changed++
--- a/components/offload/cmdfilter.go
+++ b/components/offload/cmdfilter.go
@@ -140,7 +140,7 @@ func (f *Cmdfilter) Offload(req *schemas.BifrostChatRequest, rep *components.Rep
 		// reapplyFrozen also re-Puts the stashed original for every marker in the
 		// replacement, so the expand loop keeps working across turns, and it declines
 		// content the agent has expanded (kept-verbatim).
-		if fk, _, ok := reapplyFrozen(c, f.Name(), m); ok {
+		if fk, _, ok := reapplyFrozen(c, rep, f.Name(), m); ok {
 			changed++
 			keys = append(keys, fk...)
 			continue
--- a/components/offload/collapse.go
+++ b/components/offload/collapse.go
@@ -110,7 +110,7 @@ func (cl *Collapse) Offload(req *schemas.BifrostChatRequest, rep *components.Rep
 		// on purpose — with max_frac set, CtxWindow can resolve differently mid-session
 		// (model swap, refreshed modelinfo), and a threshold that drifts above this output
 		// would otherwise flip it collapsed→full inside the cached prefix.
-		if fk, _, ok := reapplyFrozen(c, cl.Name(), m); ok {
+		if fk, _, ok := reapplyFrozen(c, rep, cl.Name(), m); ok {
 			changed++
 			keys = append(keys, fk...)
 			continue
--- a/components/offload/failed_run.go
+++ b/components/offload/failed_run.go
@@ -120,7 +120,7 @@ func (fr *FailedRun) Offload(req *schemas.BifrostChatRequest, rep *components.Re
 		// Reapply a previously-frozen collapse on EVERY turn (cache-stable), regardless
 		// of the tail boundary — the agent re-sends the original, so we must re-collapse
 		// it to the same bytes or it reverts to full and churns the cache.
-		if fk, _, ok := reapplyFrozen(c, fr.Name(), m); ok {
+		if fk, _, ok := reapplyFrozen(c, rep, fr.Name(), m); ok {
 			changed++
 			keys = append(keys, fk...)
 			continue
--- a/components/offload/freeze_test.go
+++ b/components/offload/freeze_test.go
@@ -142,10 +142,11 @@ func TestFrozenCountersMove(t *testing.T) {
 	st := store.NewMemory(store.Options{})
 	c := &components.Ctx{Session: "sCount", Store: st}
 	msg := tool("some tool output")
-	reapplyFrozen(c, "mask", &msg) // miss: nothing frozen yet
+	var rep components.Report
+	reapplyFrozen(c, &rep, "mask", &msg) // miss: nothing frozen yet
 	freeze(c, "mask", "some tool output", "short")
 	msg2 := tool("some tool output")
-	reapplyFrozen(c, "mask", &msg2) // hit
+	reapplyFrozen(c, &rep, "mask", &msg2) // hit
 	h1, m1 := FrozenStats()
 	if h1 <= h0 || m1 <= m0 {
 		t.Fatalf("hits/misses must both advance: %d->%d, %d->%d", h0, h1, m0, m1)
--- a/components/offload/mask.go
+++ b/components/offload/mask.go
@@ -77,7 +77,7 @@ func (m *Mask) Offload(req *bschemas.BifrostChatRequest, rep *components.Report,
 		// the tail boundary: the agent re-sends the original, so we must re-mask it to the
 		// same bytes or it reverts full→masked→full and churns the provider KV cache. This
 		// also skips kept-verbatim content (see reapplyFrozen).
-		if fk, _, ok := reapplyFrozen(c, m.Name(), msg); ok {
+		if fk, _, ok := reapplyFrozen(c, rep, m.Name(), msg); ok {
 			changed++
 			keys = append(keys, fk...)
 			continue
--- a/components/offload/readlifecycle.go
+++ b/components/offload/readlifecycle.go
@@ -157,7 +157,7 @@ func (rl *ReadLifecycle) Offload(req *bschemas.BifrostChatRequest, rep *componen
 		// Replay a frozen decision on EVERY turn, at any depth: the agent re-sends the
 		// original each turn, so not re-offloading it would flip the message
 		// offloaded→full→offloaded and churn the provider's KV cache.
-		if fk, _, ok := reapplyFrozen(c, rl.Name(), msg); ok {
+		if fk, _, ok := reapplyFrozen(c, rep, rl.Name(), msg); ok {
 			changed++
 			keys = append(keys, fk...)
 			continue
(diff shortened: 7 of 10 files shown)


amiddavid:
Context for reviewers: this is split out of #188 on that PR's review recommendation. #188 currently contains the same three lines and will be rebased onto `main` once this lands, at which point its `state.go` diff reduces to a single call-site change. The defect itself is independent of #188 and predates it — verified at the merge base `51fcd91`, where `reapplyFrozen` takes no `rep` parameter at all. Filed as #193.

amiddavid (commented):
Reviewed and verified independently, on the box (Go 1.26.4, `CGO_ENABLED=1`). This is the right shape for the split, and I found nothing to fix. (Recording as a comment rather than an approval only because GitHub will not let this account approve its own PR.)

**The split is clean.** Based on current `origin/main`, and the diff is exactly what a main-defect fix should be: 22 lines in `state.go` of which 3 are logic, 7 mechanical call-site updates threading `rep`, one existing test adjusted for the signature, and one new test. No reserve work rode along.

**Verified:**

- `gofmt -l .` clean, `go vet ./...` clean, `go test ./...` clean — no failures anywhere, `dash` included.
- `TestASummaryModeReplayIsNotRevertedFromTurnTwoOnward` passes, and **revert-verified**: reverting the fix to `_ = rep` compiles clean (`go vet` silent) and the named test fails. So the mutation is a real revert of the bug, not a build break — and `go test -run` resolved the name (`=== RUN` emitted), so it is also not the "no tests to run exits 0" trap.
- The blanket `rep.Irreversible` cannot mask a full-mode bug: every `freeze()` site is downstream of a `tryMark`/`commitMark` pair, so a full-mode frozen replacement always carries a marker and `len(keys) == 0` genuinely implies a degraded mode. The comment states this; I checked all 7 sites.
- All 7 `reapplyFrozen` callers updated.
- Confirmed the defect predates #188: at the merge base `reapplyFrozen` took no `rep` parameter at all and did a bare `c.Store.Put(k, ...)`, so summary-mode replacements have been reverted from turn 2 onward independently of any reserve work.

**On sequencing:** keeping the same lines in #188 until this lands, then rebasing, is the right call — stripping them pre-emptively would make #188 red for reasons unrelated to its own review. The `state.go` conflict between the two branches is expected; noting it here so it is on the record rather than looking like drift later.

The comment carries the measured evidence (turn 1 `Irreversible=true`, turn 2 `false` with the message rewritten and no keys) and explains why the pipeline reverts that combination, which is what a reader needs to not reintroduce it.

<sub>Generated with [Claude Code](https://claude.com/claude-code)</sub>

amiddavid:
Thank you for the independent verification, and particularly for revert-verifying it yourself rather than taking my word for it — including checking that the mutation *compiles* and that `go test -run` actually resolved the name. Those are the two traps that produced false evidence three times over on #188, so having them checked from the other side is worth more than the fix.

Nothing to change here. Recording what happens next so the sequencing is on the record:

* This is ready as far as I can tell. It needs a human to merge, since neither of us can approve it.
* Once it lands I rebase **#188** onto `main`, at which point #188's `state.go` diff reduces to the `commitRefresh` call alone and the conflict between the two branches disappears.
* I have not touched #188's copy of these lines in the meantime, for the reason you agreed with: stripping them would make that branch red for reasons unrelated to its own review.

One point of order on your last paragraph. You describe the comment as carrying "the measured evidence (turn 1 `Irreversible=true`, turn 2 `false`…)" — worth being precise that the numbers in the commit message and PR body are from **my** two-turn probe on a `mask` fixture, and yours independently reproduced the same shape. Two runs agreeing is stronger than one, and it should read that way rather than as a single measurement quoted twice.

amiddavid:
**Taking the correction — and it strengthens the record rather than weakening it, so worth stating precisely.**

You're right that the numbers quoted in the `state.go` comment are from your probe, and my write-up above should have said so. What I actually did was write my own two-turn `markerSummary` fixture against `extract_llm` during the round-2 review, before this split existed, and run it on `a67c597`; it moved from `turn 2 irreversible=false` to `turn 2 irreversible=true`. Then for this PR I revert-verified your test independently — mutating the fix to `_ = rep`, confirming `go vet` stayed silent, and confirming the named test resolved and failed.

So the evidence for this defect is **two independently written probes, on different fixtures, agreeing on the shape** — yours on `mask` via `reapplyFrozen`, mine on `extract_llm`'s replay branch — plus a mutation showing the fix is what makes the difference. That is a materially stronger claim than one measurement quoted twice, and it is the reason I'd treat this defect as settled rather than merely plausible.

Recording it that way here so the PR carries the accurate provenance.

PR #196, opened 2026-09-03
feat(kvcache): measure the idle keep-alive per ping — cache outcome, gaps and cost

itay-nakash:
## What this adds

Two files under `deploy/harbor/`, alongside the existing `kv_ttl_*` analysis scripts:

- **`keepalive_ping_stats.py`** — produces one row per idle keep-alive ping over a window of
  history, answering four questions per ping: did the ping itself hit the provider cache, how
  many seconds after the previous request did it fire, did the request that followed it hit and
  how long after, and what did the ping cost. Plus a summariser over that table.
- **`test_keepalive_ping_stats.py`** — 14 tests against a synthetic `requests` table. No
  production access, no network.

Nothing existing is modified. No Go code, no schema change, no runtime path is touched — this is
read-only analysis of rows the service already writes.

## Why

`proxy/keepalive.go` carries a detailed account of what the mechanism is worth, sourced from a
replay (`TestKeepAliveOnProductionSnapshot`). What we did not have was a way to ask the same
questions of *what actually shipped*, per ping, against the live ledger — in particular whether a
given ping read the entry it was meant to refresh, and whether the request it was protecting then
hit. The dashboard reports pings in aggregate (count and cost); `keepalive_saved_usd` reports the
service's own attribution. Neither answers "did this ping work, and how close was it".

## Identifying our own pings

`requests.keepalive = 1`. This is exact in one direction and a lower bound in the other, and the
distinction is documented at the top of the script because it changes how a count is read:

- **No false positives.** `KeepAlive: true` is set in exactly one non-test place in the service,
  `proxy/keepalive.go` (`record1`), and nothing ever `UPDATE`s the column afterwards — the
  strategy backfill writes `keepalive_strategy_id` only. No benchmark, simulation or campaign
  path writes a ping row. Rows predating the mechanism read `0` by the `ALTER TABLE` default, and
  correctly so: the column and the emitter shipped in the same commit (#86).
- **False negatives, all of them outside the table.** A ping refused by the tenant rate limiter,
  failed in transport, or answered 4xx returns before `record1`, so it never becomes a row. The
  table holds pings that were sent *and* answered non-4xx; `KeepAliveStats` on `/stats` holds what
  was attempted.

The collector does not take the flag on trust. It also evaluates an independent structural
fingerprint of the row shape `record1` produces — `max_tokens=1`, `stream=0`, and zero for
`messages`, `tokens_before`, `mode`, `cache_miss_reason`, `cg_latency_ms`, `ttfb_ms` — in both
directions, and prints any disagreement as a number. Over 1,038 production pings it reported
`0` flagged rows whose shape disagrees and `0` unflagged rows shaped like a ping.

## Two correctness traps this had to handle

Both were found by running against real rows, and both silently invert a headline number.

**1. `ts` means different things on the two row types.** A real request's `ts` is its *start*
(`proxy/dashcapture.go`, `TS: c.start.UnixMilli()`); a ping's is stamped in `record1`, *after* the
round trip. Every gap here is therefore start-to-start, which is also the footing
`CachePolicy.Idle` and the provider's lifetime rule both use. It matters: corrected, `gap_before_s`
clusters at 270–272s on a 270s-Idle account with a 1.6s spread; uncorrected it scatters over
273–288s, the noise being the ping's own round trip, which reaches 28s at p95.

**2. A non-2xx row must not be taken as "the request after the ping".** In production, a ping is
frequently followed within about a second by a status-400 row carrying a single message and no
tokens, with the actual resumption arriving next. Those rows still receive a `cache_miss_reason`
from `AttributeCache`. On the first sample checked, taking the immediate next row reported 0 of 4
following requests hitting; the truth was 2 of 4, both reading the ping's prefix back token for
token. The collector steps over non-2xx rows to find the neighbouring agent request, keeps the
immediate neighbour's id alongside it, and counts what it stepped over.

## How to use it

The two subcommands split because the database is readable only by the account the service runs
as, and that account has no site-packages:

```sh
# 1. collect -- stdlib only, runs as the service account. One CSV row per ping.
sudo -u cg /usr/bin/python3 keepalive_ping_stats.py collect --out /var/tmp/pings.csv

# 2. stats -- loads that CSV into a DataFrame and prints the tables. Needs pandas.
python3 keepalive_ping_stats.py stats /var/tmp/pings.csv
```

`WHERE keepalive = 1` has no supporting index, so an unbounded run is a full scan of the live
file. Pass `--since-ms <epoch ms>` to bound it, which lets the query use `idx_requests_ts`.

There is a second source that needs no privileged access, and it is how the script was developed
and validated: a session export bundle, one directory per session each holding a `turns.csv` that
is `SELECT * FROM requests`.

```sh
python3 keepalive_ping_stats.py collect --export <bundle.zip|dir> --out pings.csv
```

Account ids are pseudonymised (`t01`, `t02`, …) and session ids hashed unless `--raw-ids` is
passed. No message content is read — `request_content` is never queried.

Tests:

```sh
python3 deploy/harbor/test_keepalive_ping_stats.py     # or run under pytest
```

## How to read the results

`collect` prints three lines to stderr: how many pings it found, the identification audit in both
directions, and the window the table covers versus the first recorded ping — so a count is read
against the history that could have contained pings rather than against "all time".

`stats` prints four blocks:

- **AVERAGES.** Watch the `n` column; it differs per row. `gap_before_s` is the configured `Idle`
  recovered from data. `gap_before_prev_real_s` skips back past intervening pings, so on a second
  ping it measures the whole span rather than the inter-ping step. `gap_after_next_real_s` is the
  one that answers "when did the human come back". `ping_cost_usd` is heavily right-skewed, so
  read the median next to the mean.
- **WHAT THE PING DID** — four *disjoint* outcomes with spend against each. A ping can read
  something and still write more than it read, so "wrote instead of read" is not a subset of
  "missed"; they are printed as separate buckets for that reason. `wrote a new entry, read
  nothing` is the failure `keepalive.go` calls the mechanism being wrong rather than expensive.
- **THE REQUEST AFTER THE PING.** The `HIT` line is what happened; the `credited` line is the
  service's own attribution, which a non-2xx row arriving first can consume. Expect them to
  disagree. The `cache_miss_reason` breakdown separates "came back too late" (`ttl_expiry`) from
  "the prefix moved" (`prefix_change`) from "a different model" (`cold_start`, which with the
  session already seen means `!seenModel` and is not addressable by pinging).
- **BY PING POSITION**, then a **timing self-check**. The self-check is the cheap evidence that
  identification and the `ts` correction are both right: the share of gaps within ±12s of the
  median is the band a 2s sweep firing one tick after `Idle` produces. It also names its two
  tails, because both have a known cause — a negative gap is a ping that was still in flight when
  a request arrived, and a gap near 2× the median is one whose preceding attempt advanced the
  timer without writing a row.

## Evidence

`14/14` tests pass. Run against the live ledger for 2026-08-23..09-02: 1,038 recorded pings across
255 sessions and 15 accounts, `$150.87` of ping spend, identification audit clean in both
directions, and `91.8%` of gaps inside the ±12s band. The same code over a session export bundle
reproduces its rows identically.

CI does not currently run Python, so `test_keepalive_ping_stats.py` is a manual gate; it is
stdlib-only apart from the two summariser tests, so it runs anywhere the collector does.

### DIFF
--- a/deploy/harbor/keepalive_ping_stats.py
+++ b/deploy/harbor/keepalive_ping_stats.py
@@ -1,0 +1,456 @@
+#!/usr/bin/env python3
+"""Per-ping forensics for the idle keep-alive: did OUR ping hit the provider cache, how
+long after the previous request did it fire, did the request that followed it hit, and
+what did the ping cost.
+
+Two subcommands, and the split is not stylistic: the service database is readable only by the
+account the service runs as, and that account has no site-packages. So the half that reads it
+uses nothing but the standard library, and the half that wants a DataFrame runs afterwards as
+whoever has pandas.
+
+  collect   stdlib only (sqlite3 + csv). Reads the dashboard database READ-ONLY, so it runs as
+            the service account. Writes one CSV row per ping. Also accepts a session export
+            bundle instead, which needs no privileged access at all.
+  stats     pandas. Loads that CSV into a DataFrame and prints the tables.
+
+A ping is identified by requests.keepalive = 1. That is exact in one direction and a lower
+bound in the other, and the difference matters when reading a count off this script:
+
+  No false positives. `KeepAlive: true` is set in exactly ONE non-test place in the service,
+  proxy/keepalive.go:964 (record1), and nothing ever UPDATEs the column afterwards — the
+  strategy backfill writes keepalive_strategy_id only (dash/keepalivestrategybackfill.go:151).
+  No benchmark, simulation or campaign path writes a ping row. Rows older than the mechanism
+  read 0 by the ALTER TABLE default, and correctly so: the column and the emitter shipped in
+  the same commit (50e3966, #86), so there is no window of pings that predates the flag.
+
+  False negatives, all of them OUTSIDE the table. A ping that never got a row: refused by the
+  tenant rate limiter, failed in transport, or answered 4xx — that last one returns before
+  record1 (proxy/keepalive.go:906) so a REFUSED ping is invisible here. The table therefore
+  holds pings that were sent AND answered non-4xx; the process counters in KeepAliveStats
+  (/stats) hold what was attempted. Read "pings" below as recorded pings, not as pings sent.
+  Pings answered 5xx ARE recorded, with status >= 500 and no usage — the status line in the
+  summary is where they show up.
+
+One caveat about the service's OWN attribution columns, which this script carries but does not
+compute: keepalive_pings is an in-process counter consumed by keeper.arrive() during the
+PREPARATION of the next request on the session (proxy/proxy.go:1109), before that request is
+sent upstream and therefore before its status is known. So the first row to arrive takes the
+credit whatever becomes of it, and keepalive_saved_usd is only nonzero on a row that read more
+than it wrote. In the production export checked here, the row that arrived first was a 400 every
+time, so the credit landed on a row that read nothing and $0.00 was attributed — while the real
+resumption a second later read the ping's prefix back token for token. Read
+next_real_keepalive_saved_usd as the service's own attribution, and next_real_cache_hit together
+with next_real_cache_read as what actually happened.
+
+`collect` does not trust the flag alone: it also evaluates an INDEPENDENT structural
+fingerprint of the row shape record1 produces, in both directions, so a mislabelled row shows
+up as a number rather than as a silently wrong average.
+
+No message content is read — request_content is never touched. Tenant ids are
+pseudonymized and session ids hashed unless --raw-ids is passed.
+"""
+import argparse, csv, glob, hashlib, io, os, sqlite3, sys, tempfile, zipfile
+
+DB = "/var/lib/context-guru/cg.db"
+
+# Columns pulled for every row of a ping-bearing session. Enough to compute the four
+# metrics, and enough to audit one suspicious row without going back to the database.
+COLS = [
+    "id", "ts", "tenant_id", "session_id", "keepalive", "model", "agent", "status",
+    "cache_read", "cache_write", "cache_write_1h", "fresh_input", "output_tokens",
+    "cost_usd", "upstream_ms", "cache_miss_reason", "keepalive_pings",
+    "keepalive_saved_usd", "keepalive_strategy_id",
+    # The independent fingerprint. record1 sets max_tokens=1 and leaves every one of the
+    # rest at its zero value, because a ping has no agent request behind it to fill them.
+    "max_tokens", "stream", "messages", "tokens_before", "mode", "cg_latency_ms", "ttfb_ms",
+]
+
+SQL = f"""
+WITH ka(tenant_id, session_id) AS (
+  SELECT DISTINCT tenant_id, session_id FROM requests
+  WHERE keepalive = 1 AND session_id <> '' AND ts >= ?
+)
+SELECT {', '.join('r.' + c for c in COLS)}
+FROM requests r JOIN ka ON r.tenant_id = ka.tenant_id AND r.session_id = ka.session_id
+WHERE r.ts >= ?
+ORDER BY r.tenant_id, r.session_id, r.ts, r.id
+"""
+
+OUT_COLS = [
+    # identity / provenance
+    "ping_id", "tenant", "session", "ping_ts_ms", "ping_seq", "pings_in_span",
+    "model", "agent", "status", "strategy_id",
+    # 1. did the ping hit, and 4. what did it cost
+    "ping_cache_hit", "ping_cache_read", "ping_cache_write", "ping_cache_write_1h",
+    "ping_fresh_input", "ping_output_tokens", "ping_wrote_more_than_read",
+    "ping_cost_usd", "ping_cost_unpriced", "ping_upstream_ms",
+    # 2. how long after the message before it
+    "gap_before_s", "gap_before_raw_s", "prev_id", "prev_is_ping", "prev_ts_ms",
+    "gap_before_prev_real_s", "prev_real_id", "prev_real_cache_read", "prev_real_cache_write",
+    # 3. the message after: did it hit, how long after
+    "gap_after_s", "next_id", "next_is_ping", "next_ts_ms",
+    "gap_after_next_real_s", "next_real_id", "next_real_cache_hit",
+    "next_real_cache_read", "next_real_cache_write", "next_real_cache_miss_reason",
+    "next_real_cost_usd", "next_real_keepalive_pings", "next_real_keepalive_saved_usd",
+    "next_real_credited", "nonok_rows_before", "nonok_rows_after",
+    # data quality
+    "fingerprint_ok", "no_prev_row", "no_next_row",
+]
+
+
+def fingerprint(r):
+    """Is this row shaped like something record1 wrote, judged without the keepalive flag?
+
+    record1 fills the ping's own tokens, cost, status and max_tokens=1 and leaves
+    everything an agent request would carry at zero: no messages, no pre-compaction token
+    count, no operating mode, no cache-miss attribution (a ping is deliberately excluded
+    from it), no in-process latency. A real /v1/messages call cannot have zero messages.
+    """
+    return (r["max_tokens"] == 1 and r["stream"] == 0 and r["messages"] == 0
+            and r["tokens_before"] == 0 and r["mode"] == ""
+            and r["cache_miss_reason"] == "" and r["output_tokens"] <= 1
+            and r["cg_latency_ms"] == 0 and r["ttfb_ms"] == 0)
+
+
+def start_ms(r):
+    """When this request went on the wire, for a row of either kind.
+
+    The two writers stamp ts differently: a real request carries its START
+    (proxy/dashcapture.go, TS: c.start.UnixMilli()) while a ping carries the moment
+    record1 booked it, which is AFTER the round trip. Every gap here is start-to-start,
+    because that is the footing the provider's cache lifetime and the keeper's own Idle
+    are both measured on — "how long a session must be idle before the first ping,
+    measured from the previous request's START" (CachePolicy.Idle). gap_before_raw_s keeps
+    the uncorrected ts-to-ts figure for anyone reproducing this from the raw columns.
+    """
+    return r["ts"] - r["upstream_ms"] if r["keepalive"] else r["ts"]
+
+
+def sessions(rows):
+    """Group the ordered rows into sessions, yielding one list per (tenant, session)."""
+    cur, key = [], None
+    for r in rows:
+        k = (r["tenant_id"], r["session_id"])
+        if k != key and cur:
+            yield cur
+            cur = []
+        key = k
+        cur.append(r)
+    if cur:
+        yield cur
+
+
+def spans(rows):
+    """Positions of each maximal run of consecutive ping rows, as (start, length).
+
+    A run between two real requests IS one idle span: the keeper resets its per-span
+    counter when a real request re-records the session, so a run's length is K-in-practice
+    and a ping's position in the run is its ping number.
+    """
+    out, i = [], 0
+    while i < len(rows):
+        if rows[i]["keepalive"]:
+            j = i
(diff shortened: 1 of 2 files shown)


OsherElhadad:
## Review: audited the docstring against the Go, mutated the script, and probed it with synthetic data only

Reviewed at `0a894ac8` in a dedicated worktree. **Verdict: approve with nits** — one security finding worth fixing before merge, two stale code citations, and a small pile of nits. The analysis logic is sound and the test suite genuinely catches the defects that would silently invert a headline number.

**Data boundary:** this script reads the production service database (`DB = /var/lib/context-guru/cg.db`). Nothing in this review touched it, `/etc/context-guru/**`, or any existing export. Everything below was checked either by reading Go source or against synthetic SQLite databases and zip bundles built for the purpose and deleted afterwards. The production figures in the body are listed at the end as not verifiable within that boundary — they are not disputed, just not re-derived.

### CI does not run this PR

`.github/workflows/ci.yaml` has three jobs — `build-test`, `purego`, `trivy` — and none touches Python. The only `setup-python` anywhere is `docs.yaml` for the mkdocs build. **This diff is 100% Python, so the five green checks are evidence about the Go build and say nothing about it.** The body already says as much, which is the right call; recording it here so no one reads the green ticks as coverage.

Worth contrasting with #212, which faces the same problem and solves it: its Python is pinned by *Go* drift tests under `deploy/harbor/`, which `make cover` does run. If this script ever grows a claim worth defending in CI, that is the pattern available.

### Real run results

| | |
|---|---|
| Merge onto current `main` (`87377e73`) | **clean, zero conflicts** — 20 commits behind, but `mergeable: UNKNOWN` was a caching artifact |
| `python3 test_keepalive_ping_stats.py` (no pandas, 3.9.25) | `Ran 14 tests` — **11 pass, 3 ERROR** with a clean `ModuleNotFoundError` |
| `uv run --with pandas python3 test_keepalive_ping_stats.py -v` | `Ran 14 tests … OK` — **14/14** |
| pandas genuinely importable | verified at symbol level (`from pandas import DataFrame`), not just `import pandas` |

**Zero tests silently skipped** — the pandas-dependent ones fail loudly rather than skipping, which is the honest failure mode. (Checked specifically because an unreadable root-owned `site-packages` can import as an empty namespace package, making `import pandas` succeed with `__file__ = None`; that hazard does not manifest here.)

### Docstring-claim audit — the script's correctness argument, checked against the Go

The docstring is written to be checkable, so it was checked. Every substantive claim holds; two citations do not.

| Claim | Cited | Status |
|---|---|---|
| `KeepAlive: true` set in exactly one non-test place | `proxy/keepalive.go:964` (`record1`) | **exact**, at merge-base and `main` |
| Backfill writes `keepalive_strategy_id` only, never `keepalive` | `dash/keepalivestrategybackfill.go:151` | **exact**; no other writer of the column repo-wide |
| Column and emitter shipped in one commit, so no pings predate the flag | `50e3966` (#86) | **confirmed** |
| A ping's `ts` is stamped after the round trip; a real request's is its start | `dashcapture.go` `TS: c.start.UnixMilli()` | **confirmed** |
| `keepalive_saved_usd` nonzero only when the row read more than it wrote | — | **confirmed** (`dash/event.go:823-830`) |
| `AttributeCache` not computed for a ping | — | **confirmed, and structurally stronger than claimed** — `record1` builds its own `dash.Event` and never enters the path that calls it |
| 4xx returns before `record1`, so a refused ping is invisible | `proxy/keepalive.go:906` | **stale — it is `:908`** |
| `keeper.arrive()` consumes `keepalive_pings` before the next request's status is known | `proxy/proxy.go:1109` | **wrong — it is `:1143`** |

**Minor · docs · both citations were wrong when written, not drifted.** I checked the two files at the PR's own merge-base (`51fcd916`) as well as at `main`: neither was touched by the 20 intervening commits, and `keeper.arrive(` is at **1143 in both**, with line 1109 in both landing inside an unrelated comment block about repair/tool-injection deferral. A reader following `:1109` to check the attribution claim finds nothing related to it and may reasonably conclude the claim is unfounded — when it is in fact correct. Please re-grep the symbols and fix both numbers; the underlying claims stand.

### Mutation testing — four defects introduced in the script, one at a time

| Defect | Caught by |
|---|---|
| `start_ms()` drops the `ts` correction (returns raw `ts` for a ping) | **5 tests** |
| Fingerprint check made one-directional (`unflagged` hardcoded to 0) | **1 test** |
| Trust the `keepalive` flag alone (`fingerprint()` always `True`) | **the same 1 test** |
| Mis-attribute cost/outcome to the immediate next row, bypassing the non-2xx step-over | **2 tests** |

**Nothing went uncaught.** But note the middle two: the "independent structural fingerprint, evaluated in both directions" safeguard — which the body calls out as a specific correctness guarantee — has **exactly one test defending it** (`test_identification_audit_counts_both_directions`). That test is well built (its 5-row fixture carries one flagged-but-not-shaped row *and* one shaped-but-not-flagged row, so it is not a cardinality-1 fixture restating totals). Still, delete that one test and both directions go dark at once.

### Findings

**Minor, and I'd fix it before merge · `deploy/harbor/keepalive_ping_stats.py:322`** — `os.chmod(args.out, 0o644)` runs **unconditionally, including under `--raw-ids`**, when the CSV carries real `tenant_id` plus per-row dollar figures (`ping_cost_usd`, `next_real_cost_usd`).

This repo has already decided this exact question, in the sibling script, with a comment saying why:

```python
# kv_ttl_cost_model.py:1750-1756
# 0600, and created that way rather than chmod'd afterwards: the payload carries
# by_user keyed by TENANT ID with each account's request count and dollar total, so a
# world-readable default (0644 minus umask) hands per-customer spend to every local
# user. O_CREAT|O_TRUNC with the mode set at creation leaves no window where the file
# exists with wider permissions.
fd = os.open(args.json, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
```

The sharp version: this is not a missing restriction, it is an **active widening** — `open(args.out, "w")` would inherit the umask, and the `chmod` then explicitly makes it world-readable. Failure scenario: an operator debugging one customer runs `collect --export bundle.zip --raw-ids --out /var/tmp/pings.csv` (a realistic use of `--raw-ids`, since debugging usually means you already know the tenant) and per-customer spend becomes readable by every local account. Fix: `0o600` when `--raw-ids` is set — or unconditionally, as the sibling does, since the pseudonymised default does not need 644 either. Adopting the sibling's atomic `os.open` idiom matters once the mode narrows, for the reason its own comment gives.

**Minor · the `agent` column has no `--raw-ids` gate.** `tenant` and `session` are pseudonymised by default; `agent` is exported verbatim. `dash/event.go:agentFromUserAgent` maps known clients to eight fixed tokens but otherwise falls back to **up to 32 characters of the raw lower-cased `User-Agent`**. Pre-existing Go behaviour, newly surfaced by this export. Low practical severity, but the docstring's "No message content is read" paragraph currently reads as a blanket privacy claim and this is the one column it does not fully cover — worth a sentence there.

**Nits:** dead `io` import (`:52`, zero references); no `unittest.skipUnless(HAS_PANDAS, …)` around the three pandas tests, so the body's "runs anywhere the collector does" claim is documented rather than self-enforcing; no test for a **5xx ping with all-zero usage** although the docstring calls that case out explicitly (I built it synthetically — the logic is correct, landing in the "no usage at all (see status)" bucket with `ping_cost_unpriced` correctly staying 0, it is just untested); no test pinning the "session with real rows and zero pings contributes no output rows" filter (verified empirically, unasserted).

### Verified safe

- **Zip-slip / absolute-path members**: crafted both a `../../../../tmp/...` member and an absolute-path member; neither escaped the `mkdtemp`'d root (stdlib `extractall` sanitises both). Note this protection is *incidental* to using `zipfile.extractall` — nothing in the script asserts it, so switching extraction libraries would silently reintroduce the class. No zip-bomb guard, which I'd accept for an operator-invoked forensics tool.
- **SQL injection**: the only string-built SQL interpolates names from the hardcoded `COLS` list; `--since-ms` is bound with `?` placeholders and `argparse type=int` rejects non-numeric input first.
- **Read-only on every path**: `collect()` opens `file:{db}?mode=ro` uniformly; `collect_export()` never touches sqlite. No `ATTACH`, no write, no temp file beside the DB; the export tempdir is mode 0700.
- **No content or credentials on any error path.** `request_content` is verifiably never selected — not in `COLS`, not in `SQL`, confirming the docstring's own claim.
- **The stdlib/pandas split is honoured in fact**: no `pandas` at module scope; it is imported lazily inside `cmd_stats`/`summarize` only, so `collect` really can run as the service account.
- **Edge cases probed synthetically**: n=0 and n=1 pings (gap fields come back `None`, not `0`, matching the documented contract; the self-check block is skipped entirely at n≤1 rather than taking a median of nothing), all-identical gaps, **median = 0** (`1.75 * 0` makes the near-2× check `g.gt(0)` — well defined, no `ZeroDivisionError`), a negative gap and a ~1.8×-median gap together, and percentages from a zero denominator (guarded by `pct()`). Notably, a genuinely free no-usage row is distinguished from an unpriced had-usage row — the "$0.00 read as free" trap this repo has hit before is handled.

### No docs gap

The diff adds no page and no `mkdocs.yml` entry, and that matches house convention: `analyze.py`, `analyze_content.py`, `cachecost.py`, `cgenv.py` and `diffsum.py` all ship without one. The scripts that do get pages are the ones backing a published investigation. So the docstring *is* the documentation here, which is why it was held to a doc's standard above.

### Not verifiable within the data boundary

Taken on the author's word, not disputed: the 1,038 pings / 255 sessions / 15 accounts / $150.87 over 2026-08-23..09-02; the 91.8% inside the ±12s band and the 270-272s clustering vs the uncorrected 273-288s scatter; the "0 flagged rows disagree, 0 unflagged rows shaped like a ping" audit result; and the "every ping followed within ~1s by a 400" export observation. Each is author-reported production evidence whose *mechanism* is independently verifiable in the Go source, which is what the audit table above does.

OsherElhadad:
Pushed `ab27bdf` to address the review at #196.

**Security fix (the one worth blocking on):** `cmd_collect()` was doing
`open(args.out, "w")` then `os.chmod(args.out, 0o644)` unconditionally,
including under `--raw-ids`, on a CSV carrying `tenant_id` plus per-row
dollar figures. Switched to the same idiom `kv_ttl_cost_model.py` already
uses: `os.open(args.out, os.O_WRONLY|os.O_CREAT|os.O_TRUNC, 0o600)`, so the
file is created 0600 and never exists world-readable, even transiently.

Verified with a real invocation over a synthetic DB (no production data
touched):
```
before (open+chmod):  644
after  (os.open 0600): 600
```

**Stale docstring citations**, checked against Go at this branch's
checkout: `proxy/keepalive.go:906` → `:908` (the 4xx-early-return guard),
`proxy/proxy.go:1109` → `:1143` (`keeper.arrive(`). Both fixed.

**Nits:** dropped the dead `io` import; added a sentence noting `agent` is
exported verbatim regardless of `--raw-ids`; added `unittest.skipUnless`
guards on the three pandas-only tests in `test_keepalive_ping_stats.py` so
a pandas-less run reports "skipped" instead of erroring; added a test for
a 5xx ping with all-zero usage and one for a session with no pings
contributing no output rows.

Self-test output — 16/16 pass either way, pandas-less and with pandas:
```
$ python3 deploy/harbor/test_keepalive_ping_stats.py -v
Ran 16 tests in 0.080s
OK (skipped=3)

$ uv run --with pandas python3 deploy/harbor/test_keepalive_ping_stats.py -v
Ran 16 tests in 0.168s
OK
```

**On the CI-coverage point:** I didn't add a CI job or a `--self-test`
flag — this script already ships `test_keepalive_ping_stats.py`, which
your mutation testing already showed catches real defects. Wiring that
suite into CI (a Python job in `ci.yaml`, or pinning it via a Go drift
test the way #212 does) seems worth doing, but I'd rather leave that as a
proposal than bolt on a CI job unreviewed — happy to open it as a
follow-up if that's wanted.

`git merge-tree origin/main HEAD` is still clean (exit 0, no conflict
markers) after this push.

Issue #201, opened 2026-09-03
expand: an expand permanently un-compacts its content and costs one prefix flip, undocumented and uncounted; in-place replacement was never built


amiddavid:
## What the code actually does today, since it is documented nowhere

Two mechanisms were apparently considered when expand was designed. **Only one is in the tree**, and its cross-turn behaviour is not what the other's rationale assumed.

**The in-turn path.** The model calls `context_guru_expand`; the proxy intercepts it (`proxy.go:~1650`), and `expand.Continuation` (`expand/response.go:69`) appends two messages **to the upstream request body** — the assistant message carrying the tool_use, then a user message with one `tool_result` block per call holding the original content — then re-invokes upstream, up to `maxExpandRounds` (3). The content lands at the **end**, so the cached prefix is untouched during the turn. That part works as intended.

**It does not persist as its own turn.** On the intercepted path the client never sees the tool_use/tool_result pair — the proxy answers it and returns only the final assistant response. So the next turn the client re-sends its own transcript without the expanded content. The expansion is transient.

**What persists is a flag, not the content.** `offload.MarkKeptVerbatim` writes `cg:keep:<content-hash>`, and thereafter every offloader skips that content (`skipReduce` → `isKeptVerbatim`). So from the next turn the **original message goes upstream in full, at its original position**.

## The consequence, measured

Turn N sent compacted bytes at that position; turn N+1 sends the full original there. That is a change inside the cached prefix, so it costs a **cache-write of the suffix** — at ~11.5x a cache read, which is the cost the tail gate exists to avoid everywhere else.

Measured directly (`mask`, `keep_recent: 0`, `min_tokens: 20`, one 4,200-token tool output):

```
turn 1 sent 73 tokens (compacted)
turn 2 sent 4,200 tokens (full)   gates=map[marker_or_kept_verbatim:1]
```

So "expand does not hurt the cache" is true **within** the turn and false **across** it. The flip happens once per expanded content, on the turn after the expand.

That is arguably correct behaviour — re-compacting would loop the agent straight back into another expand, which is exactly what `cg:keep:` exists to prevent, and one cache-write is cheaper than an unbounded expand loop. But it is a real cost that is currently **undocumented, uncounted, and invisible**: no counter distinguishes "prefix flipped because the agent expanded" from any other cache-write, so an operator cannot see it and a benchmark cannot attribute it.

## The alternative that was considered and does not exist

[...]

Issue #199, opened 2026-09-03
offload: the replay depth-gate bypass is justified per-message but keyed per-content, so a replay can splice at depth into uncompacted content


amiddavid:
## The mismatch

`extract_llm`'s same-session replay bypasses the cache-tail depth gate — the lookup is at `extract_llm.go:890` and the gate at `:932` — and the bypass is justified in the block's own comment:

> SAME-SESSION replay first. **This session already sent these compacted bytes on an earlier turn**, so the provider's cached prefix holds the COMPACTED form and replaying it is byte-identical — which is why it is safe at any depth, cached prefix included.

That justification is about **this message**. The lookup is keyed on **content** (`extract.ContentKey`), not on position. So the same content appearing at a *different* position also takes the bypass — and at a position that was never compacted, the provider holds the **original**, so splicing the compaction there changes bytes inside the cached prefix and forces a cache-write of the suffix. That is the cost the gate exists to prevent, at ~11.5x a cache read.

`extract_llm_sweep`'s replay path has the same shape.

## Why the window is narrow, and what widens it

Mostly self-limiting, which is why this is filed as a question rather than a defect:

* **The common case is not affected.** The client re-sends the original transcript every turn (see `collapse.go`, `failed_run.go`, `agentdiet.go` — "the agent re-sends the original each turn"), so a message compacted last turn arrives **unmarked again at the same position**, and the replay re-derives the same bytes. The provider holds the compacted form there. Nothing flips. This is the case the bypass is for.
* **A genuinely new occurrence arrives in the tail.** It is appended, so its index exceeds `MaxCachedIdx`, and `TailOnly` (`components/component.go:414-419`) already permits acting there — the bypass buys nothing. By the time that position is at depth it was compacted on its arrival turn.

So reaching "content at a position that is already at depth and was never compacted there" needs one of:

1. **the boundary moved** — an agent self-compaction resets the cached-prefix boundary (`compaction_resets` counts these);
2. **the component did not act on the arrival turn** — `llm_every_n_requests` throttle, `llm_max_per_request` cap, below `min_tokens`, a model decline, or (after #188) a refused reserve slot;
3. **a different component wrote the record** — see #197; the reuse is content-addressed and cross-turn, so `extract_llm` can encounter an id the sweep ruled on at a position `extract_llm` never touched.

## Evidence available, and what it does not show

[...]

PR #205, opened 2026-09-03
fix(proxy): tell an unreadable usage block apart from a provider that reported none, and record the shape once

### LINKED ISSUE #200: proxy: an unrecognised usage dialect is indistinguishable from no usage, and degraded silently for 4,015 requests

amiddavid:
Closes #200. Independent of #204 (#190) — both branch from `main`.

## The defect

`parseUsage` returning `ok=false` meant **five** different things and nothing distinguished them. Two are benign. The rest mean **token accounting is offline**, on a request that otherwise looks perfect: a healthy 200, correct savings counters, correct latency, correct everything else, and `fresh_input_tokens` / `cache_read_tokens` / `cache_write_tokens` quietly at 0.

It ran that way for **4,015 of 4,015 requests** in one benchmark iteration and was found two iterations later, in a post-mortem chasing a different question. The usage *was* in the body — LOCA priced those requests from it — so the proxy read a response that carried usage and found nothing in it.

## What is classified, and why it is not one counter

`usageMiss` names all five, and the reason rides on the `cg.request` log line as `usage_miss` so a benign miss is legible without being alertable:

| `usage_miss` | Meaning | Counted | Remedy |
|---|---|---|---|
| `absent` | no usage block anywhere we know to look | — | nothing |
| `all_zero` | recognised, every tier zero | — | nothing; legitimate |
| `no_body` | empty response | — | nothing |
| `unparsed_dialect` | a block **is** there, no recognised spelling read it | `usage_unparsed` | add the dialect |
| `unreadable_body` | the bytes were not a whole document, so the block was hidden | `usage_unreadable` | raise `sniffMax` / buffer |

The last two get counters, and **separate** ones, because their remedies are opposite: one means add the dialect the provider is speaking, the other means the bytes examined were not a whole document and the fix is upstream of any parser. One counter would have said "accounting is offline" without saying which half of the stack to look at — and putting either in with the benign cases is this issue's own defect, restated. Same argument as `expand/unresolved.go`'s malformed/missing split and #188's `stash_refused`/`stash_missing`.

## A fourth cause the issue did not list

`sniffer.bytes` returns `head + "\n" + tail` once a response passes `sniffMax` (64 KiB) each way, and that is **not valid JSON**. Reported as an unrecognised dialect it would send someone hunting for a field name that is not missing.

It is **narrower** than "the response was truncated", and the code and both test tables say so: gjson *scans* rather than walking a tree, so a spliced document whose top-level `usage` survived is read correctly and never reaches this classification. It fires only when the splice actually hid the block — the only case where anything is lost. Worth flagging because the obvious fixture (a malformed body) is read correctly and would have made the test vacuous. It did, on the first run; both tables now carry the note and the shape that genuinely hides the block.

## The shape record, which is the part that unblocks the dialect fix

On the **first** unaccounted response per process, `cg.usage_unaccounted` logs the response's **key names**: the top-level keys, where a usage block was found, and the key names inside it. No values, and never the body.

That turns "usage_reported is false" into "the provider is sending camelCase", and it needs **no captured body, no instrumented capture hop and no extra request** — so it works from any deployment that hits the gap rather than only from the one rig with a capture hop in the path. A body dump on this workload writes kilobytes of transcript content to disk per response; this record cannot, by construction, which is why it can ship on by default at DEBUG.

Once per process because a run with this gap has it on *every* request — 4,015 of 4,015 above — so a per-response record is a log line per request.

## Deliberately not here: the dialects themselves

Adding the camelCase or nested spellings needs a real response body to get the field names right. Guessing between `cacheWriteInputTokens` and `cacheCreationInputTokens` produces a parser that looks correct and reads zero — this issue's failure reproduced in the code meant to fix it. The nested paths here are probed for **existence only** and never read, which is what keeps the classification dialect-agnostic.

## Verification

`gofmt -l .` clean, `go vet ./...` clean, `go test ./...` all packages pass, `go test -race` clean over `proxy` and `metrics` (Go 1.26.4, eval box).

Four new tests, all revert-verified against a mutation collapsing the classification back to one undifferentiated miss:

| Test | Reverted → |
|---|---|
| `TestEveryUsageOutcomeIsDistinguishable` — 12 shapes, one per outcome per transport, incl. Bedrock camelCase and nested `response.usage` | "classified absent, want unparsed_dialect" |
| `TestOnlyAnAccountingOutageMovesTheUsageCounters` — the rule itself: alertable cases move their own counter, all four benign shapes move neither | "usage_unparsed moved by 0, want 1" |
| `TestTheShapeRecordNamesTheDialectAndCarriesNoContent` — answers the dialect question, once per process, no content and no values | the record is absent entirely |
| `TestTheShapeRecordSaysWhenTheBytesWereNotAWholeDocument` — `valid_json=false`, so a spliced window is not misread as a field-name gap | no record |

The no-content assertion was verified against a **second** mutation, because "the record does not contain this string" passes trivially when there is no record, and would also pass on a record that logged the body under a different key. With `usageShapeAttrs` changed to log the body it fails on its own subject: *"the shape record leaked response CONTENT, which is the one thing it must never do"*.

Two contract tests updated as designed: `TestStatsShapeIsUnchanged` (both fields added to the reviewed contract) and `TestEverySnapshotFieldIsExportedOrExempt` — the latter listed in `notExportedWhy` rather than read off `s`, for the reason that map's own preamble gives: the `/stats` handler fills these *after* `renderMetrics` takes its snapshot, so a `promLine` off `s` would export a permanent 0 while passing the test, which is precisely the silent-zero failure this change exists to report.

## What this unblocks

One request through a proxy on this branch produces the key-shape record, which answers the camelCase question that #199 and the `parseUsage` dialect fix are both waiting on — and #201's cost claim, which is unmeasurable while the three cache-token fields read 0. Per the validation session, no captured body or usage shape exists anywhere today (the capture hop's flap log records request metadata only; LOCA's 11 MB log has none of the twelve candidate spellings), so this is the cheapest available instrument.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1360,6 +1360,10 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 	// can never delay or fail a request.
 	var usage Usage
 	var usageOK bool
+	// WHY, when the answer is no. `usage_reported=false` meant five different things — two
+	// benign, three a measurement outage — and nothing distinguished them, so a run that had
+	// stopped accounting tokens on every request read as healthy (#200). See usageMiss.
+	usageWhy := usageMissNoBody
 	// status/rounds/expanded exist for the lifecycle log line: the upstream status is
 	// the one fact about a request an operator asks for first, and until now it reached
 	// the dashboard row and nothing else.
@@ -1407,7 +1411,8 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 		lg.Info("cg.request", "status", status, "serve_ms", msSince(reqStart, time.Time{}),
 			"upstream_rounds", rounds, "expands", expanded, "sse", sse, "sse_buffered", sseBuffered,
 			"fresh_input", usage.FreshInput, "cache_read", usage.CacheRead,
-			"cache_write", usage.CacheWrite, "output", usage.Output, "usage_reported", usageOK)
+			"cache_write", usage.CacheWrite, "output", usage.Output, "usage_reported", usageOK,
+			"usage_miss", usageWhy.String())
 	}()
 	for round := 0; ; round++ {
 		rounds = round + 1
@@ -1479,10 +1484,11 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			sse = sse || isSSE
 			// Stream straight through, sniffing usage from a bounded head+tail window as
 			// the bytes go by (no buffering of the whole response, no added latency).
-			first, u, ok := h.stream(w, resp)
+			first, u, why, ok := h.stream(w, resp)
 			if !sseBuffered {
 				sseFirstByte = first
 			}
+			usageWhy = why
 			if ok {
 				usage, usageOK = u, true
 			} else {
@@ -1530,10 +1536,11 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			// A raw expand call on this path is repaired on the REQUEST side instead
 			// (expand.RepairToolResults).
 			sse = true
-			first, u, ok := h.stream(w, resp)
+			first, u, why, ok := h.stream(w, resp)
 			if !sseBuffered {
 				sseFirstByte = first
 			}
+			usageWhy = why
 			if ok {
 				usage, usageOK = u, true
 			} else {
@@ -1549,10 +1556,11 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			// bail below ends the client's turn with a complete stream rather than cutting
 			// it off mid-message.
 		}
-		if u, ok := responseUsage(resp.Header.Get("Content-Type"), respBody); ok {
-			usage, usageOK = u, true
+		if u, why, ok := responseUsageWhy(resp.Header.Get("Content-Type"), respBody); ok {
+			usage, usageOK, usageWhy = u, true, why
 		} else {
 			usage.StopReason = u.StopReason // same reasoning as the stream path above
+			usageWhy = why
 		}
 		if isSSE && withheld == nil {
 			// The whole round is already on the wire. Either it never called expand, or it
@@ -1772,7 +1780,7 @@ func setUpstreamAuth(dst http.Header, up upstream) {
 //
 // It also returns the instant the client got its first byte (zero if the body was
 // empty), which is the SSE TTFB accounting in serve.
-func (h *Handler) stream(w http.ResponseWriter, resp *http.Response) (firstByte time.Time, u Usage, ok bool) {
+func (h *Handler) stream(w http.ResponseWriter, resp *http.Response) (firstByte time.Time, u Usage, why usageMiss, ok bool) {
 	defer resp.Body.Close()
 	copyHeaders(w.Header(), resp.Header)
 	w.WriteHeader(resp.StatusCode)
@@ -1795,8 +1803,11 @@ func (h *Handler) stream(w http.ResponseWriter, resp *http.Response) (firstByte 
 			break
 		}
 	}
-	u, ok = responseUsage(resp.Header.Get("Content-Type"), sn.bytes())
-	return firstByte, u, ok
+	// The window, not the response: over sniffMax each way this is head+"\n"+tail, which is not a
+	// whole JSON document. responseUsageWhy reports that as `unreadable_body` rather than as an
+	// unrecognised dialect, because the remedy is here and not in any parser.
+	u, why, ok = responseUsageWhy(resp.Header.Get("Content-Type"), sn.bytes())
+	return firstByte, u, why, ok
 }
 
 // msSince returns milliseconds from start to at (falling back to now if the
@@ -1905,6 +1916,9 @@ func (h *Handler) stats(w http.ResponseWriter, r *http.Request) {
 	snap.FrozenHits, snap.FrozenMisses = offload.FrozenStats()
 	// Process-wide, like FrozenHits, and so filled outside the cast below: a removal declined
 	// for want of a stash slot is counted by the COMPONENT, whatever store instance refused it.
+	// Process-wide like the counters below: the parser is package-level, so whatever tenant's
+	// response carried the unreadable usage, this is the process's total.
+	snap.UsageUnparsed, snap.UsageUnreadable = UsageGaps()
 	snap.StashRefused = offload.StashRefusals()
 	// The dangling-marker counter, process-wide for the same reason: it is a COMPONENT that
 	// replayed a marker whose payload had gone, whichever store instance lost it. Reported
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1488,7 +1488,11 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			if !sseBuffered {
 				sseFirstByte = first
 			}
-			usageWhy = why
+			// noteUsageMiss, not a bare assignment: usageOK is sticky-true across expand rounds
+			// while this was last-write-wins, so a request whose round 1 parsed usage and whose
+			// round 2 continuation carried none logged `usage_reported=true usage_miss=absent`.
+			// A pair that can contradict itself undoes the reason for having the reason.
+			usageWhy = noteUsageMiss(usageWhy, why, usageOK || ok)
 			if ok {
 				usage, usageOK = u, true
 			} else {
@@ -1540,7 +1544,7 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			if !sseBuffered {
 				sseFirstByte = first
 			}
-			usageWhy = why
+			usageWhy = noteUsageMiss(usageWhy, why, usageOK || ok) // see the other stream path
 			if ok {
 				usage, usageOK = u, true
 			} else {
@@ -1557,10 +1561,10 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			// it off mid-message.
 		}
 		if u, why, ok := responseUsageWhy(resp.Header.Get("Content-Type"), respBody); ok {
-			usage, usageOK, usageWhy = u, true, why
+			usage, usageOK, usageWhy = u, true, noteUsageMiss(usageWhy, why, true)
 		} else {
 			usage.StopReason = u.StopReason // same reasoning as the stream path above
-			usageWhy = why
+			usageWhy = noteUsageMiss(usageWhy, why, usageOK)
 		}
 		if isSSE && withheld == nil {
 			// The whole round is already on the wire. Either it never called expand, or it
--- a/proxy/usage.go
+++ b/proxy/usage.go
@@ -263,6 +263,20 @@ func parseSSEUsageWhy(raw []byte) (Usage, usageMiss, string) {
 	// a dialect gap, so a streamed all-zero block — legitimate — read as an outage. Per-event
 	// parseUsageWhy costs nothing extra (the boolean form called it anyway) and each event's own
 	// reason is already exact.
+	// THE TWO PARSERS ARE NOT SYMMETRIC, and the asymmetry is here: this one can never return
+	// usageMissUnreadable. It only calls parseUsageWhy with a block usagePresent already accepted,
+	// and that function reaches its ValidBytes branch only when no block was found anywhere — so
+	// `worst` ranges over absent/all_zero/unparsed_dialect and nothing else. A `data:` line cut
+	// mid-JSON yields no present block, so the stream reads as "the provider streamed no usage".
+	//
+	// That leaves usage_unreadable reachable for a sniffed JSON response and structurally
+	// unreachable for a sniffed STREAM, which is where a head+tail window would hide a block —
+	// #200's own defect, in the corner this classification does not cover. Narrow in practice:
+	// Anthropic puts input_tokens in message_start at the front of the head window, so `found` is
+	// true and the answer is `parsed`. Closing it properly means passing the sniffer's own
+	// knowledge that it spliced (s.total > len(s.head)) into responseUsageWhy, which would also
+	// make the buffered case exact instead of inferred from ValidBytes. Filed rather than guessed
+	// at here.
 	worst := usageMissAbsent
 	worstPayload := "" // the event `worst` came from, so the record describes the right one
 	for _, line := range strings.Split(string(raw), "\n") {
--- a/proxy/usage.go
+++ b/proxy/usage.go
@@ -240,14 +240,22 @@ func parseUsageWhy(body []byte) (Usage, usageMiss) {
 // merged across events, taking the maximum of each — a later event repeating a
 // value must not double it.
 func parseSSEUsage(raw []byte) (Usage, bool) {
-	u, why := parseSSEUsageWhy(raw)
+	u, why, _ := parseSSEUsageWhy(raw)
 	return u, why == usageMissNone
 }
 
 // parseSSEUsageWhy is parseSSEUsage plus #200's classification. An event stream that carried a
 // usage block no event of it could be read is the streamed form of the dialect gap, and it must
 // not be reported as "the provider streamed no usage" — the remedies differ.
-func parseSSEUsageWhy(raw []byte) (Usage, usageMiss) {
+// The third return is the PAYLOAD OF THE EVENT THAT SET THE REASON, for the shape record.
+//
+// The record used to describe the stream's LAST data: event, which on the Anthropic-family
+// transport is `message_stop` — usage arrives in `message_start`. So a streamed unrecognised
+// dialect produced a record of the terminal event: no usage_at, no usage_keys, nothing about the
+// dialect, and it consumed one of the bounded record slots to say it. That is exactly the shape a
+// streamed Bedrock aws/claude-* response has, which is where the gap this record exists to
+// diagnose would sit. Returning it from here also collapses two walks over the transport into one.
+func parseSSEUsageWhy(raw []byte) (Usage, usageMiss, string) {
 	var out Usage
 	found := false
 	// The WORST reason any event produced, which is what makes a streamed miss as precise as a
@@ -256,6 +264,7 @@ func parseSSEUsageWhy(raw []byte) (Usage, usageMiss) {
 	// parseUsageWhy costs nothing extra (the boolean form called it anyway) and each event's own
 	// reason is already exact.
 	worst := usageMissAbsent
+	worstPayload := "" // the event `worst` came from, so the record describes the right one
 	for _, line := range strings.Split(string(raw), "\n") {
 		line = strings.TrimSpace(line)
 		if !strings.HasPrefix(line, "data:") {
@@ -276,6 +285,7 @@ func parseSSEUsageWhy(raw []byte) (Usage, usageMiss) {
 			one, why := parseUsageWhy([]byte(`{"usage":` + u.Raw + `}`))
 			if why > worst {
 				worst = why // usageMiss is ordered benign -> alertable; see its constants
+				worstPayload = payload
 			}
 			if why != usageMissNone {
 				continue
@@ -289,12 +299,12 @@ func parseSSEUsageWhy(raw []byte) (Usage, usageMiss) {
 		}
 	}
 	if found {
-		return out, usageMissNone
+		return out, usageMissNone, ""
 	}
 	// Every block that appeared was rejected, and `worst` says why — an unrecognised spelling is
 	// alertable, an all-zero block is not, and no block at all is `absent` because that is what
 	// `worst` starts as.
-	return out, worst
+	return out, worst, worstPayload
 }
 
 // responseUsage picks the right parser for a response's content type. The returned
@@ -318,8 +328,13 @@ func responseUsageWhy(contentType string, body []byte) (Usage, usageMiss, bool) 
 	sse := strings.Contains(contentType, "event-stream")
 	var u Usage
 	var why usageMiss
+	// doc is the JSON document the shape record should DESCRIBE, which for a stream is the one
+	// event that produced the reason rather than the whole transport or its last event.
+	doc := body
 	if sse {
-		u, why = parseSSEUsageWhy(body)
+		var judged string
+		u, why, judged = parseSSEUsageWhy(body)
+		doc = []byte(judged)
 	} else {
 		u, why = parseUsageWhy(body)
 	}
@@ -331,7 +346,9 @@ func responseUsageWhy(contentType string, body []byte) (Usage, usageMiss, bool) 
 		case usageMissUnreadable:
 			usageUnreadable.Add(1)
 		}
-		recordUsageShape(sse, body)
+		// len(body) so the record still reports the whole response's size — that is what says a
+		// window was spliced — while describing the document that actually failed to parse.
+		recordUsageShape(sse, len(body), doc)
 	}
 	return u, why, why == usageMissNone
 }
@@ -424,8 +441,8 @@ func noteUsageShape(key string) bool {
 // is false" into "the provider is sending camelCase", which is the difference between an open
 // question and a one-line fix — without needing a captured body, an instrumented capture hop, or
 // any extra request.
-func recordUsageShape(sse bool, body []byte) {
-	attrs := usageShapeAttrs(sse, body)
+func recordUsageShape(sse bool, total int, doc []byte) {
+	attrs := usageShapeAttrs(sse, total, doc)
 	if !noteUsageShape(usageShapeKey(attrs)) {
 		return
 	}
@@ -464,20 +481,24 @@ func ResetUsageShapeRecordForTest() {
 
 // usageShapeAttrs builds the shape record's key/value pairs. Separated from the logging so a test
 // can assert what it does and — more to the point — what it does NOT contain.
-func usageShapeAttrs(sse bool, body []byte) []any {
-	doc := body
-	if sse {
-		// One event, so the record describes a JSON object rather than a transport. The last
-		// data: line is where both dialects put their terminal usage.
-		doc = []byte(lastSSEPayload(body))
-	}
-	attrs := []any{"sse", sse, "bytes", len(body),
+//
+// doc is the JSON document that failed to parse: the whole body for a buffered response, and for a
+// stream the ONE EVENT the classifier judged (see parseSSEUsageWhy) rather than the transport or its
+// last event. total is the whole response's size, which is the figure that says a window was
+// spliced.
+func usageShapeAttrs(sse bool, total int, doc []byte) []any {
+	attrs := []any{"sse", sse, "bytes", total,
 		// Not parseable means the window was spliced rather than the dialect being strange —
 		// stated in the record so it cannot be misread as a field-name gap.
 		"valid_json", gjson.ValidBytes(doc),
 		"top_level_keys", objectKeys(gjson.ParseBytes(doc))}
 	for _, p := range append([]string{"usage"}, nestedUsagePaths[:]...) {
-		if u := gjson.GetBytes(doc, p); u.Exists() {
+		// usagePresent, not Exists, and for the same reason the classifier uses it: a null or empty
+		// block resolves but names no fields, so the record pointed at `usage` and printed
+		// usage_keys=[] while a real nested block held the answer — and keyed as `usage|`, which
+		// collides with a genuinely empty top-level block. The record must look where the
+		// CLASSIFIER looked.
+		if u := gjson.GetBytes(doc, p); usagePresent(u) {
 			return append(attrs, "usage_at", p, "usage_keys", objectKeys(u))
 		}
 	}
@@ -497,22 +518,6 @@ func objectKeys(v gjson.Result) []string {
 	})
 	sort.Strings(keys)
 	return keys
-}
-
-// lastSSEPayload returns the last non-empty `data:` payload in an event stream, which is where
-// both dialects put the terminal usage. Empty when there is none.
-func lastSSEPayload(body []byte) string {
-	lines := strings.Split(string(body), "\n")
-	for i := len(lines) - 1; i >= 0; i-- {
-		line := strings.TrimSpace(lines[i])
-		if !strings.HasPrefix(line, "data:") {
-			continue
-		}
-		if p := strings.TrimSpace(strings.TrimPrefix(line, "data:")); p != "" && p != "[DONE]" {
-			return p
-		}
-	}
-	return ""
 }
 
 // stopReasonPaths are where a terminal reason lives, per dialect and per transport:
--- a/proxy/usage.go
+++ b/proxy/usage.go
@@ -55,6 +55,9 @@ type Usage struct {
 // later, in a post-mortem chasing a different question.
 type usageMiss uint8
 
+// THE ORDER IS PART OF THE TYPE, ascending in severity: parseSSEUsageWhy compares two reasons
+// with `>` to keep the worst one any event of a stream produced. A new value goes at its severity
+// position, not at the end.
 const (
 	usageMissNone       usageMiss = iota // tiers were read; not a miss
 	usageMissNoBody                      // nothing to look at (empty response). BENIGN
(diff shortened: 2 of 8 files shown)


amiddavid:
Follow-up in 2da9bd5, from the coref validation session's correction — comment and docs only, no behaviour change.

`unreadable_body` is reachable **only** from the sniffed path (`Handler.stream`, taken when neither proxy-injected tool is advertised), because that is the only place usage is read from a bounded window rather than the whole body. With either tool advertised — which `inject_expand: always` guarantees from the first turn — `proxy.go`'s `default:` branch reads the body whole and the window cannot apply.

Worth recording because the call-site distinction has now misled two sessions in **opposite directions**. I floated the spliced window as a competing explanation for iteration 024's 4,015-of-4,015; it cannot be one, since those runs set `INJECT_EXPAND=always` so `advertised` was true (`proxy.go:1348`) and the full body was read at `:1552`. The #188 reviewer hit the mirror image — a `parseUsage` test that passed on the stream branch while the traffic took the buffered one. Same error both times: treating the two `responseUsage` call sites as interchangeable when the bytes reaching them are not.

Neither counter changes, and `valid_json` in the shape record already settles which case a reader has without knowing their path. This only stops them forming the wrong hypothesis first.

**Scoping consequence for #199 and the dialect fix:** for traffic that advertises either injected tool — the common case — camelCase-or-nested is the *only* remaining explanation for a miss, so the shape record's `usage_keys` is the whole answer in one request.

proxy/usage.go
```diff
@@ -129,25 +241,158 @@ func parseSSEUsage(raw []byte) (Usage, bool) {
 			out.Output = max64(out.Output, one.Output)
 		}
 	}
-	return out, found
+	switch {
+	case found:
+		return out, usageMissNone
+	case sawBlock:
```
amiddavid: **`sawBlock` makes three benign streamed shapes alertable, and one of them is the OpenAI dialect's normal output.**

The comment concedes "a stream whose only usage block is all-zero over-reports by one". It is worse than one, because `gjson.Result.Exists()` is `t.Type != Null || len(t.Raw) != 0` — a JSON `null` has `Raw == "null"`, so **`"usage": null` sets `sawBlock`**, `parseUsage` falls to its `default:` branch, and the stream is classified `unparsed_dialect`.

OpenAI-dialect streaming sends `"usage": null` on every chunk unless the caller sets `stream_options.include_usage`, and this proxy routes those through `Handler.stream` → `responseUsageWhy` (`proxy.go:1489`, the non-Anthropic `case isSSE:`). So on the `openai_upstream` route, `usage_unparsed` — documented as "add the dialect the provider is speaking" — is incremented by every healthy streamed response. Verified on this branch:

```
openai stream chunk with usage:null  -> why=unparsed_dialect ok=false unparsed+1 unreadable+0
sse recognised all-zero              -> why=unparsed_dialect ok=false unparsed+1 unreadable+0
sse usage empty object               -> why=unparsed_dialect ok=false unparsed+1 unreadable+0
```

That is this issue's own defect with the sign flipped: instead of an outage reading as healthy, healthy traffic reads as an outage — and it costs the same thing, an operator who cannot tell from the counter which they have.

The cheapest fix that keeps the coarseness you want: only set `sawBlock` when the block is a **non-empty object** (`u.IsObject() && len(u.Get("@keys").Array()) > 0`, or just `u.IsObject()` plus a `Raw != "{}"` check), and treat "every block seen was recognised and all-zero" as `usageMissZero`. `parseUsage` already distinguishes those two internally — `parseUsageWhy` per event would give you the exact reason without a second pass, which is the same asymmetry the split at `:118` was introduced for.

proxy/usage_gap_test.go
```diff
@@ -0,0 +1,180 @@
...
+	}{
+		{"an unrecognised dialect", "application/json", `{"usage":{"inputTokens":10}}`, 1, 0},
+		// The window has to be spliced in a way that actually HIDES the block: gjson scans rather
+		// than walking, so a truncated document whose top-level `usage` survives is still read
+		// correctly and is no problem at all. That is why this case is narrower than "the bytes
+		// were truncated" — see the same shape in TestEveryUsageOutcomeIsDistinguishable.
+		{"an unreadable window", "application/json",
+			`{"content":[{"text":"x"` + "\n" + `"usage":{"input_tokens":1}}`, 0, 1},
+		{"a provider that said nothing", "application/json", `{"stop_reason":"end_turn"}`, 0, 0},
+		{"a recognised all-zero block", "application/json", `{"usage":{"input_tokens":0}}`, 0, 0},
```
amiddavid, line 111: **This table is transport-blind, which is why the counter rule it names is violated on the stream path.**

Every row is `application/json`. `TestEveryUsageOutcomeIsDistinguishable` above deliberately covers "one per outcome per transport" and is the stronger table for it — but it does not assert on the counters, and this one does. So the rule *the counter must not be moved by a benign outcome* is asserted only where `parseUsageWhy` runs, and never where `parseSSEUsageWhy` does, which is the half that breaks it (see the note at `usage.go:247`).

Two rows would have caught it and are worth adding whichever way you fix the classifier:

```go
{"a streamed all-zero block", "text/event-stream",
    "data: {\"type\":\"message_delta\",\"usage\":{\"output_tokens\":0}}\n", 0, 0},
{"a streamed null usage field", "text/event-stream",
    "data: {\"choices\":[{\"delta\":{}}],\"usage\":null}\n", 0, 0},
```

And `TestEveryUsageOutcomeIsDistinguishable` has no `usageMissZero` row for SSE either, so that value is currently unreachable-by-test on the transport where it is misclassified. Both tables want the same discipline the first one already states: one row per outcome **per transport**, with the transport axis complete rather than sampled.

proxy/usage.go
```diff
@@ -129,25 +241,158 @@ func parseSSEUsage(raw []byte) (Usage, bool) {
...
+// while everything else about the request looks healthy.
+func UsageGaps() (unparsed, unreadable int64) {
+	return usageUnparsed.Load(), usageUnreadable.Load()
+}
+
+// usageShapeOnce keeps the shape record to the FIRST alertable response per process. It is a
+// diagnostic, not a metric: one record names the gap, and one per response would put a line in
+// the log for every request of a run that has the gap on all of them (4,015 of 4,015, in the
+// iteration that motivated this).
+var usageShapeOnce sync.Once
```
amiddavid: **One `Once` for two classes means the first alertable response wins, and it is unlikely to be the one that answers the question.**

The record is the part of this change that unblocks #199 and the dialect fix. It is gated on a single process-wide `sync.Once` shared by `unparsed_dialect` and `unreadable_body`, so whichever fires first consumes it. Verified on this branch — a spliced window, then the camelCase response:

```
spliced window first, camelCase next -> NO RECORD: the once was consumed by the
                                        unreadable_body, so the camelCase question
                                        is unanswered
```

Combine that with the `sawBlock` finding at `:247` and the odds get worse: on any deployment carrying OpenAI-dialect streamed traffic, a benign `"usage": null` chunk is alertable and will almost certainly be the *first* alertable response in the process, so the record is spent before the interesting response ever arrives. The run this ships for then produces the counters and no shape.

One `Once` per `usageMiss` value is the minimum. Better, and still bounded: key it on the record's own identity — `usage_at` plus the sorted `usage_keys` — in a small `sync.Map` capped at, say, 8 distinct shapes. That is one line per *distinct dialect* rather than one per process, which is what the diagnostic is actually for (a multi-provider deployment has more than one answer), and it still cannot write a line per request: a run with the gap on all 4,015 requests has one shape.

Also: `ResetUsageShapeRecordForTest` assigning `usageShapeOnce = sync.Once{}` is a plain data race against a concurrent `Do`. Harmless as used today (both callers are serial tests), but `-race` will find it the moment someone marks one of those tests `t.Parallel()`. Whatever replaces the `Once` should get a mutex-guarded reset.

proxy/proxy.go
```diff
@@ -1549,10 +1556,11 @@ func (h *Handler) serve(w http.ResponseWriter, r *http.Request, provider bschema
 			// bail below ends the client's turn with a complete stream rather than cutting
 			// it off mid-message.
 		}
-		if u, ok := responseUsage(resp.Header.Get("Content-Type"), respBody); ok {
-			usage, usageOK = u, true
+		if u, why, ok := responseUsageWhy(resp.Header.Get("Content-Type"), respBody); ok {
+			usage, usageOK, usageWhy = u, true, why
 		} else {
 			usage.StopReason = u.StopReason // same reasoning as the stream path above
+			usageWhy = why
```
amiddavid: **Across expand rounds, `usageWhy` is last-write-wins while `usageOK` is sticky-true, so the log line can contradict itself.**

`usageOK` is only ever set to `true` (`:1560`, `:1492`, `:1544`) and never reset; `usageWhy` is assigned unconditionally on every round. A request whose round 1 parsed usage and whose round 2 (post-`expand` continuation) carried none logs `usage_reported=true usage_miss=absent`. The reverse pairing is reachable too. Since the whole point of `usage_miss` is to let someone read `usage_reported=false` and know which of the five it was, a pair that can disagree undoes some of that.

Smallest fix: keep the reason of the round that left the request unaccounted — `usageWhy = why` in the `ok` branch, and `if !usageOK { usageWhy = why }` in the `else`.

Separately, worth a sentence in `UsageGaps`' doc: these count **responses**, not requests, so a multi-round request can move them more than once. The docstring already says "how many responses", so it is consistent — but `docs/reference/routes.md` invites comparison against a request count, and the 4,015-of-4,015 framing in the PR body is a per-request figure.

proxy/usage.go
```diff
@@ -57,9 +116,44 @@ type Usage struct {
...
+		// tool is advertised. With one advertised (inject_expand: always makes that true from the
+		// first turn) a non-streamed response is read whole and this cannot fire, which is worth
+		// knowing before treating it as a competing explanation for a miss: that mistake has been
+		// made twice, in both directions, from reading the call sites out of order.
+		//
+		// Narrower than "the response was truncated", and deliberately so: gjson SCANS rather
+		// than walking a tree, so a spliced document whose top-level `usage` survived is read
+		// correctly above and never reaches here. This fires only when the splice actually hid
+		// the block, which is the only case where anything is lost.
+		if !gjson.ValidBytes(body) {
```
amiddavid, line 209: Minor, and possibly deliberate: the `nestedUsagePaths` probe runs **before** `ValidBytes`, so a spliced window that hid the top-level `usage` but left, say, `response.usage` scannable is classified `unparsed_dialect` rather than `unreadable_body`. I think the current order is defensible — a block that *was* found really does mean a dialect is missing, whatever else is wrong with the bytes — but it is the one ordering that can produce the mis-signal the paragraph above says it exists to prevent, and `valid_json` in the record is what saves the reader. One clause saying the precedence is chosen (found-block wins over unparseable-document) would keep the next person from "fixing" it the other way.

amiddavid (commented):
The classification is the right shape and the `unreadable_body` cause you added is a real one — the narrowness argument (gjson scans, so a splice that leaves the top-level `usage` visible is read correctly) checks out, and the fixture note is worth keeping.

**But I think the change is currently net-negative on the streamed transport, and the run it is meant to unblock will probably produce no shape record at all.** Two findings, both verified by running probes against this branch on the eval box rather than by reading:

```
=== proxy, branch 2da9bd5, go test -run TestReviewProbe
openai stream chunk with usage:null  -> why=unparsed_dialect ok=false unparsed+1
sse recognised all-zero              -> why=unparsed_dialect ok=false unparsed+1
sse usage empty object               -> why=unparsed_dialect ok=false unparsed+1
spliced window first, camelCase next -> NO RECORD (the once was consumed)
```

1. Three benign streamed shapes move `usage_unparsed`, the counter whose whole meaning is "add the dialect". One of them — `"usage": null` — is what the OpenAI dialect puts in **every** streamed chunk unless `stream_options.include_usage` is set, so on the `openai_upstream` route this counter is driven by healthy traffic.
2. Because the shape record is one `sync.Once` for both alertable classes, the first of those benign streams (or any `unreadable_body`) consumes it, and the camelCase response that follows logs nothing. The datum #199 and the dialect fix are waiting on is the first thing this loses.

(1) also explains why review did not catch it: `TestOnlyAnAccountingOutageMovesTheUsageCounters` has no SSE row, so it asserts the rule on the one transport where the rule holds. Details inline. Everything else — the counter split, the no-content assertion and its second mutation, `valid_json` in the record, 2da9bd5's call-site scoping — I have no objection to.

proxy/usage.go
```diff
@@ -129,25 +288,231 @@ func parseSSEUsage(raw []byte) (Usage, bool) {
...
+		doc = []byte(lastSSEPayload(body))
+	}
+	attrs := []any{"sse", sse, "bytes", len(body),
+		// Not parseable means the window was spliced rather than the dialect being strange —
+		// stated in the record so it cannot be misread as a field-name gap.
+		"valid_json", gjson.ValidBytes(doc),
+		"top_level_keys", objectKeys(gjson.ParseBytes(doc))}
+	for _, p := range append([]string{"usage"}, nestedUsagePaths[:]...) {
+		if u := gjson.GetBytes(doc, p); u.Exists() {
+			return append(attrs, "usage_at", p, "usage_keys", objectKeys(u))
```
amiddavid, line 531: **`usageShapeAttrs` still uses `Exists()` where the classifier now uses `usagePresent`, so the record can name the wrong location.**

The loop here is `if u := gjson.GetBytes(doc, p); u.Exists()`. `parseUsageWhy` was changed to `usagePresent` for exactly this reason, and the two now disagree: a body with a top-level `"usage": null` *and* a real nested block is classified from the nested block, then described from the null one. Verified on this branch with `{"usage":null,"response":{"usage":{"inputTokens":10,"cacheReadInputTokens":9}}}`:

```
why=unparsed_dialect
cg.usage_unaccounted ... top_level_keys="[response usage]" usage_at=usage usage_keys=[]
```

`usage_at=usage usage_keys=[]` — it points at the null and reports no field names, while `response.usage` held the answer. It also keys the shape as `usage|`, so a second provider with a genuinely empty top-level `usage` collides with it.

One word: `usagePresent(u)` instead of `u.Exists()`. Worth doing even though the shape is speculative, because the whole reason `usagePresent` exists is that `Exists()` on a usage block does not mean what it looks like — leaving one caller on the old predicate is how that comes back.

amiddavid (commented):
Re-verified all five on the eval box against a41843c. Every one is fixed, and the null-usage finding is fixed more broadly than I reported it — you were right that `parseUsageWhy`'s top-level and nested probes had the same defect, so five rows rather than three:

```
json usage null            -> why=absent          unparsed+0
json usage empty obj       -> why=absent          unparsed+0
sse usage null per chunk   -> why=absent          unparsed+0
sse all-zero               -> why=all_zero        unparsed+0   (its own value, not the coarse one)
sse empty obj              -> why=absent          unparsed+0
sse camelCase              -> why=unparsed_dialect unparsed+1
```

`worst` with the severity-ordered constants is a better answer than the coarsening your original comment conceded, and stating that the order is part of the type is the right way to keep it. `noteUsageMiss`, `usageShapeKey` on `usage_at` + sorted `usage_keys` (and *not* the top-level keys — agreed, and for the reason you give), the mutex-guarded reset, and the RESPONSES-not-requests note all read correctly to me.

**Two things still wrong, both in the record rather than the classification** — so the counters are now right and the diagnostic they point at can still come out empty. Neither is large; the first is the one I would fix before merge, the second is a word.

On #204 I have nothing further — separate review there. And thank you for the bound on the four no-replay offloaders: a skipped component splices nothing, so nothing dangles while the skip lasts. That is the right narrowing and it is a better statement of the exposure than mine was.

proxy/usage.go
```diff
@@ -129,25 +288,231 @@ func parseSSEUsage(raw []byte) (Usage, bool) {
...
+}
+
+// usageShapeAttrs builds the shape record's key/value pairs. Separated from the logging so a test
+// can assert what it does and — more to the point — what it does NOT contain.
+func usageShapeAttrs(sse bool, body []byte) []any {
+	doc := body
+	if sse {
+		// One event, so the record describes a JSON object rather than a transport. The last
+		// data: line is where both dialects put their terminal usage.
+		doc = []byte(lastSSEPayload(body))
```
amiddavid: **The classifier scans every event; the recorder reads only the last one. An unrecognised dialect in a non-terminal event produces a record that names nothing.**

`parseSSEUsageWhy` walks all `data:` lines and keeps the worst reason any of them produced. `usageShapeAttrs` builds its record from `lastSSEPayload(body)` — a single event. When the block that drove the classification was not in the terminal event, the record describes the wrong object. Verified on this branch, a stream whose `message_start` carries camelCase and whose last event is `message_stop`:

```
why=unparsed_dialect
cg.usage_unaccounted sse=true bytes=255 valid_json=true top_level_keys=[type]
GAP: the record does not name the dialect that was actually found
```

No `usage_at`, no `usage_keys`, and `top_level_keys=[type]` from `message_stop`. Two consequences: the counter fires with nothing to act on, and via `usageShapeKey` this takes the `"|"` slot — the stable key you reserved for "no block found at all" — so the shape budget is spent on a record that says nothing.

This is the Anthropic-family transport shape, where the input-token block lives in `message_start`. Since the motivating provider is `aws/claude-*` through Bedrock, a streamed instance of exactly the gap this change exists to name can land here.

The fix that keeps the two in agreement is to record the event that produced the classification rather than the last one: have `parseSSEUsageWhy` also return the raw payload of the event that set `worst`, and pass that to `usageShapeAttrs` instead of calling `lastSSEPayload`. That also removes the current duplication, where the transport is parsed once to classify and again to describe. A cheaper version — scan events backwards for the first one containing a `usagePresent` block, falling back to the last — fixes the record but leaves the two walks able to disagree.

proxy/usage.go
```diff
@@ -129,25 +298,226 @@ func parseSSEUsage(raw []byte) (Usage, bool) {
...
 		}
 	}
-	return out, found
+	if found {
+		return out, usageMissNone, ""
+	}
+	// Every block that appeared was rejected, and `worst` says why — an unrecognised spelling is
+	// alertable, an all-zero block is not, and no block at all is `absent` because that is what
+	// `worst` starts as.
+	return out, worst, worstPayload
```
amiddavid, line 339: Non-blocking, and last thing from me: **`parseSSEUsageWhy` can never return `usageMissUnreadable`, so a spliced window that hides usage on the *streamed* transport is classified benign.**

A `data:` line cut mid-JSON does not yield a `usagePresent` block, so `worst` stays at its initial `usageMissAbsent` and the response is reported as "the provider streamed no usage" — no counter, no record. Verified on this branch:

```
spliced sse window -> why=absent  unparsed+0 unreadable+0  record: (none)
```

The reason this is worth a line: 2da9bd5 established that `unreadable_body` is reachable **only** from the sniffed path, and the sniffed path is exactly where SSE responses get a `head+"\n"+tail` window. So `usage_unreadable` fires for sniffed *JSON* bodies (which is what the fixture models, via `ValidBytes`) and is structurally unreachable for sniffed *streams* — where a splice that hides the block reads as `absent`, which is an accounting outage indistinguishable from a provider that reported none. That is #200's defect, in the corner this classification does not reach.

In practice the head window saves the common case: Anthropic puts `input_tokens` in `message_start`, which is at the front of the head, so `found` is true and the classification is `parsed`. So this is narrow, and I would not hold the PR for it.

If you do want it closed, the signal is already there and needs no new parsing: `sniffer` knows it spliced — `s.total > len(s.head)` in `bytes()`. Exposing that as a `spliced() bool` and letting `Handler.stream` pass it into `responseUsageWhy` would turn `absent` **on a spliced window** into `unreadable`, and it would make the buffered case exact rather than inferred from `ValidBytes` failing. Either way it belongs on the issue rather than in this diff, and a sentence at `worst`'s initialiser saying the streamed transport has no `unreadable` value would be enough to stop the next reader assuming symmetry between the two parsers.

amiddavid (commented):
Both fixed, verified from the record's own text on the eval box against 7a70d76:

```
sse dialect in a non-terminal event:
  cg.usage_unaccounted sse=true bytes=125 valid_json=true
    top_level_keys="[message type]" usage_at=message.usage
    usage_keys="[cacheReadInputTokens inputTokens]"

null top-level beside a real nested block:
  cg.usage_unaccounted sse=false bytes=71 valid_json=true
    top_level_keys="[response usage]" usage_at=response.usage
    usage_keys="[inputTokens outputTokens]"
```

And every shape from all three rounds re-checked for regressions — the five null/empty/all-zero cases still benign with no counter movement, camelCase still `unparsed_dialect`, both healthy shapes still `parsed`. Keeping `bytes` as the whole response while narrowing only the described document is the right split; deleting `lastSSEPayload` rather than leaving it unused is what makes the one-walk property hold. **No objection left on this branch — LGTM** (this account cannot file a formal approval on its own PR).

One last observation below. It is a residual gap rather than a defect in this change, it is non-blocking, and I would be equally happy with a comment as with code — I am raising it because it is #200's own shape on the one transport that is not covered, and because the signal it needs already exists.

I have nothing further on #204 either — separate note there.

OsherElhadad:
## Review: ran it, mutated it, and put a real spliced response through a live proxy

Reviewed at `0e12f02` against `main` (`87377e73`) in a dedicated worktree on the eval box, Go 1.26.4. **Verdict: approve** — one minor coupling gap worth a follow-up, no blockers. Details below are all measured, not read off the description.

### Real run results

| Command | Result |
|---|---|
| `gofmt -l .` | clean |
| `go vet ./...` | clean |
| `go build ./...` | 2.06s |
| `go test -count=1 ./...` | **all 30 packages ok**, 1m23.1s (proxy alone 52.1s) |
| `CGO_ENABLED=1 go test -race -count=1 ./proxy/... ./metrics/...` | both ok, proxy 220.7s, metrics 1.0s, **no races** |

**8 usage test functions actually RAN** (`grep -c '^=== RUN'`), no `SKIP`, no `[no tests to run]`:
`TestEveryUsageOutcomeIsDistinguishable` (**18 subtests**, all PASS), `TestOnlyAnAccountingOutageMovesTheUsageCounters` (12 subtests), `TestTheShapeRecordNamesTheDialectAndCarriesNoContent`, `TestTheShapeRecordDescribesTheUnreadableEventNotTheLastOne`, `TestTheShapeRecordSkipsANullBlockForTheOneThatHasTheAnswer`, `TestTheShapeRecordSaysWhenTheBytesWereNotAWholeDocument`, `TestAnEarlierShapeDoesNotSpendTheRecordForALaterDialect`, `TestTheMissReasonNeverContradictsUsageReported`.

Note the description's table lists 4 tests and "12 shapes" — the shipped reality is 8 functions and an 18-row table (the later commits added coverage). More than advertised, but a stale number if anyone quotes it.

### Revert-verification — every claimed test, mutated in the source

| Test | Mutation applied to source | Observed |
|---|---|---|
| `TestEveryUsageOutcomeIsDistinguishable` | dialect-recognised-but-unread branch returns `usageMissAbsent` instead of `usageMissUnparsed` | FAILED on its own subject: `classified absent, want unparsed_dialect` (2 subtests: `bedrock_converse_camelCase`, `sse_in_an_unknown_dialect`) |
| `TestOnlyAnAccountingOutageMovesTheUsageCounters` | (1) widen `alertable()` to include `usageMissZero` | **PASSED WITH THE DEFECT PRESENT** — see finding |
| ″ | (2) route `usageMissZero` into `usageUnreadable.Add(1)` in the switch | FAILED: `usage_unreadable moved by 1, want 0` |
| `TestTheShapeRecordNamesTheDialectAndCarriesNoContent` | `usageShapeAttrs` logs `"doc", string(doc)` | FAILED: `"the shape record leaked response CONTENT, which is the one thing it must never do"` |
| `TestTheShapeRecordSaysWhenTheBytesWereNotAWholeDocument` | hardcode `"valid_json", true` | FAILED: `"the record does not say the bytes would not parse, so it reads as a dialect problem when the fix is in the sniffer"` |

The security-critical assertion holds under the adversarial mutation. Good.

### Live proxy, real spliced response

Fake upstream on `:4056`, this branch's proxy on `:4057`, `house` preset:

- **No usage** → `usage_reported=false usage_miss=absent`, counters unmoved.
- **Bedrock camelCase** → `cg.usage_unaccounted sse=false bytes=153 valid_json=true top_level_keys="[content id type usage]" usage_at=usage usage_keys="[cacheReadInputTokens inputTokens outputTokens]"`, then `usage_miss=unparsed_dialect`; `/metrics` → `cg_usage_unparsed_total 1`. **No values in the record** — confirmed on the wire, not only in a test.
- **A genuinely spliced window** (usage block placed past the 64 KiB head so it lands in the discarded gap; real response 143,455 bytes) → `bytes=131073 valid_json=false`, `usage_miss=unreadable_body`, `cg_usage_unreadable_total 1`. `131073 = 64<<10 + 1 + 64<<10` confirms the `head+"\n"+tail` math exactly. **The client still received the full 143,455-byte body** — fail-open confirmed live, not just asserted.
- First splice attempt accidentally left the block in the surviving tail and was read correctly as `parsed` — an unplanned live confirmation of the description's "gjson scans" claim, and of why the obvious fixture would have been vacuous.

### Security

- **Log injection**: key names containing a literal newline and a `"` round-trip through `usage_keys` into the JSON handler as one well-formed line, both escaped. No forged records.
- **Hostile object**: a 100,000-key response produced a **567-byte** log line — `objectKeys` returns `false` from its `ForEach` at 64 keys, which stops gjson's iteration, so the cap bounds CPU as well as output.
- Distinct-shape budget (`usageShapeMax = 8`) is mutex-guarded rather than a racy `sync.Once` reset; clean under `-race`.
- Both new series carry an empty label — flat counters, no cardinality risk.
- DEBUG is **not** default anywhere (`internal/logging.go` → `info`), and the record passes through the same credential-scrubbing `ReplaceAttr` as every other line except under `CG_LOG_PLAIN=1`.
- Record carries no URL, model, tenant or session id.
- **Fail-open under abuse**: nil / zero-length / non-UTF8 / 5,000-deep-nested / 10 MB / duplicate-key bodies — no panic. `go test -fuzz` 60s, **~989,000 execs, 589 corpus entries, no crash**.

### Findings

**Minor · code quality · `proxy/usage.go:94` and `:356-362`** — `usageMiss.alertable()` and the `switch why` that increments the counters are two independent sources of truth for "which reasons are alertable". Widening `alertable()` alone — exactly the edit someone adding a 6th `usageMiss` value would make — **passes the whole suite**: `recordUsageShape` starts firing on the new case while no counter moves and nothing catches it. Suggest driving the shape-record gate and the counter off one switch/map keyed on `usageMiss`, or a test asserting `alertable()`'s truth set equals the switch's case set. This is the one finding I'd like to see closed, since it is the same "two labels, one condition, silent drift" shape the PR is about.

**Minor · design, disclosed** — `nestedUsagePaths` is a fixed 4-entry list, so a 5th real nesting (`result.usage`, say) reads as `absent` rather than `unparsed_dialect`. Documented and acceptable, but it is #200's own shape at smaller scale; worth watching via the shape record.

**Traced and confirmed accurate, not defects:**
- Precedence is nested-block-found > unreadable-bytes > absent — matches the description.
- `parseSSEUsageWhy`'s `worst` can only be `{absent, all_zero, unparsed_dialect}`; the `unreadable` branch is genuinely unreachable from there, so a streamed+spliced response is indistinguishable from streamed+absent. Commit `0e12f02` **discloses this rather than hiding it**, and the docs state a sufficient condition without overclaiming. Verified in code, not taken from the comment.
- `all_zero` cannot be told from "a parser read the wrong field and got 0" — pre-existing, not worsened here.
- Every field/metric name in `docs/reference/routes.md` matches the code; both new series trace snapshot field → `promexport` line → docs → test.

1. **The three PRs collide on one contract surface** — `metrics.Snapshot`, `notExportedWhy`, `proxy/promexport.go`, `proxy/stats_golden_test.go`, `docs/reference/routes.md`. I merged all three onto `main` by hand: conflicts in 4 files, all purely additive adjacency. After a union resolution the tree builds, `go vet` is clean, and **both contract tests pass in the union** (2 functions RAN, both PASS). No semantic incompatibility — but whoever merges second and third must re-run those two tests rather than trusting a green from the standalone branch.
2. **Every commit in all three PRs carries a duplicated `Signed-off-by:` trailer** (2 per commit, with `Assisted-By:` sandwiched between). DCO passes, so it is cosmetic, but it looks like `commit -s` on an amend that already had one.
3. The commits correctly use `Assisted-By:`, but these **PR bodies end with a "🤖 Generated with …" line**, which `CLAUDE.md` explicitly forbids in favour of exactly that trailer.
4. `proxy` carries a **pre-existing flake** unrelated to all three PRs: `TestFeedbackMailCannotInjectHeaders`. Mail is sent from a background goroutine that reads the process-global `CG_MAIL_DEV_SINK` **at send time**, while each test sets its own sink via `t.Setenv` + `t.TempDir` — so a late mail from a finished test lands in the next test's sink and the "exactly one real subject line" assertion counts 2. Seen once here under concurrent load; `main` passed 3/3 full `proxy` runs in isolation. Worth its own issue.

### Not verified

Real provider traffic — everything above is fixtures plus one fake upstream. The description is right that no captured body or usage shape exists anywhere yet, which is precisely what this record exists to fix; it can't be closed retroactively by a review. Also could not force the mail flake to reproduce, so I can neither rule it in nor out under load.

PR #208, opened 2026-09-04
fix(expand): stop sending recovered content upstream twice, and split the two skip reasons apart

### LINKED ISSUE #201: expand: an expand permanently un-compacts its content and costs one prefix flip, undocumented and uncounted; in-place replacement was never built

amiddavid:
Closes #201. Independent of #204 and #205 — all three branch from `main`.

Does the whole issue: document the cross-turn behaviour, count the flip, and fix the duplication. One measurement changes what the third one is *for*.

## The repair path sent the recovered content upstream twice, on every turn, forever

Not in the issue. Two mechanisms, each individually right, compose into it: `repairExpandErrors` rewrites the client's `No such tool available` tool_result with the original on every later turn (the client keeps its own copy of the error), **and** the same call marks that content kept-verbatim, so `skipReduce` also leaves the original message uncompacted at its own position. One copy in place, one in the tool_result.

Measured, both arms, `codesafe`, one 200-line tool output:

| | upstream bytes | copies |
|---|---|---|
| without the expand round-trip | **252** | 1 (marker + head peek) |
| with it, every later turn | **21,511** | **2** (400 probe hits) |

An **~85x** blowup against the compacted form, permanently — far larger than the one-off prefix flip the issue describes, which is what makes bullet 3 a fix for a live defect rather than a design preference. The control arm is load-bearing: without it, "the content appears once" would also pass on a pipeline that never compacted that message.

**The fix is a pointer, not a deletion.** The issue proposed deleting our tool_use/tool_result pair. Deleting messages shrinks the message count, and `modes.Boundary` reads `n < prev` as `compactionResets.Add(1)` with boundary 0 — so our own deletion would be attributed to *the agent* compacting its transcript. That is the counter-conflation defect again, inside the fix for it. A pointer keeps the message count unchanged: no boundary disturbed, no index shifting, no message left with an empty content array.

**The pointer is written only when the content really is there.** The agent's own compaction can drop the message the content came from while keeping the round-trip, and then the tool_result is the model's *only* copy. `contentPresent` checks by exact comparison against the values gjson decodes, not a substring scan of the raw body — JSON escaping makes a raw scan both false-negative and false-positive. It excludes the block being written, which the **existing idempotency test caught me on**: on a second pass the repaired block held the original, matched itself, and the repair replaced the content with a note pointing at what it had just removed.

## The split bullet 2 asks for already existed, at 3 of 11 sites

`kept_verbatim_after_expand` was raised by `extract_sweep`, `extract_llm` and `failed_run`; eight others raised the conflated `marker_or_kept_verbatim` for the same condition, which also covers the benign "already carries a marker". Gates reach `/stats` per component, so that was a **published** counter reporting two causes with opposite readings — and a published experimental figure was corrected twice off it before anyone noticed the label was the problem. Fourth instance of the #200 rule.

`skipReduce` now returns *which* reason, as two named constants, and all eleven sites raise it. `marker_or_kept_verbatim` is retired, with the three component docs, two offload tests and one dash fixture updated. `smartcrush` raised no gate at all here, so its share of both reasons was invisible even in the conflated form.

`dedup`, `extract` and `linecap` have **no `reapplyFrozen` path**, so `skipReduce` is the only place they can report an expansion — this gives those three a clean signal for the first time, and that is where most of the previously invisible share sits. (Same four the reviewer independently flagged on #204.)

## The flip is counted, and only when there is one

`reapplyFrozen` declined on kept-verbatim *before* looking up the frozen decision, so it could not tell an established compaction being abandoned — a suffix cache-write at ~11.5× a read — from content that was never compacted, where nothing flips. It now consults the decision inside that branch only: one extra `Get` on content the agent actually expanded, not on the hot path.

`expand_prefix_flips` at `/stats`, `cg_expand_prefix_flips_total` at `/metrics`. Documented as **per turn per message**, not per distinct content: every later turn re-observes the same abandonment while only the first is a real cache-write. It is a deliberate cost — re-compacting would bounce the agent into another expand — so counting it makes the trade visible instead of assumed.

## Verification

`gofmt -l .` clean, `go vet ./...` clean, `go test ./...` all packages pass, `go test -race` clean over `components/offload`, `expand`, `proxy`, `metrics` (Go 1.26.4, eval box).

Four new tests, each revert-verified. The two that could pass for the wrong reason were checked in **both** directions:

| Mutation | Failure |
|---|---|
| conflate the two gates again | `want gate already_marked, got map[kept_verbatim_after_expand:1]` |
| never count the flip | `counted 0 flips, want 1` |
| count **every** kept-verbatim skip as a flip | `counted 1 flips for content that was never compacted: the counter is measuring expansions, not cache-writes` |
| always write the content | `turn 1 sent the content 2.0 times (400 probe hits, 21511 bytes)` + `carries no pointer` |
| assume presence instead of checking it | `the content is nowhere in the request: the repair replaced the model's only copy with a pointer to something that is not there` |

Two contract tests updated as designed: `TestStatsShapeIsUnchanged` and `TestEverySnapshotFieldIsExportedOrExempt` — the latter in `notExportedWhy` because the series is sourced from offload's counter rather than off `s`, since a `promLine` off `s` would export a permanent 0 while passing the test (the silent-zero shape #205 is about).

## Not in this change

**The in-place replacement of a marker by its content.** The duplication it would have fixed is fixed above at a fraction of the risk. What remains of it — deleting the dead round-trip from the transcript — needs the boundary threading described above for roughly fifty bytes a turn. The gating signals it would use (`ColdCache` / `TailOnlyCold` / `compaction_resets`) all ship, so deferring it loses no ground.

### DIFF
--- a/components/offload/agentdiet.go
+++ b/components/offload/agentdiet.go
@@ -439,10 +439,15 @@ func (d *AgentDiet) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 			continue
 		}
 		content := schema.MessageText(msg)
-		if gate, skip := skipReduce(c, content); content == "" || skip {
-			if skip {
-				rep.Gate(gate)
-			}
+		if content == "" {
+			continue
+		}
+		// Short-circuit restored: skipReduce does a store Get, and the empty string is not
+		// content — asking about it costs a lookup and would raise a spurious gate if "" were
+		// ever marked. The original `content == "" || skipReduce(...)` had this ordering for a
+		// reason and splitting the call lost it.
+		if gate, skip := skipReduce(c, content); skip {
+			rep.Gate(gate)
 			continue
 		}
 		// A frozen decision was replayed above; a second reduction of the same bytes
--- a/components/offload/agentdiet.go
+++ b/components/offload/agentdiet.go
@@ -439,7 +439,10 @@ func (d *AgentDiet) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 			continue
 		}
 		content := schema.MessageText(msg)
-		if content == "" || skipReduce(c, content) {
+		if gate, skip := skipReduce(c, content); content == "" || skip {
+			if skip {
+				rep.Gate(gate)
+			}
 			continue
 		}
 		// A frozen decision was replayed above; a second reduction of the same bytes
--- a/components/offload/state.go
+++ b/components/offload/state.go
@@ -203,7 +203,26 @@ func frozenLost(c *components.Ctx, key string) bool {
 func reapplyFrozen(c *components.Ctx, rep *components.Report, comp string, m *bschemas.ChatMessage) ([]string, int, bool) {
 	content := schema.MessageText(*m)
 	if isKeptVerbatim(c, contentKey(content)) {
-		return nil, 0, false // agent expanded this; replaying the collapse would loop
+		// The agent expanded this; replaying the collapse would loop it into another expand.
+		//
+		// COUNT THE FLIP, which needs one more lookup than declining does. Whether this turn
+		// costs anything depends on something this branch did not previously ask: was there a
+		// frozen decision? If there was, the provider is holding COMPACTED bytes for this
+		// message and we are about to send the original in full at the same position — a change
+		// inside the cached prefix, so the whole suffix is re-written at ~11.5x a read. If there
+		// was not, the message was never compacted and nothing flips.
+		//
+		// Both cases raise the same gate at the caller (kept_verbatim_after_expand), because from
+		// the offloader's point of view they are the same decision. Only this one has a price, and
+		// nothing distinguished them before — so an expand-induced cache-write was indistinguishable
+		// from any other, which is #201's second bullet.
+		//
+		// One extra Get on rare content: this branch is reached only for content the agent actually
+		// expanded, not on the hot path.
+		if _, wasFrozen := c.Store.Get(frozenKey(c.Session, comp, contentKey(content))); wasFrozen {
+			expandFlips.Add(1)
+		}
+		return nil, 0, false
 	}
 	repl, ok := c.Store.Get(frozenKey(c.Session, comp, contentKey(content)))
 	if !ok {
@@ -286,6 +305,31 @@ var (
 // Exported for the host's /stats, which pairs them with the store's drop/repair counts.
 func FrozenStats() (hits, misses int64) { return frozenHits.Load(), frozenMisses.Load() }
 
+// expandFlips counts turns where an ESTABLISHED compaction was abandoned because the agent had
+// expanded that content: a frozen decision existed, so the provider is holding the compacted bytes,
+// and this turn sends the original in full at the same position.
+//
+// That costs a cache-write of the whole suffix at ~11.5x a read — the exact cost the cache-tail
+// gate exists to avoid everywhere else — and it was undocumented, uncounted and invisible: no
+// counter distinguished "the prefix flipped because the agent expanded" from any other cache-write,
+// so an operator could not see it and a benchmark could not attribute it (#201).
+//
+// It is a deliberate cost, not a bug. Re-compacting would bounce the agent straight into another
+// expand, and one cache-write is cheaper than an unbounded expand loop — which is why cg:keep:
+// exists. Counting it makes the trade visible instead of assumed.
+//
+// ONE PER TURN PER MESSAGE, not one per distinct content: every later turn re-sends the same
+// original and re-observes the same abandonment, so this grows with turn count. Only the FIRST is
+// a real cache-write — after that the provider has cached the full form — so read it as "expansion
+// is costing prefix churn in this deployment", not as a count of cache-writes. Distinct content is
+// what kept_verbatim_after_expand's per-component gate approximates.
+var expandFlips atomic.Int64
+
+// ExpandPrefixFlips returns how many times a replay declined because the agent had expanded content
+// that WAS compacted before. Non-zero means expansion is re-writing cached prefixes; see
+// expandFlips for why it is per turn per message rather than per distinct content.
+func ExpandPrefixFlips() int64 { return expandFlips.Load() }
+
 // contentKey is a marker/whitespace-insensitive content hash (shared with extract's
 // result cache), so the same output re-sent across turns maps to one frozen decision.
 func contentKey(s string) string { return extract.ContentKey(s) }
@@ -326,9 +370,40 @@ func isKeptVerbatim(c *components.Ctx, ck string) bool {
 // orphan the earlier stash), or the agent expanded it and re-compacting would just
 // trigger another expand — a per-turn bounce loop. Every offloader consults this on
 // each candidate so the kept-verbatim / never-double-reduce guarantees hold uniformly.
-func skipReduce(c *components.Ctx, content string) bool {
-	return expand.HasPlaceholder(content) || isKeptVerbatim(c, contentKey(content))
+// It returns WHICH of the two, because the caller publishes it as a gate and the two reasons want
+// opposite readings (#201). "Already carries a marker" is benign — that content is already
+// compacted, and nothing about this turn changes. "The agent expanded it" is an established
+// compaction being abandoned: from this turn on the original goes upstream in full at its own
+// position, which is a change inside the provider's cached prefix.
+//
+// They shared the label `marker_or_kept_verbatim` at eight of eleven sites while three offloaders
+// (extract_sweep, extract_llm, failed_run) already raised `kept_verbatim_after_expand` for exactly
+// the same condition. Gates reach /stats per component, so that was a published counter reporting
+// both a benign outcome and a real cost — and a published experimental figure was corrected twice
+// off it before anyone noticed the label was the problem. Same rule as #200 and #188.
+func skipReduce(c *components.Ctx, content string) (gate string, skip bool) {
+	// Marker first: content that already carries one was compacted by some component, and asking
+	// the store about kept-verbatim would be a wasted Get on the common path.
+	if expand.HasPlaceholder(content) {
+		return GateAlreadyMarked, true
+	}
+	if isKeptVerbatim(c, contentKey(content)) {
+		return GateKeptVerbatim, true
+	}
+	return "", false
 }
+
+// The two gate labels skipReduce distinguishes, named constants because eleven offloaders raise
+// them and a typo in one would silently split a counter in two.
+const (
+	// GateAlreadyMarked: this content already carries an offload marker, so reducing again would
+	// double-compact and can orphan the earlier stash. Benign — nothing flips.
+	GateAlreadyMarked = "already_marked"
+	// GateKeptVerbatim: the agent expanded this content, so it must not be re-compacted (that
+	// would bounce it straight into another expand). The turn this first appears is the turn the
+	// message reverts to its full form inside the cached prefix — see ExpandPrefixFlips.
+	GateKeptVerbatim = "kept_verbatim_after_expand"
+)
 
 // --- Stash ownership (scoping GET /expand by session) ----------------------
 //
--- a/components/offload/state.go
+++ b/components/offload/state.go
@@ -331,6 +331,13 @@ func FrozenStats() (hits, misses int64) { return frozenHits.Load(), frozenMisses
 // a real cache-write — after that the provider has cached the full form — so read it as "expansion
 // is costing prefix churn in this deployment", not as a count of cache-writes. Distinct content is
 // what kept_verbatim_after_expand's per-component gate approximates.
+//
+// AND ONE EVENT, not every expand-induced cache-write. This is its only increment site: a replay
+// declined because the content was expanded. summarize has a sibling that is NOT counted here —
+// trimSpanForKeptVerbatim shortens the span to protect expanded content, which can invalidate a
+// checkpoint whose boundary reached past the new end, and re-summarizing emits different summary
+// text at a fixed prefix position. Same class of cost, correct trade, deliberately uncounted:
+// a second increment site changes what this number means and wants its own decision.
 var expandFlips atomic.Int64
 
 // ExpandPrefixFlips returns how many times a replay declined because the agent had expanded content
--- a/components/offload/state.go
+++ b/components/offload/state.go
@@ -217,9 +217,17 @@ func reapplyFrozen(c *components.Ctx, rep *components.Report, comp string, m *bs
 		// nothing distinguished them before — so an expand-induced cache-write was indistinguishable
 		// from any other, which is #201's second bullet.
 		//
-		// One extra Get on rare content: this branch is reached only for content the agent actually
-		// expanded, not on the hot path.
-		if _, wasFrozen := c.Store.Get(frozenKey(c.Session, comp, contentKey(content))); wasFrozen {
+		// store.Peek, NOT Get, and this is the load-bearing part of the probe rather than a
+		// micro-optimisation. Memory.Get SLIDES the TTL and reorders the LRU, and FrozenPrefix is
+		// one of the PINNED namespaces — so probing with Get renewed a pinned entry that this
+		// branch has just decided will never be replayed, and pinned entries count against the
+		// shared exempt budget that gates rewind-reserve admission. A diagnostic would have been
+		// applying back-pressure to the reserve. Measured: with a 100s TTL and ten turns 60s apart,
+		// the probe kept a frozen decision alive 600s in.
+		//
+		// One extra lookup on rare content: this branch is reached only for content the agent
+		// actually expanded, not on the hot path.
+		if store.Peek(c.Store, frozenKey(c.Session, comp, contentKey(content))) {
 			expandFlips.Add(1)
 		}
 		return nil, 0, false
--- a/components/offload/summarize.go
+++ b/components/offload/summarize.go
@@ -152,19 +152,30 @@ func (s *Summarize) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 	// Keep msg0 (system/first) + the last keepLast; summarize the span between — with both
 	// boundaries aligned so neither cuts inside a tool exchange. See summarizeSpan.
 	headCount, start, end := summarizeSpan(msgs, s.keepLast)
-	// Never summarize away content the agent EXPANDED. summarize replaces the span wholesale
-	// rather than editing in place, so it is the one offloader skipReduce cannot protect — see
-	// trimSpanForKeptVerbatim for why that is both a bounce loop and a broken pointer.
-	if trimmed := trimSpanForKeptVerbatim(msgs, start, end,
-		func(text string) bool { return isKeptVerbatim(c, contentKey(text)) }); trimmed != end {
-		rep.Gate(GateKeptVerbatim)
-		end = trimmed
-	}
 	// Request-level trigger: don't summarize (an LLM call) until the transcript
 	// is genuinely large / deep. Zero thresholds fire always (back-compat).
 	if !s.trigger.Fires(req, c.CtxWindow) || end <= start {
 		rep.Skipped = true
 		return nil, nil
+	}
+	// Never summarize away content the agent EXPANDED. summarize replaces the span wholesale
+	// rather than editing in place, so it is the one offloader skipReduce cannot protect — see
+	// trimSpanForKeptVerbatim for why that is both a bounce loop and a broken pointer.
+	//
+	// BELOW the trigger gate, not above it: the trim costs a contentKey hash plus a store Get per
+	// span message, and paying that on a turn summarize was never going to act on is waste — it
+	// also raised kept_verbatim_after_expand on those turns, reporting a decision nothing made.
+	// The re-test of end <= start below is what keeps the original ordering's guarantee.
+	if trimmed := trimSpanForKeptVerbatim(msgs, start, end,
+		func(text string) bool { return isKeptVerbatim(c, contentKey(text)) }); trimmed != end {
+		rep.Gate(GateKeptVerbatim)
+		end = trimmed
+		if end <= start {
+			// Nothing summarizable is left once expanded content is protected. Declining is the
+			// outcome summarize already has for an empty span, and the safe direction.
+			rep.Skipped = true
+			return nil, nil
+		}
 	}
 	model := s.modelClient // config-pinned client wins
 	if model == nil {
--- a/components/offload/summarize.go
+++ b/components/offload/summarize.go
@@ -152,6 +152,14 @@ func (s *Summarize) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 	// Keep msg0 (system/first) + the last keepLast; summarize the span between — with both
 	// boundaries aligned so neither cuts inside a tool exchange. See summarizeSpan.
 	headCount, start, end := summarizeSpan(msgs, s.keepLast)
+	// Never summarize away content the agent EXPANDED. summarize replaces the span wholesale
+	// rather than editing in place, so it is the one offloader skipReduce cannot protect — see
+	// trimSpanForKeptVerbatim for why that is both a bounce loop and a broken pointer.
+	if trimmed := trimSpanForKeptVerbatim(msgs, start, end,
+		func(text string) bool { return isKeptVerbatim(c, contentKey(text)) }); trimmed != end {
+		rep.Gate(GateKeptVerbatim)
+		end = trimmed
+	}
 	// Request-level trigger: don't summarize (an LLM call) until the transcript
 	// is genuinely large / deep. Zero thresholds fire always (back-compat).
 	if !s.trigger.Fires(req, c.CtxWindow) || end <= start {
--- a/components/offload/summarize_keptverbatim_test.go
+++ b/components/offload/summarize_keptverbatim_test.go
@@ -1,0 +1,103 @@
+package offload
+
+import (
+	"strings"
+	"testing"
+
+	bschemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/rossoctl/context-guru/schema"
+	"github.com/rossoctl/context-guru/store"
+)
+
+// asstWithCall is an assistant turn that calls a tool, so a span boundary landing after it would
+// leave the call unanswered — one of the two provider rejections summarizeSpan exists to prevent.
+func asstWithCall(id string) bschemas.ChatMessage {
+	return bschemas.ChatMessage{
+		Role: bschemas.ChatMessageRoleAssistant,
+		ChatAssistantMessage: &bschemas.ChatAssistantMessage{
+			ToolCalls: []bschemas.ChatAssistantMessageToolCall{{ID: &id}},
+		},
+	}
+}
+
+// SUMMARIZE MUST NOT SUMMARIZE AWAY CONTENT THE AGENT EXPANDED.
+//
+// It is the one offloader that consults neither skipReduce nor isKeptVerbatim, and it replaces
+// msgs[start:end] wholesale rather than editing in place. Two consequences, one pre-existing and one
+// introduced by the repair now writing a pointer instead of a second copy:
+//
+//   - the bounce loop cg:keep: exists to prevent: the agent asks for content back, gets it, and next
+//     turn it is gone again;
+//   - expand.RestoredInPlace's pointer says the content is in the transcript, on the strength of the
+//     body BEFORE the pipeline runs. If summarize removes it, the model is left with a pointer to
+//     nothing where it previously got the content.
+func TestSummarizeLeavesExpandedContentInTheTranscript(t *testing.T) {
+	expanded := strings.Repeat("the content the agent asked to have back\n", 40)
+	filler := strings.Repeat("older chatter that is fine to summarize\n", 40)
+
+	st := store.NewMemory(store.Options{})
+	MarkKeptVerbatim(st, expanded)
+
+	// msg0, then a summarizable stretch, then the expanded output inside its own exchange, then a
+	// tail. keep_last is 1 so `end` starts well past the expanded message.
+	msgs := []bschemas.ChatMessage{
+		userMsg("system-ish opener"),
+		userMsg(filler),
+		asstWithCall("call_1"),
+		tool(expanded),
+		userMsg("what next?"),
+	}
+
+	_, start, end := summarizeSpan(msgs, 1)
+	trimmed := trimSpanForKeptVerbatim(msgs, start, end,
+		func(text string) bool { return text == expanded })
+
+	if trimmed >= 3 {
+		t.Fatalf("span still ends at %d, so the expanded tool output at index 3 is inside it and "+
+			"would be summarized away", trimmed)
+	}
+	// AND the boundary must not split the exchange: index 2 is the assistant turn that called the
+	// tool, so a span ending at 3 would leave its call unanswered and the request would be rejected
+	// (400 `tool_use` ids were found without `tool_result` blocks immediately after).
+	if trimmed > 2 {
+		t.Fatalf("span ends at %d, which leaves the tool_use at index 2 inside the span with its "+
+			"result outside: the provider rejects that request", trimmed)
+	}
+	for _, m := range msgs[trimmed:] {
+		if schema.MessageText(m) == expanded {
+			return // the expanded content survives outside the span, which is the point
+		}
+	}
+	t.Fatal("the expanded content is not in the kept part of the transcript")
+}
+
+// The safe direction when there is nothing left to summarize: decline. summarize already supports
+// end <= start, so this must not become a special case — and it must certainly not summarize the
+// expanded content because the remaining span is small.
+func TestSummarizeDeclinesRatherThanTouchingExpandedContent(t *testing.T) {
+	expanded := strings.Repeat("the content the agent asked to have back\n", 40)
+	msgs := []bschemas.ChatMessage{
+		userMsg("opener"),
+		tool(expanded), // immediately after the head, so nothing is summarizable without it
+		userMsg("what next?"),
+	}
+	_, start, end := summarizeSpan(msgs, 1)
+	trimmed := trimSpanForKeptVerbatim(msgs, start, end,
+		func(text string) bool { return text == expanded })
+	if trimmed > start {
+		t.Fatalf("span is [%d,%d) and still contains the expanded message at index 1", start, trimmed)
+	}
+}
+
+// A span with no expanded content must be left exactly as summarizeSpan computed it — the trim is a
+// protection, not a new boundary policy, and shrinking a healthy span would cost real savings.
+func TestTheTrimIsANoOpWithoutExpandedContent(t *testing.T) {
+	filler := strings.Repeat("older chatter that is fine to summarize\n", 40)
+	msgs := []bschemas.ChatMessage{
+		userMsg("opener"), userMsg(filler), userMsg(filler), userMsg("what next?"),
+	}
+	_, start, end := summarizeSpan(msgs, 1)
+	if got := trimSpanForKeptVerbatim(msgs, start, end, func(string) bool { return false }); got != end {
+		t.Fatalf("trimmed a span with nothing expanded in it: %d -> %d", end, got)
+	}
+}
--- a/components/offload/summarize_keptverbatim_test.go
+++ b/components/offload/summarize_keptverbatim_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	bschemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/rossoctl/context-guru/components"
 	"github.com/rossoctl/context-guru/schema"
 	"github.com/rossoctl/context-guru/store"
 )
@@ -31,12 +32,14 @@ func asstWithCall(id string) bschemas.ChatMessage {
 //   - expand.RestoredInPlace's pointer says the content is in the transcript, on the strength of the
 //     body BEFORE the pipeline runs. If summarize removes it, the model is left with a pointer to
 //     nothing where it previously got the content.
+//
+// THIS TEST AND THE TWO BELOW EXERCISE THE BOUNDARY ARITHMETIC IN ISOLATION, with a stub predicate
+// and no store — deliberately, because the retreat rule is worth pinning on its own. They do NOT
+// establish that Offload calls it: see TestSummarizeOffloadKeepsExpandedContent, which is the test
+// that fails if the call goes away.
 func TestSummarizeLeavesExpandedContentInTheTranscript(t *testing.T) {
 	expanded := strings.Repeat("the content the agent asked to have back\n", 40)
 	filler := strings.Repeat("older chatter that is fine to summarize\n", 40)
-
-	st := store.NewMemory(store.Options{})
-	MarkKeptVerbatim(st, expanded)
 
 	// msg0, then a summarizable stretch, then the expanded output inside its own exchange, then a
 	// tail. keep_last is 1 so `end` starts well past the expanded message.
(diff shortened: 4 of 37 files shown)


expand/repair.go
```diff
@@ -132,3 +161,49 @@ func repairOne(body []byte, restored []string, ours map[string]string, callID, p
...
+// message can hold text — a plain string content, an Anthropic text block, and an Anthropic
+// tool_result block's content.
+//
+// exceptPath IS LOAD-BEARING, and leaving it out was a bug caught by the existing idempotency test.
+// On a FIRST repair the tool_result holds the client's error, so there is nothing to self-match; on
+// a SECOND pass over an already-repaired body it holds the original, matched itself, and the repair
+// replaced the content with a note pointing at the content it had just removed. The repair must be
+// idempotent — the same body can pass through twice — so the block being written is excluded from
+// the search that decides what to write into it.
+func contentPresent(body []byte, orig, exceptPath string) bool {
```
amiddavid, line 180: **The presence check cannot see what the pipeline is about to remove, and one component removes messages without consulting the mark this relies on.**

The ordering is: repair (`proxy.go:1094`) → `applyMode` → pipeline. So `contentPresent` inspects the pre-pipeline body, and `RestoredInPlace`'s claim that the note cannot go stale — "MarkKeptVerbatim is re-applied … BEFORE the pipeline reads the transcript" — is only true for components that *check* the mark. Ten do:

```
$ git grep -n 'skipReduce(\|isKeptVerbatim(' components/offload/*.go | grep -v _test
agentdiet cmdfilter collapse dedup extract extract_llm extract_sweep
failed_run linecap mask readlifecycle skeleton   (+ state.go itself)
```

`summarize` is in neither list — `git grep -n 'isKeptVerbatim\|skipReduce\|keptKey' components/offload/summarize.go` is empty — and it does not edit messages in place, it replaces the whole span `msgs[start:end]` with one summary message. So under `preset: summarize` the sequence is: repair sees the original at its position → writes the pointer → summarize summarises that span away → the model receives `[expand: the content for id HASH is present in the transcript above]` and, above it, a summary plus a marker for a *different* id. Before this change it received the content.

The likely ordering is the bad one, not the rare one: the expand round-trip is the newest thing in the transcript at repair time, so it sits in summarize's kept tail, while the original is older and therefore inside the span.

There is a pre-existing half to this that the fix would also close. Summarize re-compacting content the agent just expanded *is* the per-turn bounce loop `cg:keep:` exists to prevent — the reason every other offloader consults it. So the cheapest fix is at that end rather than this one: have `summarizeSpan` (or the span loop) skip a message whose `contentKey` is kept-verbatim, which restores the invariant this note depends on and gives summarize the guarantee the other ten already give. Doing it in `contentPresent` instead — e.g. only pointing when the summarize component is absent — would leave the loop.

If you would rather not touch summarize on this PR, the honest alternative is to say so where `RestoredInPlace` states the invariant: the note is safe for pipelines whose components honour `cg:keep:`, and `preset: summarize` is not one of them. I would not leave the current sentence, because it states the property unconditionally and a reader will rely on it.

components/offload/state.go
```diff
@@ -203,7 +203,26 @@ func frozenLost(c *components.Ctx, key string) bool {
...
+		// was not, the message was never compacted and nothing flips.
+		//
+		// Both cases raise the same gate at the caller (kept_verbatim_after_expand), because from
+		// the offloader's point of view they are the same decision. Only this one has a price, and
+		// nothing distinguished them before — so an expand-induced cache-write was indistinguishable
+		// from any other, which is #201's second bullet.
+		//
+		// One extra Get on rare content: this branch is reached only for content the agent actually
+		// expanded, not on the hot path.
+		if _, wasFrozen := c.Store.Get(frozenKey(c.Session, comp, contentKey(content))); wasFrozen {
```
amiddavid: **This `Get` is not free in the way the comment says: it slides a pinned entry's TTL, so it keeps the abandoned decision alive for the whole session.**

"One extra Get on rare content" accounts for the CPU. `Memory.Get` also does `setExpiry(e)` unless `noSlide`, and `store.FrozenPrefix` is in `DefaultPinPrefixes` — so reading the decision here renews the lifetime of a *pinned* entry that this branch has just decided will never be replayed again. Verified on this branch, 100s TTL, ten turns 60s apart:

```
turns run (this branch): frozen decision still alive 600s into a 100s TTL: true
no turns run (control):  frozen decision still alive 600s into a 100s TTL: false
```

Why it is worth a line rather than a shrug: pinned entries count against `Memory.exemptRoom()`, which is what gates *stash* admission, so a pin that no longer expires is pressure on the rewind reserve — `stash_refused`, declined removals, lost savings. That is the budget #188 introduced and #204 is currently reworking, reached from a new direction by a diagnostic.

I am not sure it is wrong. `cg:keep:` is evictable and not pinned, so if the mark lapses the replay resumes and the decision *is* wanted again — on that reading, keeping it alive is a small improvement. But that is a resource decision, and right now it is being made accidentally by a counter. Either state it ("and the read deliberately keeps the decision alive, because the keep-mark can lapse before it does"), or take a non-sliding probe — the store has no `Has`/`Peek` today, and one narrow addition would serve this and any future diagnostic read.

Worth noting a second effect if you do add one: counting once per (component, content) rather than per turn would also retire the "per turn per message, not per distinct content" caveat you have had to document in three places.

components/offload/agentdiet.go
```diff
@@ -439,7 +439,10 @@ func (d *AgentDiet) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 			continue
 		}
 		content := schema.MessageText(msg)
-		if content == "" || skipReduce(c, content) {
+		if gate, skip := skipReduce(c, content); content == "" || skip {
```
amiddavid: Small one, and the only place the eleven-site conversion changed behaviour rather than shape. The original was `if content == "" || skipReduce(c, content)`, where `||` short-circuits — `skipReduce` never ran for empty content. Hoisting the call into the init statement runs it unconditionally, so an empty message now takes a store `Get` on `keptKey(contentKey(""))`.

Harmless in outcome (the `continue` happens either way) but it inverts the ordering `skipReduce`'s own new comment establishes — marker first precisely so the store is not asked on the cheap path — and if the empty string were ever marked kept-verbatim it would raise a spurious `kept_verbatim_after_expand` on every empty message. Reordering restores both:

```go
if content == "" {
    continue
}
if gate, skip := skipReduce(c, content); skip {
    rep.Gate(gate)
    continue
}
```

metrics/metrics.go
```diff
@@ -791,7 +791,23 @@ type Snapshot struct {
...
+	//
+	// It is a DELIBERATE cost, not a defect: re-compacting would bounce the agent into another
+	// expand, and one cache-write is cheaper than an unbounded loop. This makes the trade visible
+	// rather than assumed.
+	//
+	// Per turn per message, not per distinct content — every later turn re-sends the same original
+	// and re-observes the same abandonment, and only the FIRST is a real cache-write. Read it as
+	// "expansion is churning cached prefixes here", not as a count of cache-writes. Filled by the
+	// host at serve time (the counter lives in components/offload).
+	ExpandPrefixFlips int64 `json:"expand_prefix_flips"`
```
amiddavid, line 841: The new block landed between `CompactionResets`' comment and its field, so that comment now reads as documentation for `ExpandPrefixFlips`: everything from "auto-compaction should be non-zero, and a run where it stays 0 while savings fall off a cliff mid-session…" is about `compaction_resets`, and it now sits above the wrong name. Moving the `ExpandPrefixFlips` block plus its field below `CompactionResets` fixes it and needs nothing else — in a file where the comments are the reference, this one will be read as authoritative and it is describing the neighbour.

docs/reference/routes.md
```diff
@@ -211,6 +211,10 @@ reversibility it destroyed, silently.
 | `stash_missing` | Marker replays that found **no payload** behind them: a dangling `<<cg:HASH>>` went upstream. This one *is* a broken promise — raise `ttl_seconds`. |
 | `stash_expired` | Payloads reclaimed by the TTL. With `stash_missing`, the remaining way an outstanding marker stops resolving — raise `ttl_seconds`. |
 
+| Field | Meaning |
+|---|---|
+| `expand_prefix_flips` | Turns where an **established** compaction was abandoned because the agent had expanded that content, so the original went upstream in full at its cached position — a suffix cache-write attributable to expansion, at ~11.5× a read. Deliberate: re-compacting would loop the agent into another expand, and one cache-write is cheaper than an unbounded loop. **Per turn per message**, not per distinct content — only the first is a real cache-write. See [what an expand costs](../how-to/recover-context.md#what-an-expand-costs-across-turns). |
```
amiddavid: This opens a second `| Field | Meaning |` table immediately after the stash table, which splits that table from the paragraph explaining it — the `stash_refused` "leading indicator" prose that follows now reads as commentary on `expand_prefix_flips`. Either add the row to the table above (it is the same `/stats` field list) or give it its own short subsection with a heading, since the flip is a different mechanism from the reserve counters and arguably deserves one.

Same file as #204's stash-row edits, so this will conflict; flagging so whoever merges second rebases rather than resolves by picking a side.

amiddavid (commented):
I went at the measurement first, as invited, and **it holds up.** The control arm is doing real work and the one asymmetry I found in it runs the conservative way: the control's tool message is last in a two-message transcript while the treatment's is index 1 of 4, and depth makes compaction *more* likely, not less — so a control that compacts at the tail is the harder case to pass. The exact-`200` assertion in the treatment arm is only safe because kept-verbatim leaves that message untouched, which means a stray head-peek on that path would fail it loudly rather than pass it silently; that is the right way round. Not deleting the round-trip because `modes.Boundary` reads `n < prev` as a `compactionResets` is the correct call, and `exceptPath` plus the story of how the idempotency test found it is the most useful comment in the diff.

**One finding I would fix before merge, and it is the same class as the one you avoided by not deleting messages.** `contentPresent` establishes "the content is in the transcript" against the body as it is *before* the pipeline runs (`proxy.go:1094`, ahead of `applyMode`), and the claim that the pointer cannot go stale rests on kept-verbatim being honoured. Ten components honour it. `summarize` honours neither `skipReduce` nor `isKeptVerbatim` and replaces `msgs[start:end]` wholesale, so it can delete the message the pointer names, in the same request that wrote the pointer. Detail inline at `contentPresent`.

One verified note on the flip counter's lookup, and three small ones. Everything else — the gate split across all eleven sites, the named constants, `smartcrush` finally raising anything, the flip counter's both-direction verification, and leaving the in-place replacement out with the residual stated rather than buried — I have no objection to.

Also: this branch and #204 both edit `docs/reference/routes.md`'s stash rows, so whichever merges second needs a rebase, and the `stash_expired` remedy this table still gives as "raise `ttl_seconds`" is what #204 corrects.

amiddavid (commented):
The summarize fix is right, and the part you worked out beyond my suggestion — that setting `end` to the kept-verbatim index leaves a `tool_result` at the head of the tail with its `tool_use` still in the span — is correct and is the half I had not thought through. I traced the retreat against `summarizeSpan`'s two rejection shapes and it holds, including the two-results case and the retreat-to-`start` case. `store.Peek` is the right shape: non-sliding, and declining to remove an expired entry so a probe cannot move `stash_expired` or the reserve's byte total is the detail I would have missed. Declining the per-(component, content) recount as a separate decision rather than a quiet semantics change is the right call.

**But the lesson you say paid for itself did not get applied to the fix it was about.** The three new summarize tests all drive `trimSpanForKeptVerbatim` directly with a stub predicate, so none of them touches `summarize.Offload`. I removed the call site and ran the suite:

```
$ # summarize.Offload: trimSpanForKeptVerbatim call replaced with `_ = trimSpanForKeptVerbatim`
$ go test ./components/offload/ -run 'TestSummarize|TestTheTrimIsANoOp'
ok      github.com/rossoctl/context-guru/components/offload  0.238s
$ go test ./...
(28 packages ok, 0 failures)
```

The regression you just fixed — the one you say you would have shipped — is reachable again by deleting one statement, with the entire suite green. That is the same defect as the `A == A` loop on #204, and you caught its sibling one file over in `TestTheFlipProbeDoesNotRenewTheDecisionItAsksAbout`, which is exactly the right test. Detail and the cheap fix inline.

Two smaller things: an ordering point that is free to fix, and a consequence of the trim that the new counter does not see. Nothing else outstanding — the `agentdiet` early continue, the `metrics.go` move and the `routes.md` subsection are all as reported, and taking this branch off the stash rows does make the #204 rebase trivial.

components/offload/summarize_keptverbatim_test.go
```diff
@@ -0,0 +1,103 @@
...
+	msgs := []bschemas.ChatMessage{
+		userMsg("system-ish opener"),
+		userMsg(filler),
+		asstWithCall("call_1"),
+		tool(expanded),
+		userMsg("what next?"),
+	}
+
+	_, start, end := summarizeSpan(msgs, 1)
+	trimmed := trimSpanForKeptVerbatim(msgs, start, end,
```
amiddavid, line 55: **Nothing asserts that `summarize.Offload` calls this.** All three new tests call `trimSpanForKeptVerbatim` directly and pass `func(text string) bool { return text == expanded }` — a stub, not `isKeptVerbatim`. Verified by mutation on this branch: replacing the call at `summarize.go:158` with `_ = trimSpanForKeptVerbatim` leaves these tests green *and* the whole suite green (28 packages ok, 0 failures).

So the regression is protected by the helper's own correctness and by nothing else. Three things could reintroduce it with every test passing: deleting the statement, passing a predicate that does not consult the store, or moving it below the point `end` is used.

The tell is visible in the fixture itself — line 38 builds `st := store.NewMemory(...)` and calls `MarkKeptVerbatim(st, expanded)`, and then `st` is never used, because the stub replaced it. Dead setup is usually the signature of an assertion that stopped reaching the thing it was written for.

The harness for the real test already exists: `commitgate_summarize_test.go` drives `Summarize.Offload` end to end with a fake model and a ctx helper (`:99`, `:159`, `:181`). One assertion there — a store carrying the mark, `Offload` called, and the expanded message still present in `req.Input` afterwards — fails on all three mutations above. Keep these three as the boundary-arithmetic tests they are (the two-results and retreat-to-`start` cases are worth having in isolation); the point is that the wiring needs one of its own, for the reason `TestTheFlipProbeDoesNotRenewTheDecisionItAsksAbout` states in its own header: the drift is between what the call site does and what the helper offers, so the assertion has to live where the call site is.

components/offload/summarize.go
```diff
@@ -152,6 +152,14 @@ func (s *Summarize) Offload(req *bschemas.BifrostChatRequest, rep *components.Re
 	// Keep msg0 (system/first) + the last keepLast; summarize the span between — with both
 	// boundaries aligned so neither cuts inside a tool exchange. See summarizeSpan.
 	headCount, start, end := summarizeSpan(msgs, s.keepLast)
+	// Never summarize away content the agent EXPANDED. summarize replaces the span wholesale
+	// rather than editing in place, so it is the one offloader skipReduce cannot protect — see
+	// trimSpanForKeptVerbatim for why that is both a bounce loop and a broken pointer.
+	if trimmed := trimSpanForKeptVerbatim(msgs, start, end,
```
amiddavid, line 169: Free to fix, and it only goes one way: this runs **before** the trigger gate at `:165`, so a turn where summarize declines still pays the scan. The scan is a `contentKey` hash plus a store `Get` for every message in the span, and it does not short-circuit in the common case — with nothing expanded it walks the whole span to return `end` unchanged.

Summarize previously did zero store lookups per span message; the other ten offloaders pay one per candidate, so the cost is in line with the rest of the pipeline, but paying it on turns where the component does nothing is not. Moving the trim below the trigger check and re-testing `end <= start` afterwards keeps every property (the trigger reads `req`, not `end`) and skips the work on declined turns:

```go
if !s.trigger.Fires(req, c.CtxWindow) || end <= start {
    rep.Skipped = true
    return nil, nil
}
if trimmed := trimSpanForKeptVerbatim(...); trimmed != end {
    rep.Gate(GateKeptVerbatim)
    end = trimmed
}
if end <= start {
    rep.Skipped = true
    return nil, nil
}
```

As a side benefit the gate then means what it says: right now `kept_verbatim_after_expand` can be raised on a turn where summarize was never going to act, so the per-component gate count includes turns with no decision behind them.

components/offload/state.go
```diff
@@ -286,6 +313,31 @@ var (
...
+// It is a deliberate cost, not a bug. Re-compacting would bounce the agent straight into another
+// expand, and one cache-write is cheaper than an unbounded expand loop — which is why cg:keep:
+// exists. Counting it makes the trade visible instead of assumed.
+//
+// ONE PER TURN PER MESSAGE, not one per distinct content: every later turn re-sends the same
+// original and re-observes the same abandonment, so this grows with turn count. Only the FIRST is
+// a real cache-write — after that the provider has cached the full form — so read it as "expansion
+// is costing prefix churn in this deployment", not as a count of cache-writes. Distinct content is
+// what kept_verbatim_after_expand's per-component gate approximates.
+var expandFlips atomic.Int64
```
amiddavid, line 341: A consequence of the summarize trim that `expand_prefix_flips` does not see, and I think it wants a sentence somewhere rather than code.

The trim lowers `end` below the expanded message, so a checkpoint whose `boundary` reached past it now fails the `boundary > end` guard at `summarize.go:421` and **cannot be reused**. Summarize therefore re-summarizes a shorter span from scratch, which produces different summary text at a fixed position in the prefix — a suffix cache-write. Before the trim, that checkpoint kept matching and the summary stayed byte-identical (at the cost of the content, which is the bug you fixed). So the fix trades content loss for one cache-write, which is the correct trade and the same trade `ExpandPrefixFlips` was added to make visible.

It is invisible, though: `expandFlips.Add(1)` has exactly one call site, in `reapplyFrozen`'s kept-verbatim branch, so the counter only sees the replay family. `metrics.Snapshot` and `routes.md` both describe it in general terms — "turns where an established compaction was abandoned because the agent had expanded that content" — which now has a second instance in a component the counter never reaches.

Either narrow the wording to the replay path, or count it here (a checkpoint refused *because* the trim moved `end` is exactly the event, and it is distinguishable from the ordinary `boundary > end` refusal by whether the trim fired this turn). I would take the wording change on this PR and leave the counting to a follow-up, since one more increment site changes what the number means and you have just declined a semantics change for that reason.

components/offload/summarize_keptverbatim_test.go
```diff
@@ -0,0 +1,170 @@
...
+// Offload — or passing a predicate that never consults the store, or moving the call below the point
+// end is used — leaves all three green and the whole suite green, and the regression is back. The
+// tell was in the fixture: the first test built a store and marked content, then never used it,
+// because the stub had replaced it. Dead setup is what an assertion that stopped reaching its
+// subject looks like.
+//
+// This is the same shape as the vacuous cross-check on #204, and its sibling in this package
+// (TestTheFlipProbeDoesNotRenewTheDecisionItAsksAbout) is the version that gets it right: the
+// assertion has to live where the behaviour is.
+func TestSummarizeOffloadKeepsExpandedContent(t *testing.T) {
```
amiddavid, line 120: Observation, no action needed. `expanded` here is a hand-copy of `bulkResult`'s format string (`summarize_shape_test.go:41`), so the test depends on `bulkResult("t2")` producing byte-identical text. That coupling is invisible from this file.

It is safe as written, and that is worth saying explicitly rather than leaving to luck: the precondition fails *loudly* if the coupling ever breaks — `"the fixture does not contain the content to begin with"` fires before anything else runs, so a `bulkResult` edit cannot silently turn this into a test of an absent string. That is the difference between fragile and brittle, and it is the right side of it.

If you ever want to remove the duplication, `expanded := schema.MessageText(bulkResult("t2"))` says the same thing with the coupling stated. Not worth a commit on its own.

components/offload/state.go
```diff
@@ -286,6 +313,38 @@ var (
...
+// expand, and one cache-write is cheaper than an unbounded expand loop — which is why cg:keep:
+// exists. Counting it makes the trade visible instead of assumed.
+//
+// ONE PER TURN PER MESSAGE, not one per distinct content: every later turn re-sends the same
+// original and re-observes the same abandonment, so this grows with turn count. Only the FIRST is
+// a real cache-write — after that the provider has cached the full form — so read it as "expansion
+// is costing prefix churn in this deployment", not as a count of cache-writes. Distinct content is
+// what kept_verbatim_after_expand's per-component gate approximates.
+//
+// AND ONE EVENT, not every expand-induced cache-write. This is its only increment site: a replay
```
amiddavid, line 335: This reads correctly, and the sentence that does the work is the last one — "do not read a zero here as expansion cost nothing" in `metrics.Snapshot`. A counter that names a general phenomenon while measuring one instance of it is only safe if the doc says which instance, and all three surfaces now do.

One thing worth confirming when #209 is picked up: `kept_verbatim_after_expand` on summarize's gate row is the population-at-risk signal you named, and it is now raised only below the trigger gate — so it counts turns where summarize *would* have acted and the trim stopped it, which is exactly the right denominator for that question and a tighter one than it would have been before this commit's reordering. Worth writing into #209 while the reason is fresh, since from the issue's side it will look like an unrelated ordering change.

amiddavid (commented):
**Verified independently, and both mutations fail.** I rebuilt them from scratch rather than taking the report, and checked they compile first — for the reason you gave, a mutation that does not build is not evidence:

```
--- BASELINE ---   PASS
--- MUTATION A --- (call site replaced with `if false {`, compiles)
  FAIL summarize_keptverbatim_test.go:166: summarize removed content the agent had expanded
--- MUTATION B --- (predicate replaced with `func(string) bool { return false }`, compiles)
  FAIL summarize_keptverbatim_test.go:166: summarize removed content the agent had expanded
```

And the head independently: `gofmt -l` clean, `go vet ./...` clean, `go test ./...` with zero non-ok packages.

The precondition is what makes the new test carry its weight — asserting the *unmarked* fixture does get summarized away, so "the content survived" cannot pass on a fixture summarize never touched. That is the same guard the control arm gives the duplication measurement, applied one layer down, and it is the part I would have been most likely to leave out. The reordering below the trigger gate keeps the `end <= start` guarantee, and the note on the three boundary tests saying what they do and do not establish is worth more than the tests it annotates.

**All three PRs are LGTM from me and I have nothing outstanding on any of them.** #204 and #205 were done in earlier rounds; this closes #208. Two small notes below, neither needing a reply — one is an observation about the new fixture, the other is just confirming a scoping decision reads correctly.

On the process note: recording that two mutation attempts did not compile, and that neither counted, is the right instinct and it is rarer than it should be. A mutation that fails to build produces the same red as a mutation the test caught, which makes "the test failed" the least reliable sentence in a verification report unless someone checked which kind of red it was.

OsherElhadad:
## Review: reproduced the 21,511-byte measurement, ran all five mutations plus one more, and attacked `contentPresent`

Reviewed at `13f559d9` against `main` (`87377e73`) in a dedicated worktree on the eval box, Go 1.26.4. **Verdict: approve** — one real minor gap in `contentPresent` (fails safe), two nits, no blockers. Everything below is measured.

### Real run results

| Command | Result |
|---|---|
| `gofmt -l .` / `go vet ./...` / `go build ./...` | clean |
| `CGO_ENABLED=1 go build -tags cg_skeleton ./...` | clean — the tree-sitter path compiles |
| `go test -count=1 ./...` | **all 29 packages ok**, 1m29s |
| `CGO_ENABLED=1 go test -tags cg_skeleton -count=1 ./components/offload/...` | ok, 9.57s — tag-gated tests **exercised**, not just built |

Function counts that actually RAN (`grep -c '^=== RUN'`), all under `-race`:

| Package | RAN | SKIP | FAIL | Wall |
|---|---|---|---|---|
| `components/offload` | **381** | 5 (env-gated, each printing its `CG_LIVE=1` hint first) | 0 | 118s |
| `expand` + `metrics` | **61** | 0 | 0 | — |
| `proxy` | **440** | 2 (env-gated) | 0 | 218s |
| `dash` | **538** | 6 (env-gated) | 0 | 525s |

`go list -f '{{.TestGoFiles}} {{.IgnoredGoFiles}}'` confirms `skeleton_measure_test.go` / `skeleton_safety_test.go` sit in `IgnoredGoFiles` without the tag and move into `TestGoFiles` with it — so the default-build green is *not* evidence for the skeleton path, and it was checked separately. All four new test files (`expand_gates_test.go`, `summarize_keptverbatim_test.go`, `expand_duplication_test.go`, `peek_test.go`) are in `TestGoFiles`, i.e. genuinely compiled.

### The headline measurement, reproduced independently

A throwaway test (deleted afterwards) logged bytes and probe hits directly, since the shipped test only prints on failure:

| Arm | Turn | Bytes | Probe hits |
|---|---|---|---|
| round-trip, **fixed** | 1 and 2 | 10,979 | 200 |
| control (no round-trip) | 1 and 2 | **252** | **1** |
| round-trip, **defect reintroduced** (mutation d) | 1 and 2 | **21,511** | **400** |

Exact match to the description's 252 → 21,511 / 400. Identical on turn 2, so "every later turn, forever" holds. And the control is **not** vacuous: its `cg.request` line shows `tokens_before=2201 tokens_after=34 saved=2167`, so that message really was a compaction target.

### Revert-verification — all five claimed mutations, plus one of ours

| Mutation | Site | Observed |
|---|---|---|
| (a) `skipReduce` returns one gate for both branches | `state.go:399-409` | `want gate "already_marked", got map[kept_verbatim_after_expand:1]` — exact |
| (b) never `expandFlips.Add(1)` | `state.go:230-232` | `counted 0 flips, want 1` — exact |
| (c) always `expandFlips.Add(1)` | same | `counted 1 flips for content that was never compacted: the counter is measuring expansions, not cache-writes` — exact |
| (d) always write `orig` | `repair.go:143-146` | `turn 1 sent the content 2.0 times (400 probe hits, 21511 bytes)` **and** `the repaired tool_result carries no pointer to the content` — exact, both |
| (e) assume presence | `repair.go:144` | `the content is nowhere in the request: the repair replaced the model's only copy with a pointer to something that is not there` — exact |
| **(f) ours** — flip probe uses `Store.Get` instead of `store.Peek` | `state.go:230` | `the frozen decision survived 600s of a 100s TTL: the flip probe is renewing a PINNED entry` — exact |

**No pass-with-defect.** (f) is worth noting: the `Peek`-not-`Get` choice is load-bearing and already guarded, which is the right call.

### Attacks on `contentPresent` (the load-bearing safety check)

Correct and/or fail-safe on: unicode/emoji, JSON-escaped newlines and quotes, all four roles, truncated copies, trailing whitespace, two tool_results with identical content (`exceptPath` identity by path string — no off-by-one), three-pass idempotency (body length identical from turn 2, pointer stable, no growth), and a malformed body (no panic, `restored` empty).

### Finding · minor · `expand/repair.go:180-209` — `contentPresent` cannot see array-shaped `tool_result.content`

It compares only `blk.Get("text")` and `blk.Get("content")`, and only when `Type == gjson.String`. Anthropic allows a `tool_result` block's `content` to be an **array** of sub-blocks. In that shape `blk.Get("content")` is a JSON array, nothing recurses into it, `contentPresent` reports absent, and `repairOne` falls back to `answer = orig` — **reintroducing exactly this PR's duplication, scoped to that wire shape**. It fails *safe* (re-duplicates; never loses content, never writes a pointer to nothing), which is why it is minor.

One correction to the reviewer's own write-up, and the reason I'd not leave this for later: it is **not** an unfixtured shape. `apply/apply_test.go:195-212` already asserts the pipeline compacts text *inside* that shape, at path `messages.1.content.0.content.0.text`:

```go
map[string]any{"type": "tool_result", "tool_use_id": "t1", "content": []any{
    map[string]any{"type": "text", "text": strings.Repeat("structured line\n", 40)},
    map[string]any{"type": "image", "source": map[string]any{"data": "AAAA"}},
}},
```

So the marker-creating half of this exact shape is fixtured and passing today; only the repair half cannot see it. That makes the gap live rather than theoretical, on multi-part tool results (text + image) and on any client that uniformly wraps text in a block array.

Also worth aligning with the code: the doc comment claims comparison "at the three places a message can hold text — a plain string content, an Anthropic text block, and an Anthropic tool_result block's content". True only for the string spelling of the third.

Suggested fix: when `blk.Get(f)` is an array, recurse one level and compare each sub-block's `text`/`content` string; or flatten both sides through a `schema.MessageText`-style helper instead of hard-coding two field names. Either way please extend the fixture rather than only the code — that shape is the one this check will meet in production.

Separately and in the same direction: content living **only** in Anthropic's top-level `system` field is also not found. Same fail-safe outcome, much lower likelihood; mentioning it only so the two are fixed together if you touch this.

### Verified, not taken on faith

- **`modes.Boundary` / `compactionResets`**: `repairOne` only ever does in-place `sjson.SetBytes` value replacement (content, and conditionally `is_error`) — no message added or removed on **any** branch, including "content absent, no pointer written". Message count is invariant, so the claimed non-interference holds and the decision not to delete the round-trip is justified as described.
- **Fail-open**: store-loss (`found=false` short-circuits `contentPresent`, writes the `Unavailable` placeholder), malformed JSON, and `sjson` errors on either field all degrade to the unchanged body. No panics.
- **Security**: the pointer embeds only the `hashID` the model already sent — no session id, no path, no new store key. It is inert to our own parsing (`HasPlaceholder` matches `<<cg:...>>` and the summary sentinel, not this text), so a hostile tool output resembling it cannot steer the pipeline. `store.Peek` returns a boolean only — no widened read surface, and it does not resurrect or remove expired entries.
- **Retired counter**: three surviving references to `marker_or_kept_verbatim`, all historical prose (`state.go:394`, `expand_gates_test.go:17`, `routes.md:247`). No live code, fixture, JSON or dashboard reference; `dash/cachecost_test.go` is updated to `already_marked`. Independently re-grepped and confirmed.

### Nits

- `docs/how-to/recover-context.md`, "What an expand costs, across turns": *"It happens once per expanded content… and it is counted: `expand_prefix_flips`"* reads, in isolation, as if the counter increments once per content. The next paragraph corrects it and the code is unambiguous — but a skim-reader draws the wrong conclusion before the correction lands. Swapping the paragraphs fixes it.
- The flip counter undercounts when the frozen decision has itself expired before the expand (`Peek` absent → no flip counted, though the provider may still hold the compacted bytes). Consistent with how `frozenMisses` already conflates "never frozen" with "frozen then lost" in this same file, and both the snapshot comment and `routes.md` already hedge it. Not new; noting it as a known bound.

1. **All three PRs collide on one contract surface** — `metrics.Snapshot`, `notExportedWhy`, `proxy/promexport.go`, `proxy/stats_golden_test.go`, `docs/reference/routes.md`. I merged all three onto `main` by hand: conflicts in 4 files, all purely additive adjacency. After a union resolution the tree builds, `go vet` is clean, and **both contract tests pass in the union** (2 functions RAN, both PASS). No semantic incompatibility — but whoever merges second and third must re-run those two tests rather than trusting the standalone green.
2. **Every commit in all three PRs carries a duplicated `Signed-off-by:` trailer** (2 per commit, `Assisted-By:` between them). DCO passes; cosmetic, but it looks like `commit -s` on an amend that already had one.
3. The commits correctly use `Assisted-By:`, but these **PR bodies end with a "🤖 Generated with …" line**, which `CLAUDE.md` forbids in favour of exactly that trailer.
4. `proxy` has a **pre-existing flake** unrelated to all three PRs: `TestFeedbackMailCannotInjectHeaders`. Mail is sent from a background goroutine that reads the process-global `CG_MAIL_DEV_SINK` **at send time**, while each test sets its own sink via `t.Setenv` + `t.TempDir` — so a late mail from a finished test lands in the next test's sink and the "exactly one real subject line" assertion counts 2. Observed once here under concurrent load; `main` passes 3/3 full `proxy` runs in isolation. Worth its own issue.

### Not verified

Production-scale behaviour of the fix (no live SWE-bench/Terminal-Bench arm in this environment) — the byte numbers reproduce exactly at test scale, but the fleet-scale run that presumably originated them was not repeated. And one infrastructure oddity to record rather than resolve: the background `dash` race run was reported by the task runner as exit 1 while its own log is unambiguous (`ok … 524.819s`, 538 RAN, 0 FAIL, no `DATA RACE`, no `panic:`). Treating the log as authoritative, flagging the discrepancy.

expand/repair.go
```diff
@@ -132,3 +161,88 @@ func repairOne(body []byte, restored []string, ours map[string]string, callID, p
...
+// that shape, at messages.1.content.0.content.0.text. The marker-creating half was covered and the
+// repair half could not see it.
+//
+// One level is enough and the bound is deliberate: it is the depth the provider schema allows, and
+// unbounded recursion over attacker-influenced JSON is a cost this probe should not carry.
+func blockHasText(blk gjson.Result, blkBase, orig, exceptPath string) bool {
+	for _, f := range [...]string{"text", "content"} {
+		v := blk.Get(f)
+		switch {
+		case v.Type == gjson.String:
```
amiddavid, line 238: **The excepted field is now skipped only in its string spelling, so a repaired block can match itself through the array branch.**

The pre-refactor loop skipped `exceptPath` before looking at the value at all:

```go
for _, f := range [...]string{"text", "content"} {
    if blkBase+"."+f == exceptPath {
        continue
    }
    if v := blk.Get(f); v.Type == gjson.String && ... 
```

The new loop moved that test inside `case v.Type == gjson.String`. `exceptPath` names the parent (`messages.N.content.M.content`), while a sub-block is at `…content.M.content.0.text`, so carrying it down cannot match and the array branch has no exception. Verified on this branch — the excepted block's content is an array holding the original, and it is the only copy:

```
answer block: {"role":"user","content":[{"type":"tool_result","tool_use_id":"toolu_1",
               "content":"[expand: the content for id HASH is present in the transcript above]"}]}
pointer written: true | original still anywhere in the body: false
```

That is the round-1 bug verbatim — *"the repaired block held the original, matched itself, and the repair replaced the content with a note pointing at what it had just removed"* — for the spelling this commit added.

**On reachability, so this is not oversold.** The repair writes a *string*, and this function's own doc says the client re-sends its own error every turn, so I could not construct a client path that puts the original into the excepted block in array form. I reached it by constructing the second-pass state directly — which is what `exceptPath` exists for, and what `TestRepairIsIdempotentForAnArrayShapedToolResult` asserts. That test does not cover it because its array holds the **error**: pass 1 rewrites `content` to a string, so pass 2 takes the string branch and the exception applies. The contract is asserted for the array shape but not for the state that breaks it — the same fixture-avoids-the-trap shape as the stub predicate two rounds ago.

**Two lines**, restoring the original ordering:

```go
for _, f := range [...]string{"text", "content"} {
    if blkBase+"."+f == exceptPath {
        continue // the block being written, in EITHER spelling
    }
    v := blk.Get(f)
    switch {
    case v.Type == gjson.String:
        if v.String() == orig {
            return true
        }
    ...
```

Verified in both directions on this branch. With the fix, my probe writes the content instead of the pointer (`pointer written: false | original still anywhere: true`) and both controls keep their current behaviour — array-holding-the-error still gets the content, and the multi-part original elsewhere still gets the pointer, which is this commit's target case. `gofmt` clean; `./expand`, `./proxy` and `./components/...` all pass.

For the test: the array-shape idempotency fixture wants a second case where the array holds the **original** rather than the error, since that is the only state in which the exception is load-bearing for this spelling.

amiddavid (commented):
Checked 93ae3d6 rather than took the report, and the array fix does what it says — Control B below confirms the multi-part shape now resolves. **But the refactor dropped `exceptPath`'s exception for the array spelling**, so the self-match the round-1 idempotency test caught is reachable again for exactly the shape this commit added. Detail and a two-line fix inline, verified in both directions.

Worth saying plainly: **the second reviewer found two things I read and passed.** `alertable()` and the counter switch as two sources of truth is the better catch of the pair — I commented on that function's neighbours in round 1 and did not see that the predicate and the switch were independent lists. Their mutation (widen the predicate only, the edit a 6th value invites) is the right one and `usageCounters` plus `TestEveryUsageMissIsClassifiedAsBenignOrCounted` is the right fix: it makes construction carry the half construction can carry and a test carry the half it cannot. Same for the `apply_test.go` observation on this PR — that the marker-creating half of the array shape was already fixtured at `messages.1.content.0.content.0.text` while the repair half was blind is the kind of asymmetry I should have looked for, since I had the two halves in front of me.

d9fb04b on #204 is also a correction to a comment I read and let stand: I traced `remove`'s callers in round 1, saw `Get`'s lazy-expiry branch, and still did not notice the comment claimed `sweepExpired` was the only path. Being exact about which claim carries the conclusion is the right resolution.

PR #211, opened 2026-09-05
fix(sweep): stop crediting declined turns, see the prefix bust, and name the fallback failure

OsherElhadad:
## What this changes

Three defects found while measuring `extract_llm` and `extract_llm_sweep` over the 16-day
production corpus (220,933 requests, 892 model calls). Two are telemetry, one is economics.
Every figure below carries the query or the `file:line` behind it; full working in
`cg-research/results/extractllm-findings.md` and the page at
`cg-research/artifacts/extract-llm-deep-dive.html`.

Mechanism claims are read at the **deployed** commit `3a1b438`, not at `main` — prod runs a
side branch that is 2 commits ahead of and 22 behind `main`. All three defects are live on
`main` as well, which is why they are fixed here rather than merely reported.

### 1. `ExtractSweep.Offload` reported `mutated` for turns it declined

`Offload` has two return paths and only the tail one set `rep.Skipped`. The inventory floor
returns early:

```go
if len(cands) < e.minInventory {        // defaultMinInventory = 10
    rep.GateN("sweep_inventory_below_min", len(cands))
    return keys, nil                   // never reaches the changed == 0 guard
}
```

Outside the pre-expiry window phase 1 collects no candidates at all, so `len(cands) = 0 < 10`,
the early return fires, and `dash/event.go`'s `Mutated = !Reverted && !Skipped` reads `true`.
`GateN` is a no-op at `n <= 0`, so those turns recorded neither a gate nor a skip.

**29,321 of 29,372 production rows** read `mutated=1, acted=0, skipped=0`:

```sql
select acted, mutated, skipped, count(*) n,
  sum(events like '%sweep_offered%') has_offered,
  sum(gates like '%sweep_inventory_below_min%') has_below_min,
  sum(gates like '%not_in_pre_expiry_window%') has_notwindow
from request_components where component='extract_llm_sweep' group by 1,2,3;

-- acted mutated skipped     n  offered  below_min  not_window
--   0      0       1        8      8        0          0        <- reached the ask; guard worked
--   0      1       0    29321      7        7      29310        <- the early return
--   1      1       0       43     11        0         32
```

`sweep_offered` is emitted before the floor check, so it partitions the two exits cleanly: all
8 correctly-skipped rows have it, only 7 of the 29,321 do, and those same 7 are the only ones
with a non-zero `sweep_inventory_below_min`. 29,310 + 4 + 7 = 29,321 exactly.

**No dollar figure is affected** — `acted` derives from `saved_gross` and is correct. What is
wrong is the `mutated` census: the component is credited with 29,364 mutations, of which 29,321
are a component that declined. Fixed by guarding the outcome once in a `defer`, so both exits
and any future one are covered; the duplicated tail check is what went wrong here.

### 2. `sweep_prefix_cache_read_ZERO` could not see the harm it exists to catch

The counter tested `usage.CacheRead == 0`, so a **partial** hit silenced it however large the
write. The prefix ask's whole justification is that it *reads* the agent's cached prefix; when
the entry has gone it *writes* it instead, at 1.25× fresh on the agent's own model rather than
a tenth of it — which is precisely what `extract_econ.go` cited when it rejected prefix reuse.

Three production asks did exactly that, for **704,925 opus `cache_write` tokens = $3.42 = 73 %
of this component's entire spend**. Two of the three had read 39,805 tokens of a stable
sub-prefix, so `CacheRead == 0` was false and **$2.26 of the $3.42 went uncounted**:

```
req      idle before  ask read  ask write     cost  agent's own read / write
151760         1.3 s    39,805    260,604  $1.2685      48,040 / 572,018
175582       299.4 s         0    240,309  $1.1565      42,572 / 805,897   <- only this one fired
172127       276.1 s    39,805    204,012  $0.9920           0 / 687,058
```

Now counted separately as `sweep_prefix_cache_write`, because a zero read and a non-zero write
name different failures with different fixes. The existing counter keeps its exact meaning.

Worth recording alongside the fix, since it bears on whether to re-enable the component: on all
three turns **the agent's own request also re-created its prefix**, so the sweep walked into a
miss rather than causing one. But the trigger guarantees that — `sweeping()` fires only when
`0 < CacheTTLMs − IdleMs <= 60 s`, i.e. in the last minute of the entry's life, which is when a
prefix re-read is most likely to fail. No window width fixes that; the permission to act and the
need for a live prefix are mutually exclusive by construction.

### 3. `block_fallback` said the wrong thing about which failure it declined on

Widening the condition in change 2 also put the `blockFallback` branch on the partial-hit path, where
it still wrote *"the prefix ask read nothing from cache"* — false at `CacheRead = 39,805` with a
260,604-token write. Branched, so the two cases stay distinguishable.

One thing worth stating plainly, and a correction to how this section originally described the
string's reach:

- **Correction: the string is NOT unreachable — it is live in production and is now tested.** This
  section originally claimed `rec.Component` stays `""` on this path, so `Offload` discards the row
  on its `call.rec.Component != ""` guard, and it shipped only a counters test on that premise. That
  premise held for the unit test's own `rep := &components.Report{}`, but not for production:
  `components/pipeline.go`'s `runOne` sets `Report{Component: comp.Name(), ...}` before any component
  runs, and `extract_sweep.go` assigns `r.rec = components.ModelCall{Component: rep.Component, ...}`
  before the `block_fallback` branch is ever reached. So the row survives, reaches `rep.Calls`, and
  the branched `Rejection` string is exactly what an operator reads on `extraction_calls.rejection` —
  the column this same PR calls out as already mis-analysed twice. Added
  `TestBlockFallbackDecliningAWritePublishesTheCorrectRejection`, which sets `rep.Component` the way
  the pipeline does and asserts the exact published string.
- **A policy change worth stating plainly:** `block_fallback: true` previously declined only on a
  zero read and now also declines on any partial hit that wrote. A write on the agent's own prefix
  bills at 1.25× fresh on the *agent's* model rather than a tenth of it, so declining is the point of
  the switch — but this is a change to a config-gated brake, not just a counter split.

### 4. `cg.db` and `cg-control.db` sit untracked, and `*.db` is not ignored

The proxy writes its dashboard and control DBs with default names into whatever directory it is
started from, so a proxy run inside a checkout leaves production-shaped tenant data in the working
tree — untracked, but one `git add -A` away from a repo with three remotes. Both files were **0 bytes
with no tables**, so nothing leaked this time. Three lines of `.gitignore`.

### 5. The floor guard: the measurement, not the literal — and **no literal changes**

An earlier version of this branch raised `housellm`'s `min_tokens` from 3,000 to 5,000. **Dropped.**
The guard that replaces it carries the per-band cost measurement and the per-session replay
multiplier and derives the floor from them, so the next edit has to argue with the data. It asserts
the shipped 3,000 as a **lower bound**, and it currently passes.

Two corrections are baked in because the first version of the guard had both:

- **It named a constant `intervalTop` and gave it $14.71, which is the MEDIAN token value**, not the
  ceiling. The p90 is **$31.25**. With the median standing in for the top, the same rule selects 5,000
  instead of 2,000 — it **reversed the direction of a config change**. That number has now moved three
  times on the same 873 calls: an imported k=12 from another corpus, k=12 again through a mislabelled
  band, and the median through a constant's **name**.
- **It asserted equality, not a bound.** A higher floor only ever declines, and the bands are
  non-monotonic in cost (2–3k at $14.27 is cheaper than 3–5k at $15.76), which puts the ordering
  inside the noise of 25–359 calls per band. Equality would have pinned the literal to a rule fitting
  that noise.

The open-ended top band is skipped explicitly: treating its sentinel as a token count would derive a
floor of that size and **silently disable the component**.

### 6. The deterministic-only arm, as a runnable test

The question "would the free character window have done as well as the paid model leg?" had never
been measured, and it is the one arm that answers it without spending anything: with `model == nil`,
`strategyOrder` collapses to `["deterministic"]` and no call is issued. Skips without `CG_H0_DIR`, so
it costs CI nothing; the doc comment carries the export query and both internal-validity limits.

One gotcha recorded in the source because it cost real time: **the first version of this file was
named `..._arm_test.go`, which Go reads as the GOARCH filename build constraint.** It landed in
`IgnoredGoFiles` on amd64 and `go test` reported "no tests to run" and **exited 0** — a green pass for
a test that never executed. Detect that class with
`go list -f '{{.IgnoredGoFiles}}' ./...`, never with an exit code.

## Tests

Seven, and each guarding changed logic was confirmed to **fail without its fix**:

- `TestSweepDeclinedAtInventoryFloorReportsSkipped` — both exits, table-driven. Reverted, it
  reproduces the production row shape exactly:
  `gates: map[not_in_pre_expiry_window:1]` with `Skipped=false`, which is request 141575.
- `TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit` — asserts the write counter fires on a
  39,805-token read with a 260,604-token write, **and** that the zero-read counter does not, so
  the two are not collapsed.
- `TestBlockFallbackDistinguishesAZeroReadFromAWrite` — table-driven over both cases, asserting the
  right counter fires, the wrong one does not, and a declined ask leaves the transcript untouched.
- `TestBlockFallbackDecliningAWritePublishesTheCorrectRejection` — asserts `rep.Calls` itself, with
  `rep.Component` set the way `pipeline.go` sets it in production; see the correction to section 3
  above.
- `TestPanicClearsSkippedSetByTheComponentsOwnDefer` — a component's own deferred `Skipped`-guard
  (the pattern this PR's fix #1 introduced) can run during the same panic unwind that reaches
  `runOne`'s `recover`, which sets `Reverted`. Fixed `runOne` to clear `Skipped` there, so a
  `Report` never carries both — closing a half-state fix #1's own defer made newly reachable.
- `TestHousellmFloorClearsTheBandsThatLoseAtEveryK` — passes at the shipped 3,000, and verified to
  fail at 1,000 with the derived bound of 2,000 named in the message.
- `TestDeterministicOnlyArmOverCapturedBodies` — a measurement harness rather than a guard; skips
  without its corpus.

Also added `sweep_prefix_cache_write` to the counters list in
`docs/components/extract_llm_sweep.md`, which had gone stale for this PR's own change (no code
wiring needed — `metrics/metrics.go` and `proxy/promexport.go` already iterate gates generically).

`go build ./...`, `go vet`, and `go test ./...` are green.

## Deliberately not in this PR

Five things this investigation measured and is **not** shipping. Each is in the write-up with its
query.

- **The per-output floor.** An earlier draft of this branch raised the `housellm` literal from 3,000
  to 5,000 and I committed it. **Withdrawn and dropped from the branch.** The break-even band it
  rested on was mislabelled: solving `$4.755 + (k−1)×$0.3814` for the three printed values gives
  k = 3.80 / **12.00** / 27.11, so the middle figure was not this corpus's median but the 12.0
  imported from `docs/components/extract_llm.md` and measured elsewhere. The measured band is
  **$5.82 / $14.71 / $31.25**, and at it the change is worth **+$1.27 / +$0.13 / −$1.99** — $0.13 at
  the median, on a 7 % overshoot well inside the noise of a 70-session median. The rule the guard
  encoded now selects 2,000, which the shipped 3,000 already clears. The real exposure is six
  tenants overriding `min_tokens` to 500, which is a deployment fix.
- **Routing to the free deterministic window.** A paired experiment over the 468 candidate bodies
  stored in full shows the free window removes **2.56× more tokens for $0.00**, and class-gated
  routing would avoid **$4.51 of $5.02** and 3,619 s. Declined: the window keeps a median **33.1 %**
  of characters and drops **44.2 %** of identifiers against the paid program's **30.8 %**. And the
  paired corpus **excludes the paid leg's only winning size band entirely** — max candidate is 6,405
  tokens, so the 8–15k band at $4.05/MTok has zero representation — which biases the comparison
  toward the free arm. The next measurement is reward and steps, not tokens.
- **Widening the sweep's pre-expiry window.** Struck, not deferred. Grouping its 19 firing requests
  on whether it removed anything: **7 removal turns wrote 3,363,850 agent `cache_write` tokens
  against 4 same-idle-band controls' 4,907 — 686× at comparable growth**, all ten accepted rows at
  ≤0.196 cache warmth and all four no-removal warm rows at ≥0.997, and one request that grew by
  **exactly zero tokens** and wrote 592,279. `extract_llm` refuses depth, so the sweep is the only
  component editing inside the already-cached region, and a mid-history edit re-hashes everything
  after it. Revised net **−$19.6 to −$27.2**, not the ledger's −$3.60 (attributed, not proven; n=7
  vs 4). Its own premise fails too: it fired on a **live** cache on 13 of 19 requests and on a
  genuinely expired one **once**.
- **A pre-filter for the 364 calls that bought nothing.** Scored on both sides of the ledger, no
  free content predictor pays across the band. `looksLikeFileRead` — the predicate the codebase
  already has, whose comment calls line-numbered dumps irreducible — is **net negative at every
  point**, because those are the *most* profitable class the component sees (83 calls, 71.1 %
  acceptance, 169,171 tokens saved).
- **The two Starlark-generation fixes.** 96 of 356 dead calls ($1.3356, 20 minutes of model time)
  are programs the interpreter could not run, but they are ~six distinct bugs rather than one. The
  reply-budget theory is **refuted** — only 5 of 39 cut-off replies reach 4,000 output tokens
  against a 4,096 cap; the median is 859. The two small ones are the `while`-loop refusal (8 calls)
  and taking the first fenced block rather than only a leading one (5 calls; `stripFences` at
  `internal/extract/extract.go:397` strips only a **leading** fence). Together $0.22, and every
  prompt edit re-runs the acceptance corpus by design.

## Two defects reported earlier in this investigation that turned out not to exist

Recorded because both were mine and both flattered the conclusion:

- **"`saved_usd` is gross and nothing nets the component's own spend."** `dash/query.go:876`
  computes `NetUSD = SavedUSD − LLMCostUSD` **in the deployed binary**, and `dash/ui/app.js:2602-2611`
  already flags any component whose amortised and first-removal verdicts disagree in sign — which is
  exactly this component's situation, spelled out for the operator. The figure was on the dashboard
  the whole time.
- **"269 acted rows are silently unpriced."** All 269 are non-200 requests (401/408/429/502/503) with
  every token column *and* `cost_usd` at zero. The upstream never billed them, so `Event.Price`'s
  early return is correct. I priced a "$5.69 correction" by falling through to the fresh-input branch
  on rows with no token data; a reviewer independently produced the same error at $0.90. It is a
  property of the column — `saved_gross > 0` on an unbilled request — not of either of us.

## Scope note

`extract_llm_sweep` was removed from all nine tenant configs on 2026-09-03, so changes 1 and 2 have
no effect on current production traffic. They are shipped because the code is still reachable, and
because change 2 is the instrument anyone would need before deciding whether to re-enable it — a
decision the measurement above says should be **no**.

Net cash impact of this PR: **$0.** Its value is that a component reporting a decline as a mutation,
and a tripwire that could not see the harm it was built for, are how the sweep ran for three days
unexamined.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,3 +191,10 @@ site/
 # Personal/local operator notes — deployment-host facts, not repo documentation.
 # Never pushed.
 CLAUDE.local.md
+
+# SQLite databases. The proxy writes its dashboard and control DBs with default
+# names in whatever directory it is started from, so a proxy run inside a checkout
+# leaves production-shaped tenant data in the working tree, untracked but committable.
+*.db
+*.db-wal
+*.db-shm
--- a/components/offload/extract_sweep.go
+++ b/components/offload/extract_sweep.go
@@ -882,7 +882,16 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 		}
 		if e.blockFallback {
 			r.gate("sweep_fallback_blocked")
-			r.rec.Rejection = "the prefix ask read nothing from cache and block_fallback is set; " +
+			// SAY WHICH FAILURE THIS WAS. The condition above now admits a partial hit that WROTE,
+			// and on such a call "read nothing from cache" is simply false -- CacheRead can be 39,805.
+			// This string is persisted to extraction_calls.rejection, which is the column the whole
+			// rejection taxonomy is built from and which has already been mis-analysed twice, so a
+			// wrong reason here becomes a wrong finding later.
+			why := "read nothing from cache"
+			if usage.CacheWrite > 0 {
+				why = "re-created the agent's prefix instead of reading it"
+			}
+			r.rec.Rejection = "the prefix ask " + why + " and block_fallback is set; " +
 				"declining rather than paying again for a full-price transcript read"
 			return nil, r
 		}
--- a/components/offload/extract_sweep.go
+++ b/components/offload/extract_sweep.go
@@ -271,6 +271,18 @@ func (e *ExtractSweep) Offload(req *bschemas.BifrostChatRequest, rep *components
 	var cands []sweepCand
 	var keys []string
 	changed := 0
+	// Skipped is a property of the OUTCOME, not of one exit, and this function has two: the
+	// inventory floor below returns early, and every turn outside the pre-expiry window reaches
+	// it with no candidates at all. Setting it only on the tail return made 29,321 of 29,372
+	// production rows report `mutated=1` for a component that had DECLINED -- because
+	// dash/event.go computes Mutated as !Reverted && !Skipped, and GateN is a no-op at n<=0, so
+	// those turns recorded neither a gate nor a skip. A deferred guard covers both exits and any
+	// future one; the alternative, a copy of the check at each return, is what went wrong here.
+	defer func() {
+		if changed == 0 {
+			rep.Skipped = true
+		}
+	}()
 	// eligible counts candidates that cleared every gate this component knows about. Compared with the
 	// inventory's size below, to catch a pre-filter that thinned it -- see the comment at the append
 	// site for why that is the failure worth a tripwire.
@@ -553,9 +565,6 @@ func (e *ExtractSweep) Offload(req *bschemas.BifrostChatRequest, rep *components
 		}
 	}
 
-	if changed == 0 {
-		rep.Skipped = true
-	}
 	return keys, nil
 }
 
@@ -856,8 +865,21 @@ func (e *ExtractSweep) adjudicate(req *bschemas.BifrostChatRequest, c *component
 	//
 	// Note what neither mode can do: prevent the fresh read that already happened on THIS call. The
 	// counter is what tells an operator the window is mistimed.
-	if !fellBack && usage.CacheRead == 0 {
-		r.gate("sweep_prefix_cache_read_ZERO")
+	//
+	// A WRITE trips it as surely as a zero read, and testing the read alone let the harm through.
+	// Production: three asks re-created the agent's prefix for 704,925 opus cache_write tokens,
+	// $3.42 and 73% of this component's entire spend -- and two of them had read 39,805 tokens of
+	// a stable sub-prefix, so `CacheRead == 0` was false and the counter stayed silent on $2.26 of
+	// it. A partial hit is not evidence of safety: the write is the harm, at 1.25x fresh on the
+	// AGENT's model rather than a tenth of it, and it is exactly what extract_econ.go cited when
+	// it rejected prefix reuse.
+	if !fellBack && (usage.CacheRead == 0 || usage.CacheWrite > 0) {
+		if usage.CacheWrite > 0 {
+			r.gate("sweep_prefix_cache_write")
+		}
+		if usage.CacheRead == 0 {
+			r.gate("sweep_prefix_cache_read_ZERO")
+		}
 		if e.blockFallback {
 			r.gate("sweep_fallback_blocked")
 			r.rec.Rejection = "the prefix ask read nothing from cache and block_fallback is set; " +
--- a/components/offload/extract_sweep_test.go
+++ b/components/offload/extract_sweep_test.go
@@ -23,7 +23,11 @@ import (
 type fakeAsker struct {
 	reply     string
 	cacheRead int
-	err       error
+	// cacheWrite is what the provider says it WROTE for this ask. Non-zero means the agent's own
+	// prefix was re-created at the agent model's write rate, which is the harm the component's
+	// counters exist to catch -- see TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit.
+	cacheWrite int
+	err        error
 	// viaTool reports the reply as having arrived through the proxy's structured-answer tool rather
 	// than as text, which is the one thing about a reply the parser CANNOT infer: it reads a
 	// tool_use input and a JSON array in prose identically.
@@ -40,8 +44,8 @@ func (f *fakeAsker) Ask(_ context.Context, session, ask string) (string, compone
 	if f.err != nil {
 		return "", components.PrefixUsage{}, f.err
 	}
-	return f.reply, components.PrefixUsage{CacheRead: f.cacheRead, Fresh: 40, Output: 90,
-		ViaTool: f.viaTool}, nil
+	return f.reply, components.PrefixUsage{CacheRead: f.cacheRead, CacheWrite: f.cacheWrite,
+		Fresh: 40, Output: 90, ViaTool: f.viaTool}, nil
 }
 
 func (f *fakeAsker) ask() string  { s, _ := f.lastAsk.Load().(string); return s }
@@ -912,5 +916,77 @@ func TestSweepDoesNotAttributeAReplyShapeToTheFallback(t *testing.T) {
 	if rep.Events["sweep_answered_via_tool"] != 0 || rep.Events["sweep_answered_via_prose"] != 0 {
 		t.Errorf("the fallback was attributed a reply shape it could not have had (events: %v)",
 			rep.Events)
+	}
+}
+
+// A component that DECLINED must not report itself as having mutated the request, and the inventory
+// floor is a decline. It returns early, so the tail's `changed == 0` check never saw it: 29,321 of
+// 29,372 production rows read `mutated=1, acted=0, skipped=0`, which dash/event.go's
+// `Mutated = !Reverted && !Skipped` makes impossible for a component that actually skipped. Both
+// exits are asserted, because the two reach the floor with different candidate counts and only one
+// of them records a gate at all (GateN is a no-op at n<=0).
+func TestSweepDeclinedAtInventoryFloorReportsSkipped(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		idleMs   int64
+		wantGate string
+	}{
+		// In the window, but sweepReq offers 2 candidates against the default floor of 10.
+		{"in the window, below the inventory floor", 4 * 60 * 1000, "sweep_inventory_below_min"},
+		// Outside it, phase 1 collects nothing, so BOTH counters are suppressed at zero and the
+		// only record of the turn is the window gate.
+		{"outside the window, no candidates at all", 30 * 1000, "not_in_pre_expiry_window"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			asker := &labelAsker{verdict: "drop", needed: "none"}
+			asker.cacheRead = 19595
+			e := newSweep(t, "") // defaultMinInventory, NOT newSweepSmall's floor of 1
+			req := sweepReq()
+			c := preExpiryCtx("s", asker, store.NewMemory(store.Options{}))
+			c.IdleMs = tc.idleMs
+			rep := &components.Report{}
+			if _, err := e.Offload(req, rep, c); err != nil {
+				t.Fatalf("Offload must fail open: %v", err)
+			}
+			if n := atomic.LoadInt64(&asker.calls); n != 0 {
+				t.Fatalf("a declining turn made %d asks", n)
+			}
+			if !rep.Skipped {
+				t.Errorf("a component that declined reports Skipped=false, so dash reads mutated=1 "+
+					"for a turn that changed nothing (gates: %v)", rep.Gates)
+			}
+			if rep.Gates[tc.wantGate] == 0 {
+				t.Errorf("no %s gate recorded (gates: %v)", tc.wantGate, rep.Gates)
+			}
+		})
+	}
+}
+
+// The prefix ask's whole justification is that it READS the agent's cached prefix. When the entry has
+// gone it WRITES it instead, at 1.25x fresh on the agent's own model -- 704,925 opus tokens and $3.42
+// across three production calls, 73% of this component's entire spend. Testing `CacheRead == 0` alone
+// missed two of those three: they read 39,805 tokens of a stable sub-prefix and wrote a quarter of a
+// million, so the counter stayed silent on $2.26 of the harm. A partial hit is not evidence of safety.
+func TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit(t *testing.T) {
+	asker := &labelAsker{verdict: "keep", needed: "none"}
+	asker.cacheRead = 39805 // a real partial hit, so CacheRead == 0 is FALSE
+	asker.cacheWrite = 260604
+	e := newSweepSmall(t, "")
+	rep := &components.Report{}
+	if _, err := e.Offload(sweepReq(), rep,
+		preExpiryCtx("s", asker, store.NewMemory(store.Options{}))); err != nil {
+		t.Fatal(err)
+	}
+	if n := atomic.LoadInt64(&asker.calls); n != 1 {
+		t.Fatalf("expected ONE ask, got %d (gates: %v)", n, rep.Gates)
+	}
+	if rep.Gates["sweep_prefix_cache_write"] == 0 {
+		t.Errorf("an ask that re-created the agent's prefix for %d tokens recorded nothing "+
+			"(gates: %v)", asker.cacheWrite, rep.Gates)
+	}
+	// And the pre-existing zero-read counter must NOT fire, because the read did happen. The two
+	// name different failures and collapsing them would lose the distinction the fix is for.
+	if rep.Gates["sweep_prefix_cache_read_ZERO"] != 0 {
+		t.Errorf("a partial HIT was counted as a zero read (gates: %v)", rep.Gates)
 	}
 }
--- a/components/offload/extract_sweep_test.go
+++ b/components/offload/extract_sweep_test.go
@@ -990,3 +990,48 @@ func TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit(t *testing.T) {
 		t.Errorf("a partial HIT was counted as a zero read (gates: %v)", rep.Gates)
 	}
 }
+
+// block_fallback now declines on a partial hit that WROTE, not only on a zero read, so the two cases
+// must stay distinguishable. The counters are what a reader actually gets: on this path the call row
+// is DROPPED (foldFallback returns early because the fallback deliberately never ran, so
+// rec.Component stays "" and Offload's `call.rec.Component != ""` guard discards it), which means the
+// Rejection string built alongside is unreachable today. That is a pre-existing hole of the same class
+// the comment at foldFallback documents for the no-asker path -- recorded here rather than fixed,
+// because adding a row where none existed is a behaviour change this evidence does not cover.
+func TestBlockFallbackDistinguishesAZeroReadFromAWrite(t *testing.T) {
+	for _, tc := range []struct {
+		name        string
+		read, write int
+		wantGate    string
+		notGate     string
+	}{
+		{"zero read", 0, 0, "sweep_prefix_cache_read_ZERO", "sweep_prefix_cache_write"},
+		{"partial hit that wrote", 39805, 260604, "sweep_prefix_cache_write", "sweep_prefix_cache_read_ZERO"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			asker := &labelAsker{verdict: "drop", needed: "none"}
+			asker.cacheRead, asker.cacheWrite = tc.read, tc.write
+			e := newSweepSmall(t, "block_fallback: true\n")
+			req := sweepReq()
+			original := schema.MessageText(req.Input[1])
+			rep := &components.Report{}
+			if _, err := e.Offload(req, rep,
+				preExpiryCtx("s", asker, store.NewMemory(store.Options{}))); err != nil {
+				t.Fatal(err)
+			}
+			if rep.Gates["sweep_fallback_blocked"] != 1 {
+				t.Fatalf("block_fallback did not decline (gates: %v)", rep.Gates)
+			}
+			if rep.Gates[tc.wantGate] == 0 {
+				t.Errorf("did not record %s (gates: %v)", tc.wantGate, rep.Gates)
+			}
+			if rep.Gates[tc.notGate] != 0 {
+				t.Errorf("recorded %s for the wrong failure (gates: %v)", tc.notGate, rep.Gates)
+			}
+			// And declining must leave the transcript alone, which is the point of the switch.
+			if schema.MessageText(req.Input[1]) != original {
+				t.Error("a declined ask modified a message")
+			}
+		})
+	}
+}
--- a/components/offload/extract_sweep_test.go
+++ b/components/offload/extract_sweep_test.go
@@ -992,12 +992,13 @@ func TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit(t *testing.T) {
 }
 
 // block_fallback now declines on a partial hit that WROTE, not only on a zero read, so the two cases
-// must stay distinguishable. The counters are what a reader actually gets: on this path the call row
-// is DROPPED (foldFallback returns early because the fallback deliberately never ran, so
-// rec.Component stays "" and Offload's `call.rec.Component != ""` guard discards it), which means the
-// Rejection string built alongside is unreachable today. That is a pre-existing hole of the same class
-// the comment at foldFallback documents for the no-asker path -- recorded here rather than fixed,
-// because adding a row where none existed is a behaviour change this evidence does not cover.
+// must stay distinguishable. This asserts the counters, which is one of two things a reader sees: see
+// TestBlockFallbackDecliningAWritePublishesTheCorrectRejection for the other, the Rejection string
+// itself. (549714c claimed the call row is dropped here because rec.Component stays "" -- that is only
+// true of THIS test's own `rep := &components.Report{}`. In production, pipeline.go sets
+// rep.Component = comp.Name() before Offload ever runs, and the assignment at extract_sweep.go's
+// `r.rec = components.ModelCall{Component: rep.Component, ...}` happens before this decline branch, so
+// the row survives and the Rejection string is published. See the other test for the fix.)
 func TestBlockFallbackDistinguishesAZeroReadFromAWrite(t *testing.T) {
 	for _, tc := range []struct {
 		name        string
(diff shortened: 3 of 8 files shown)


OsherElhadad:
## Review: revert-verified all three fixes, plus two findings the tests do not cover

Reviewed at `f633429a` (base is current `main`, 0 commits behind) in a dedicated worktree. **Verdict: approve with nits.** All three defects are real, correctly fixed, and each regression test fails — naming its own subject — when the original defect is reintroduced *in its exact original form*. Two things are worth acting on: a new panic half-state, and an "unreachable" rationale in a commit message that is factually wrong for the case it is given for.

**Data boundary:** every production figure in the description (the 220,933-request / 16-day corpus, the 29,321/29,372 row split, the $3.42 / 704,925-token prefix write, the band economics) is **not verifiable within the data boundary** — no reviewer here read the service database or any export of it. Those numbers are treated as given; what was checked is the *mechanism* behind each, against the code.

### Real run results

| Command | Result |
|---|---|
| `gofmt -l .` / `go vet ./...` | clean |
| `go build ./...` | clean, 1.9s |
| `go test -count=1 ./...` | **all 29 packages ok**, 1m23s (dash 76.5s, proxy 41.4s) |
| `CGO_ENABLED=1 go test -race ./components/offload/... ./config/... ./internal/extract/... ./dash/... ./proxy/...` | **all 5 ok, 0 races**, 7m57s — **1,723 `=== RUN`**, 0 FAIL, 19 SKIP |

Per-package counts that actually RAN (non-race): `components/offload` **373**, `config` **267**, `internal/extract` **101**.

**The filename trap does not recur.** `go list -f '{{.TestGoFiles}} {{.IgnoredGoFiles}}'`:
- `internal/extract` — `zz_freeleg_test.go` is in **TestGoFiles**, `IgnoredGoFiles=[]`. Compiled, as its name intends.
- `components/offload` — `IgnoredGoFiles=[skeleton.go skeleton_measure_test.go skeleton_safety_test.go]`, pre-existing behind `cg_skeleton`, untouched here.
- `config` — `IgnoredGoFiles=[]`.

Of the 19 SKIPs, one is new: `TestDeterministicOnlyArmOverCapturedBodies` needs `CG_H0_DIR` pointing at a corpus exported from `extraction_calls.before_gz`. That is production data and not synthesisable inside the boundary, so **it is reported as unverified, not as a pass** — it skipped in every run. Worth knowing that this test can only ever be exercised by someone with corpus access.

### Revert-verification — each defect restored to its original form

| Defect | Mutation | Observed failure |
|---|---|---|
| 1 · `Skipped` unset on the early return | removed the `defer` guard and restored the pre-PR shape exactly: bare `return keys, nil` at the inventory floor **plus** the old inline `if changed==0 {rep.Skipped=true}` back before the tail return | `TestSweepDeclinedAtInventoryFloorReportsSkipped` — **both subtests FAIL**: `a component that declined reports Skipped=false, so dash reads mutated=1 for a turn that changed nothing (gates: map[below_output_floor:1 sweep_inventory_below_min:1])` |
| 2 · counter blind to partial-hit writes | restored `if !fellBack && usage.CacheRead == 0 {` as the sole condition | `TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit` FAILS: `an ask that re-created the agent's prefix for 260604 tokens recorded nothing (gates: map[below_output_floor:1 sweep_kept:1])` |
| 3 · `block_fallback` said "read nothing" on a write | same widened-condition revert (it is what routes the write case into `block_fallback` at all) | `TestBlockFallbackDistinguishesAZeroReadFromAWrite` FAILS on `partial_hit_that_wrote`: `block_fallback did not decline (gates: map[below_output_floor:1])`; the `zero_read` subtest correctly still passes |
| *bonus* · the housellm floor guard | set `extract_llm.min_tokens` to 1000, below the derived floor | `TestHousellmFloorClearsTheBandsThatLoseAtEveryK` FAILS: `housellm ships min_tokens: 1000, below the measured floor of 2000. Bands under 2000 cost $32.41-$79.75 per MTok saved against $31.25 on even a top-decile session` |

That last one answers the question worth asking of any test that pins a measured number: **the floor is genuinely derived**, by looping the bands table against `p90TokenValue`, not copied from a log. The failure message names the computed number. A log goes stale like a build; this does not.

Mechanism spot-checks behind the three: `GateN`/`EventN` (`components/component.go:646-654,690-698`) are confirmed no-ops at `n<=0` — the exact mechanism blamed for the silent miscount; `dash/event.go:525` is `row.Mutated = !r.Reverted && !r.Skipped` as stated; and the fixtures use the **exact** production values (`CacheRead=39805` / `CacheWrite=260604` for request 151760), not approximations.

### Findings

**Minor · `components/offload/extract_sweep.go:281-285` — the deferred guard also fires during panic unwind, creating a `Reverted && Skipped` half-state that did not exist before.**

Go runs deferred functions during a panic. If `Offload` panics before `changed` is incremented — the common case, since `changed++` happens deep in phases 1/3 — the defer sets `rep.Skipped=true`, and *then* the panic reaches `pipeline.runOne`'s `recover()` (`components/pipeline.go:101-109`), which sets `rep.Reverted=true` on the same `Report` **without clearing `Skipped`**. Pre-PR this was essentially unreachable: the old inline check sat immediately before the clean tail return, with nothing after it that could panic.

Traced every consumer: `dash/event.go`'s `Mutated` is unaffected (false either way), and `apply/apply.go:88-98`'s `logDecisions` switch tests `Reverted` before `Skipped`, so the human-readable verdict stays right. But the raw `skipped` column (`dash/store.go:416`, `dash/query.go`) will now read `true` on some panic rows where it never could before — a future "count skipped rows" query that does not also exclude `Reverted` would misread it. Not a blocker, and no current derived metric is corrupted. Flagging it because it is precisely the ambiguous-half-state class this PR is otherwise hunting. Cheapest fix: have the defer skip when `rep.Reverted` is set, or clear `Skipped` in `runOne`'s recover.

**Minor · commit `549714c`'s rationale for defect 3 is wrong, and the consequence is an untested published field.**

The commit message says the new `Rejection` string is "unreachable today" because `foldFallback` returns early and `rec.Component` stays `""`, and the shipped test therefore asserts only gates. **That is false for the case it is given for**, and I verified the mechanism myself rather than relaying it:

- `r.rec = components.ModelCall{…}` is assigned at **`extract_sweep.go:812`**.
- The `!fellBack && (usage.CacheRead == 0 || usage.CacheWrite > 0)` branch that writes the new string is at **`:876-894`** — *after* 812.
- `rep.Component` is always non-empty in production: `components/pipeline.go:81` builds `Report{Component: comp.Name(), …}`.
- And `:763`'s `if r.rec.Component == ""` is **not a discard — it fills `Component` in**, precisely so the caller's `call.rec.Component != ""` guard cannot drop the row. Its own comment says so.

So the row is built, retained, and carries the new string. An independent probe test (written, run, deleted) confirmed it: with `rep.Component` set as production sets it, `len(rep.Calls) == 1` and `Calls[0].Rejection` is exactly the new "re-created the agent's prefix instead of reading it" text. The claim *is* correct for the genuinely-early paths (no-asker, ask-failed — those give `len(rep.Calls) == 0`), and the commit conflates those with this reachable one.

Net: the string is live in production **and has zero test coverage of its content**, because the coverage was skipped on a premise that does not hold. Given this PR's own words that this column "has already been mis-analysed twice, so a wrong reason here becomes a wrong finding later," that is the one gap I would close — a `rep.Calls`-asserting test, and a corrected rationale so the wrong reason does not stand in history.

**Minor · docs · `docs/components/extract_llm_sweep.md` (~264-276) is now stale.** Its "## Counters" section is an explicitly exhaustive, operator-facing enumeration of the gates this component raises. It names `sweep_prefix_cache_read_ZERO` and **not** the new `sweep_prefix_cache_write`. This diff touches no docs. To be clear about what does *not* need changing: the counter needs no code wiring — `metrics/metrics.go` and `proxy/promexport.go` both iterate `range cs.Gates`/`r.Gates` generically, so `/stats` and `cg_component_gate_declines_total` already carry it — and `docs/reference/routes.md` correctly needs no row, since this is a per-component gate rather than a snapshot field. It is one line in the component page.

**Nit · the "both exits" coverage claim does not hold.** Both subtests of `TestSweepDeclinedAtInventoryFloorReportsSkipped` reach the **same** early return (`:500`) under two different gate reasons. Neither exercises the **other** exit (`:568`, the tail return after `adjudicate` kept everything) — which is the path whose duplicated inline check this PR *deleted*. Nothing in the suite regression-tests that the tail exit still sets `Skipped`: `TestSweepCountsKeepEverythingSeparatelyFromAFailure` walks that exact shape but never asserts `rep.Skipped`. Low risk, since one `defer` now covers both by construction — but the coverage claim is stronger than the fixture.

**Nit · the headline production shape is untested.** Request 175582 had `CacheRead=0` **and** `CacheWrite=240309` — the one row that fired under the old counter. Probed directly (temporary test, reverted): both `sweep_prefix_cache_read_ZERO` and `sweep_prefix_cache_write` fire together, which reads as the intended two-distinct-facts design rather than a bug. Untested, though, and it is the shape the description leads with.

### Checked and clean

- **Security**: nothing new logs content, keys, paths, session or tenant ids. The new gate name and the branched `Rejection` strings are static literals, never built from request content — no cardinality or log-injection surface. No unbounded growth.
- **`.gitignore`**: `git ls-files | grep -E '\.db(-wal|-shm)?$'` is empty and no `.db*` file appears anywhere in history, so the new rule cannot mask a needed file and there is no ignored-but-tracked half-state.
- **`config/config_more_test.go`** guards a real, derived lower bound against the shipped `housellm` preset's `extract_llm.min_tokens: 3000` — not a tautology (see the bonus mutation above).
- **Consumers of the corrected `mutated` semantics** — all get better numbers with no changes of their own: `dash/event.go:525` (the derivation), `dash/store.go:416`, `dash/query.go:233,249,636,752,767,777` (note `ActedStructural = Mutated - Acted` at `:777` inherits the correction directly), `metrics/metrics.go:234,339,438,517-518,920-922`, `proxy/promexport.go:553-577` (`cg_component_runs_total{outcome="mutated"}`), `dash/ui/app.js:2339,2736,2845,2848,3393,3753` (verdict badges and the "ran" filter). `deploy/grafana/dashboards/context-guru.json` also references it, but that file is not what production mounts, so no prod effect.

### Cross-PR note (#212, #211, #196)

Unlike the previous trio, **these three do not collide**: merged onto `main` in order with **zero conflicts**, and on the merged tree `gofmt`/`go vet`/`go build` are clean with `kvcache`, `components/offload`, `config`, `internal/extract` and `deploy/harbor` all passing. #211 is the only one of the three already based on current `main` (#212 is 70 commits behind, #196 is 20).

OsherElhadad:
Addressed the review at pull/211#issuecomment-5558475749. Pushed `798d40e` to `fix/extract-llm-waste-removal`.

## Verified the finding first

Re-derived the mechanism at this HEAD before changing anything: `components/pipeline.go:81` sets
`rep.Component = comp.Name()` before any component runs; `extract_sweep.go`'s
`r.rec = components.ModelCall{Component: rep.Component, ...}` (around line 812) fills `rec.Component`
before the `block_fallback` decline branch (~876-894) is ever reached; and the "IDENTITY FIRST" block
in `foldFallback` *fills* `Component` in when empty, it does not discard the row. `549714c`'s rationale
("`rec.Component` stays `""`, so the row is discarded") only holds for the *existing tests'* own
`rep := &components.Report{}` construction — not for production. Confirmed the review's read was
correct.

## What changed

- **`components/offload/extract_sweep_test.go`**: added
  `TestBlockFallbackDecliningAWritePublishesTheCorrectRejection`. It sets `rep.Component` the way
  `pipeline.go` actually sets it, then asserts `rep.Calls` has exactly one row with a non-empty
  `Component` and the exact Rejection text. Also rewrote the stale doc-comment on the neighboring
  `TestBlockFallbackDistinguishesAZeroReadFromAWrite` that repeated the wrong "row is DROPPED" claim.
- **`components/pipeline.go` / `components/pipeline_test.go`**: fixed the other thing the review
  called "worth acting on" — the panic half-state. A component's own deferred `Skipped`-guard (the
  pattern this branch's fix #1 introduced) can run during the same panic unwind that reaches
  `runOne`'s `recover`, which sets `Reverted` without clearing `Skipped`. Added `rep.Skipped = false`
  alongside `rep.Reverted = true` in `runOne`'s recover — the one place every component's panic routes
  through — and a regression test, `TestPanicClearsSkippedSetByTheComponentsOwnDefer`.
- **`docs/components/extract_llm_sweep.md`**: added `sweep_prefix_cache_write` to the Counters list
  (the "Minor · docs" nit — one line, no code wiring needed, `metrics/metrics.go` and
  `proxy/promexport.go` already iterate gates generically).
- **PR body**: corrected section 3's "unreachable today" claim in place, and updated the Tests section.

## Revert-verification (source mutated, never the test)

**New Rejection test** — reverted the branched reason to the pre-PR flat string in
`extract_sweep.go` (removed the `usage.CacheWrite > 0` branch entirely, restoring exactly what
`549714c` fixed):
```
go test ./components/offload/... -run TestBlockFallbackDecliningAWritePublishesTheCorrectRejection -v
```
FAILED, naming its own subject:
```
extract_sweep_test.go:1066: Rejection = "the prefix ask read nothing from cache and block_fallback is
set; declining rather than paying again for a full-price transcript read", want "the prefix ask
re-created the agent's prefix instead of reading it and block_fallback is set; declining rather than
paying again for a full-price transcript read"
```
Restored the file, rebuilt, reran: PASS.

**Panic half-state test** — removed the new `rep.Skipped = false` line from `pipeline.go`'s recover:
```
go test ./components/ -run TestPanicClearsSkippedSetByTheComponentsOwnDefer -v
```
FAILED, naming its own subject:
```
pipeline_test.go:139: Reverted && Skipped both true: {Component:boomafterskipped Kind:reformat
TokensBefore:3 TokensAfter:3 ... Skipped:true Reverted:true ... Err:panic: kaboom ...}
```
Restored the file, rebuilt, reran: PASS.

## Ready-checklist

- `gofmt -l .` — clean.
- `go vet ./...` — clean.
- `go build ./...` — clean.
- `go test -count=1 ./...` — **all 29 packages ok**. Verbose run: **2,548 `=== RUN`**, 0 FAIL, 28 SKIP.
  `components/offload` specifically: **374 `=== RUN`** (was 373 in the review's own non-race count,
  +1 for the new test). `components`: **33 `=== RUN`** (+1 top-level for the new panic-half-state
  test, plus its assertions).
- `CGO_ENABLED=1 go test -race -count=1 ./components/offload/...` — **ok**, 0 races, 106.2s.
- `git merge-tree --write-tree HEAD origin/main` — resolves cleanly to a tree (no conflict markers);
  still merges onto `main` with no conflicts.
- Pushed as a normal fast-forward: `f633429..798d40e fix/extract-llm-waste-removal` on
  `osher-fork` (no rebase, no force).

## Decided NOT to change, and why

The review's two remaining **"Nit"** items (as opposed to the "Minor" items above) are left as-is:

- **The "both exits" coverage claim** (`TestSweepDeclinedAtInventoryFloorReportsSkipped` only
  exercises the early-return exit, not the tail-return exit `TestSweepCountsKeepEverythingSeparatelyFromAFailure`
  walks). The reviewer calls this "low risk, since one `defer` now covers both by construction" and
  flags it as a documentation-strength mismatch, not a defect — no behavior to fix, only a stronger
  test to write. Left for a follow-up rather than expanding this PR's test surface further.
- **The headline production shape untested** (`CacheRead=0` and `CacheWrite=240309` together, request
  175582). The reviewer already probed this directly (temporary test, reverted) and confirmed both
  gates fire together as intended — a design confirmation, not a bug. Same call: worth a permanent
  test eventually, not a blocker here.

Both are testing-coverage suggestions the review itself marks as lower severity than the two "worth
acting on" items (the wrong rationale and the panic half-state), which this push addresses.

OsherElhadad (commented):
## Review

I checked out `fix/extract-llm-waste-removal`, resolved the merge conflicts with `main`, and verified each of the three claims independently.

### Claim (a): declined turns wrongly reported as `mutated`
Confirmed present on `main` before this branch (`components/offload/extract_sweep.go`, the inventory-floor early return at `len(cands) < e.minInventory` never touched `rep.Skipped`, only the tail return did). This branch's fix — a single deferred guard (`defer func() { if changed == 0 { rep.Skipped = true } }()`, now at extract_sweep.go:703 post-merge) — covers both existing exits and any future one, which is the right shape for this bug (a per-exit copy is exactly what caused it). `TestSweepDeclinedAtInventoryFloorReportsSkipped` reproduces both exits and is table-driven.

### Claim (b): prefix "bust" (partial hit that writes) undetected
Confirmed: `sweep_prefix_cache_read_ZERO` tested only `usage.CacheRead == 0`, so a partial hit that still wrote a large prefix went uncounted — this is real money (the PR's production numbers: $2.26 of $3.42 uncounted). The fix widens the trip condition to `usage.CacheRead == 0 || usage.CacheWrite > 0` and separates the two into distinct counters (`sweep_prefix_cache_write` vs `sweep_prefix_cache_read_ZERO`) rather than collapsing them, which is the right call since they're different failures with different remedies. `TestSweepPrefixWriteTripsTheCounterEvenOnAPartialHit` explicitly checks the write counter fires *and* the zero-read counter does not on a partial hit, so the two stay distinguishable.

### Claim (c): `block_fallback` rejection string named the wrong failure
Confirmed and fixed correctly — widening the counter condition in (b) put the write-only case on the same branch that unconditionally said "read nothing from cache," which is false when `CacheRead` is nonzero. The fix branches the `Rejection` string on `usage.CacheWrite > 0`. Also good: the PR corrects its own earlier premise that this string was unreachable in production (it isn't — `pipeline.go`'s `runOne` sets `rep.Component` before any component runs, so the row survives `Offload`'s `call.rec.Component != ""` guard) and adds `TestBlockFallbackDecliningAWritePublishesTheCorrectRejection`, which sets `rep.Component` the way the pipeline does. That correction is exactly right — I checked `components/pipeline.go`'s `runOne` and `extract_sweep.go`'s `r.rec = components.ModelCall{Component: rep.Component, ...}` and confirmed the row is not discarded.

### Bonus fix worth calling out
`TestPanicClearsSkippedSetByTheComponentsOwnDefer` closes a real half-state that fix (a)'s own deferred guard made reachable: if a component panics after its own defer has already set `Skipped = true`, `runOne`'s `recover()` sets `Reverted = true` without clearing `Skipped`, so a `Report` could carry both. `dash/event.go`'s `Mutated` derivation (and any "skipped rows" query) would misread that. The fix clears `Skipped` alongside `Reverted` in `runOne`. Good catch, and the test targets it precisely (a component that panics after setting Skipped in its own defer).

### Security
No sensitive data leaks in the new rejection strings — they describe token counts and gate names only. Noted (pre-existing, not introduced by this PR) that the unrelated `sweep_ask_failed`/`block_fallback` path a few lines below embeds `err.Error()` into `rec.Rejection`, which is persisted to `extraction_calls.rejection`; worth a follow-up to confirm no transport error ever carries request content, but out of scope for this PR since it's not touched here.

### Docs
`docs/components/extract_llm_sweep.md` was deleted on `main` during a docs restructuring (content moved to `docs/components/advanced-offload.md`) — this created the one real merge conflict. I resolved it by dropping the deleted file (matching main's restructure) and adding `sweep_prefix_cache_write` to the counters list in `advanced-offload.md` instead, which is where the equivalent list now lives.

### Verification performed
- `go build ./...`, `go vet ./...` clean after merge.
- `go test ./components/offload/... ./components/... ./config/... ./metrics/...` — all pass, including all 7 new tests from this PR run individually with `-v`.
- `go test ./...` — passes except 3 pre-existing failures (`TestCalculatorPricesFromTheRowsModelNotABlend`, `TestNavHashCompatibility`, `TestInstallReportsPATHFromTheSourceFallbackToo`) confirmed present on a pristine clone of `main` with zero relation to this PR's files.
- Read the pre-fix code on `main` directly to confirm each of the three defects independently (not just from the PR's own description) before checking the fix.

### Merge conflicts
Rebase/merge with `main` is done and pushed (`main` had grown `extract_sweep.go` from ~1,150 to ~1,850 lines in the interim — a large but line-shifted diff, not a conceptual conflict; the auto-merge applied all three code fixes onto the current file correctly, verified by grep + reading the merged result). Branch is now `MERGEABLE`.

Nothing else additional to fix — the diff is tight and each claim's test would fail without its corresponding fix (verified by reading the pre-fix code, not just trusting the PR description).

PR #227, opened 2026-09-08
fix(plugin): make the printed recovery command the real launch path

### LINKED ISSUE #225: check-proxy.sh's recovery command drops --config, --anthropic-upstream and --idle-exit, so pasting it turns keep-alive off and bypasses a configured gateway
Found during the round-4 review of #222, which fixed the same class of defect in `start-proxy.sh`'s success note (#224). This one is in the sibling script and is pre-existing on `main` from #196 / #212 and the upstream work, so it gets its own issue rather than riding that branch.

## What happens

When a session's requests are being routed to a proxy that is not answering, `check-proxy.sh` prints a recovery command for the user to paste (`scripts/check-proxy.sh:71-74`):

```bash
To fix it now, in a terminal (the dashboard flags matter: without them /dashboard/ is a 404, and
this proxy would then hold the port for the whole --idle-exit window with no way to notice why):
  context-guru-proxy --listen 127.0.0.1:${PORT} --preset ${CLAUDE_PLUGIN_OPTION_PRESET:-cache} \
    --dashboard --dashboard-db "${STATE_DIR}/dashboard-${PORT}.db"
```

That command does not reproduce the proxy the plugin itself starts. `check-proxy.sh` contains no mention of `UPSTREAM` or `--config` at all (0 matches), while `start-proxy.sh` passes both. Three configured things are silently dropped:

| Dropped | `start-proxy.sh` | Consequence of pasting the printed command |
|---|---|---|
| `--config <keepalive yaml>` | `:264-267` | **keep-alive comes back OFF** after the user explicitly turned it on. The file exists and is what `start-proxy.sh` keys on, so the hook-started proxy had it and this one does not. |
| `--anthropic-upstream` | `:248-251` | the proxy goes **straight to api.anthropic.com**, bypassing the gateway that holds the credential. `plugin.json`'s own `upstream` description: "on at least one pod, rewriting model names: bypassing it sends model ids the real API has never heard of and every request fails." |
| `--idle-exit` | `:273` | the flag defaults to `0` = never exit (`cmd/context-guru-proxy/main.go:98`), so the pasted proxy runs until killed — the opposite of the `idle_exit` option's promise that "nothing is left running on your machine". |

And `--preset ${CLAUDE_PLUGIN_OPTION_PRESET:-cache}` names the plugin option, which is **not** what is in effect whenever a keep-alive config exists, because `--config` replaces the preset entirely (`loadConfig` returns `config.Load(path)` and never consults `preset`). That is exactly the misreport #224 fixed for the success note, in the same three lines that also drop the config.

## Why it matters more than a wrong string

This text is shown at the one moment the user is already broken and least able to check the plugin's reasoning: requests are hanging with no error. It is presented as "to fix it now", so it will be pasted. The result is a proxy that answers `/healthz` — so every signal the plugin surfaces afterwards says healthy — while quietly running with a feature the user pays for turned off, and, for anyone behind a gateway, failing every request for a reason nothing in the output points at.

## Suggested fix

Build the printed command from the same values `start-proxy.sh` uses, rather than re-deriving a subset of them: include `--config` when `keepalive-${PORT}.yaml` exists, include `--anthropic-upstream` when an upstream is configured, include `--idle-exit`, and report the preset actually in effect (the config file's, when one is in play) the way `start-proxy.sh` now does after #224.

Worth a test in the shape of `TestStartProxyReportsThePresetActuallyInEffect`: with a keep-alive config and an upstream configured, assert the printed command contains `--config` and `--anthropic-upstream` and does not name the option's preset. The two scripts deriving the same launch line independently is the root cause, so a test that pins them together is more valuable than one that pins the string.

## Note

Reported by the reviewing session rather than fixed there, per the repo rule about defects on `main` unrelated to the branch under review. Related: #224 (the same misreport in `start-proxy.sh`'s note) and #223 (per-option fallback discovery, which makes `CLAUDE_PLUGIN_OPTION_PRESET` legitimately a `plugin.json` default rather than anything the user set).

### LINKED ISSUE #226: check-proxy.sh's printed recovery command drops --config and --upstream and names the option's preset, so pasting it silently disables keep-alive and bypasses the gateway
Found during the round-4 review of #222 and handed to the operator to place rather than folded into that branch. Pre-existing on `main` from the keep-alive work (#196 / #212) and the upstream work.

## The defect

When nothing answers on the port, `check-proxy.sh` prints a manual recovery command (`scripts/check-proxy.sh:73`):

```bash
context-guru-proxy --listen 127.0.0.1:${PORT} --preset ${CLAUDE_PLUGIN_OPTION_PRESET:-cache} \
  --dashboard --dashboard-db "…/dashboard-${PORT}.db"
```

Three problems, all silent for whoever pastes it:

1. **It names the plugin option's preset**, which is the same misreport #224 fixed in the sibling script. `CLAUDE_PLUGIN_OPTION_PRESET` also does not reach a Bash tool call, so an agent pasting this gets `cache` regardless of configuration (#221's class).
2. **It omits `--config`.** `start-proxy.sh` passes `--config "${STATE}/keepalive-${PORT}.yaml"` whenever that file exists. A user who explicitly enabled keep-alive — and is paying for the idle pings — gets a proxy with **keep-alive off**, and nothing says so. Because `--config` also *replaces* the preset, the pasted `--preset` is not even the thing that would have been in effect.
3. **It omits `--upstream`.** `check-proxy.sh` never mentions the upstream at all. If one is configured, the recovery proxy **bypasses the gateway holding the user's credential** — which `install`'s own prose says makes every request fail on at least one platform whose gateway rewrites model names.

So the printed remedy quietly produces a differently-configured proxy than the one it is replacing, at the moment the user is already troubleshooting.

## Suggested fix

Derive the recovery command from the same inputs `start-proxy.sh` uses, rather than restating a subset: include `--config` when the keep-alive file exists, include `--upstream` when one is configured, and either read the preset from the config in play or omit `--preset` entirely when `--config` is passed (since it has no effect then).

The cheaper and more robust alternative is to stop printing a hand-rolled command at all and point at `start-proxy.sh` — the script that already gets all of this right — e.g. `"${CLAUDE_PLUGIN_ROOT}/scripts/start-proxy.sh" --unrouted`. That cannot drift from the real launch path the way a duplicated command list can, and this issue is precisely that drift.

Worth a test in the shape of `TestStartProxyReportsThePresetActuallyInEffect`: with a keep-alive config present, the printed recovery command must not silently drop it.

## Related

- #221 — the `CLAUDE_PLUGIN_OPTION_*`-in-a-Bash-tool-call class
- #224 — the same preset misreport in `start-proxy.sh`'s success note
- #223 — the per-option discovery fallback

amiddavid:
Closes #225. Closes #226.

`check-proxy.sh` runs on `UserPromptSubmit` and, when nothing answers on the routed port, prints a command for the user to run in a terminal. That command was a hand-rolled `context-guru-proxy --listen … --preset …` — a second copy of what `start-proxy.sh` builds — and it had drifted from the real launch path **four ways at once**:

| Omission | What pasting it did |
|---|---|
| named `CLAUDE_PLUGIN_OPTION_PRESET` | reported a preset `--config` replaces anyway |
| no `--config` | **turned keep-alive off** for someone who enabled it and was paying for the pings |
| no `--anthropic-upstream` | **bypassed a configured gateway** — on a platform whose gateway rewrites model names, every request then fails |
| no `--idle-exit` | left a proxy that **never exits**, holding the port |

All silent, at the moment the user is already troubleshooting a hung session.

## Delegation, not a repaired flag list

A corrected list would still be a second copy of `start-proxy.sh`'s command line — and a second copy is precisely what drifted, since those four flags were each missed as they were added. The note now points at `start-proxy.sh`, which this script already invokes fourteen lines earlier for its automatic recovery attempt. Anything that script gains is picked up for free.

Where there is no starter to point at (wrong `CLAUDE_PLUGIN_ROOT`, half-installed plugin), the note says so and directs the user at `/context-guru:install`. It deliberately does **not** fall back to a hand-written command: reintroducing the duplicate in the one case where nothing can be verified is how this defect would come back.

## Why `--preset` and `--idle-exit` gained argument forms

A bare `start-proxy.sh` would **silently do nothing** from a terminal. Settings `env` values reach hook processes, not interactive shells, so `ANTHROPIC_BASE_URL` is absent there and the script's own gate declines and exits 0. Every `CLAUDE_PLUGIN_OPTION_*` is absent for the same reason, so it would also start on 8787 with the default preset — #221's class, reintroduced by the fix.

So the printed command must carry `--unrouted`, the port, and the preset explicitly. An env prefix was the obvious alternative and is the wrong shape — `start-proxy.sh`'s own gate comment records what it cost a human:

> A human pasted a long env-prefixed invocation, the paste split across two lines so the assignments became a no-op statement, the script ran unrouted and exited here without a word — no output, no log, no pidfile. Three rounds of guessing followed.

`--port`, `--upstream` and `--bin` already existed as arguments for that reason. `--preset` and `--idle-exit` were the two left, and now match, with the same `takes_value` validation and the same argument-beats-option precedence as `--bin`.

## The test runs the command, it doesn't read it

`TestCheckProxyRecoveryCommandIsTheRealLaunchPath` extracts the printed command and **executes** it — against a keep-alive config and a configured upstream — then asserts on the argv the proxy is actually launched with, plus the pidfile (a recovery proxy `uninstall` can't find is unstoppable). The second run clears every `CLAUDE_PLUGIN_OPTION_*`, because that is what a terminal looks like and is the whole reason the flags must be there.

```
--listen 127.0.0.1:40053  --idle-exit=90m  --dashboard  --dashboard-db …
--anthropic-upstream http://gw.example:4000  --config …/keepalive-40053.yaml
```

A wording assertion would have passed on all four original omissions.

## One existing test changed, not deleted

`TestCheckHookFinishesInsideItsOwnTimeout` asserted the literal `--dashboard`, commented *"the printed recovery command must not recreate the 404 dashboard"*. The note no longer restates flags, so that string is gone **by design** — and a failing assertion is the right place to be made to justify a change like this.

The property did not move to prose. It moved somewhere stronger: the new test asserts `--dashboard` in the **launched argv**. What the timeout test checks now is that this path still hands the user something runnable (`start-proxy.sh`, `--unrouted`), which is what it can observe from where it stands.

## Verification

Eval box, Go 1.26.4, `go test -race -count=1`, `CGO_ENABLED=1`. `-count=1` throughout, since the test cache does not track shell scripts read at runtime. **Eight mutations**, each proven to land, each failing the test named for it, each file restored byte-identical (`cmp`):

```
hand-rolled command line restored     -> the-real-launch-path        FAILED
--unrouted dropped                    -> FAILED (self-gates, launches nothing)
--port dropped                        -> FAILED (recovers on 8787, not the configured port)
--upstream branch dropped             -> FAILED (missing --anthropic-upstream)
--idle-exit dropped                   -> FAILED (missing --idle-exit)
start-proxy --idle-exit arg removed   -> takes-preset-and-idle-exit  FAILED (option beat argument)
start-proxy --preset arg removed      -> takes-preset-and-idle-exit  FAILED
no-starter branch given a fallback    -> no-starter-to-point-at      FAILED
```

Full suite: **29 packages ok, 0 failures, exit 0**. `gofmt -l` empty, `go vet` clean.

## On the two issues

#226 is my own duplicate of #225, filed before I saw it. #225 is the better of the two — it caught `--idle-exit`, which I had missed. Both are closed here and now cross-reference each other.

### DIFF
--- a/context-guru-plugin/plugin_test.go
+++ b/context-guru-plugin/plugin_test.go
@@ -950,7 +950,17 @@ func TestCheckHookFinishesInsideItsOwnTimeout(t *testing.T) {
 	for _, want := range []string{
 		"nothing is answering there",
 		"Your request will hang with no error message",
-		"--dashboard", // the printed recovery command must not recreate the 404 dashboard
+		// This used to assert the literal "--dashboard", because the note restated the proxy's whole
+		// command line and a copy of it that forgot those flags left the user with a 404 dashboard.
+		// The note no longer restates anything — it points at start-proxy.sh, which always passes them
+		// — so the string is gone by design rather than by regression.
+		//
+		// The property itself did not move to prose: TestCheckProxyRecoveryCommandIsTheRealLaunchPath
+		// RUNS the printed command and asserts --dashboard in the argv the proxy is actually launched
+		// with, which is strictly stronger than finding the word in a paragraph. What is left to check
+		// here is that this path still hands the user something runnable at all.
+		"start-proxy.sh",
+		"--unrouted",
 	} {
 		if !strings.Contains(flat, want) {
 			t.Errorf("the diagnostic is missing %q; output was:\n%s", want, out)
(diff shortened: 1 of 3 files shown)


amiddavid (commented):
## Review — round 1

Delegating instead of restating is the right fix, and the two decisions you flagged as mid-implementation
changes of plan both check out. Two findings, and the first one is on the exact thing you asked me to be
suspicious of — your instinct was right, but not for the reason you gave.

### 1. The `--dashboard` property did not move to the argv test. It is unasserted in both.

`--dashboard` is a **prefix of `--dashboard-db`**, and `strings.Contains` cannot tell them apart. So the
needle in `TestCheckProxyRecoveryCommandIsTheRealLaunchPath` is satisfied by the `--dashboard-db` flag
that follows it, and the flag whose absence caused the 404 can be deleted outright with everything green:

```
delete `--dashboard \` from start-proxy.sh's launch line
  -> TestCheckProxyRecoveryCommandIsTheRealLaunchPath   PASS
  -> TestCheckHookFinishesInsideItsOwnTimeout           PASS
  -> argv actually launched:
     --listen 127.0.0.1:36045 --idle-exit=90m --dashboard-db /…/dashboard-36045.db
       --anthropic-upstream http://gw.example:4000 --config /…/keepalive-36045.yaml
     (no bare --dashboard anywhere)
```

And it matters, because the flag is `flag.Bool("dashboard", envBool("DASHBOARD", false), …)`
(`cmd/context-guru-proxy/main.go:106`) — default **false**. Without it the dashboard is off and
`/dashboard/` 404s, which is precisely what the assertion you retired existed to prevent. The retirement
reasoning was sound; the replacement just cannot fail. Net effect of the swap is that the property went
from weakly asserted to not asserted.

Fix is a token boundary rather than a substring — split the argv on spaces and compare, or search for
`" --dashboard "` in a space-padded copy. Only this needle has the collision: `--listen`,
`--anthropic-upstream`, `--config` and `--idle-exit=` are all followed by their values, so they are
unambiguous as written.

This is the one I would hold the merge for. It is also the third instance in two PRs of a guard matching
a *form* of the thing — label, sentence, and now prefix — which might be worth a habit rather than a
third fix: when a needle is a flag name, assert the token.

### 2. The printed command interpolates unquoted, so a configured upstream can be silently truncated

The starter path is quoted; `PRESET`, `IDLE_EXIT` and `UPSTREAM` are not. This is a command a human
pastes into a shell, so a value carrying shell syntax makes the pasted command do something other than
what it reads as. `&` in a gateway URL is the realistic case, and the outcome is the failure mode this PR
exists to remove:

```
CLAUDE_PLUGIN_OPTION_UPSTREAM=http://gw.example:4000/v1?tenant=acme&mode=chain
PRINTED:  "…/start-proxy.sh" --unrouted --port 35459 --preset cache --idle-exit 24h \
            --upstream http://gw.example:4000/v1?tenant=acme&mode=chain
PASTED:   exits 0, proxy starts, no warning
LAUNCHED: --anthropic-upstream http://gw.example:4000/v1?tenant=acme     <- &mode=chain gone
```

The `&` backgrounds the invocation and the rest of the URL becomes a separate statement, so the user gets
a running proxy chained to a **different gateway than the one they configured** — silently, while
troubleshooting. A space does the same by word-splitting: `--upstream http://gw.example:4000/a b` starts
a proxy pointed at `…/a`.

`--preset "${PRESET}" --idle-exit "${IDLE_EXIT}" --upstream "${UPSTREAM}"` closes it. Worth a row in the
new test with an `&` in the upstream, since that is the value most likely to carry one and the assertion
you already have on argv catches it exactly.

### The three things you asked about

**Gate semantics: your reading is correct.** `--unrouted` sets `START_UNROUTED=1` → `FORCE=1`, which
skips the `case "${ANTHROPIC_BASE_URL:-}"` block entirely (`start-proxy.sh:191-210`). Without it, an
unset `ANTHROPIC_BASE_URL` falls to the `*)` branch, which writes one dated `declined:` line to the
port's log and `exit 0` with **nothing on stdout** — so a bare invocation from a terminal is silent and
does nothing, exactly as you describe. The whole recovery path does depend on this, and it holds.

**The widened argument surface is justified, not scope creep.** Delegation is only usable if every value
the hook can see has an argument form, and the alternative is the env prefix this same file records
costing a human three rounds of debugging. Two flags, parsed by the existing `takes_value` machinery,
with the same precedence as `--bin`, and the precedence note about `--config` still winning over
`--preset` is accurate (`loadConfig` returns `config.Load(path)` and never reads `preset`). I would have
done the same thing. Mutation-wise the parsing is real: dropping `--unrouted` from the printed command
fails both the new test ("offers no runnable recovery command") and the timeout test.

**Relaxing another test's assertion was the right call, and I would have approved it as reasoned — see
finding 1 for why it did not land.** The principle you applied is correct: an assertion on the launched
argv is strictly stronger than the same word appearing in a paragraph, and moving it out of a test about
timeouts is a tidying rather than a loss. Keep the swap; make the new needle actually discriminate.

### Verified

`go test -count=1 -race ./...` on the plugin package: **ok, 36.5s**, `gofmt -l` empty, `go vet` clean
(eval box, Go 1.26.4, `CGO_ENABLED=1`). New tests all pass as written. Both scripts restored
byte-identical (`cmp`) after every mutation and my probe file removed from the box. Everything above ran
under `-count=1`.

amiddavid:
## Both findings fixed — e993f70

### 1 (blocker) — and it's worse than a vacuous assertion

You're right, and right about the mechanism: `--dashboard` is a prefix of `--dashboard-db`, so the substring is satisfied by the flag that follows it. Verified directly before fixing — the needle matches with the bare flag deleted.

What makes this one bad is *where* it happened. That assertion is the one I retired the timeout test's literal `--dashboard` check **in favour of**, arguing that asserting the launched argv is strictly stronger than finding a word in a paragraph. The argument was right; the implementation didn't deliver it. So for one round the property had no working check anywhere — and I'd reported it as strengthened. Same shape as the `0600` fixture in #222's `TestSettingsPreservesFileMode`, which is how this whole line of work started.

Flags are matched as whole tokens now (`hasArgvFlag`), and value-carrying flags by **value** (`argvFlagValue`), so a *truncated* value fails and not only a missing flag.

### 2 — fixed, and your example is the one that convinced me

The printed command isn't run by the script; it's re-parsed by the user's shell. `http://gw.example:4000/v1?tenant=acme&mode=chain` backgrounds at the `&`, so the paste exits 0, starts a proxy, and chains it to `…?tenant=acme` — silent partial success, the same failure mode as everything else in this PR.

All three values are single-quoted now. Single rather than double deliberately: double quotes still expand `$` and backticks in the user's shell, so a value containing either would be silently rewritten at paste time.

The test gains an `&`-upstream subtest and asserts the **full** upstream reached the proxy, since a presence check passes on the truncated value:

```
--anthropic-upstream http://gw.example:4000/v1?tenant=acme&mode=chain
```

### Verification

Ten mutations under `-count=1`, each proven to land, each reverted byte-identical. The two that matter:

```
--dashboard deleted from start-proxy's launch line -> FAILED   (green before this change)
the upstream interpolated unquoted                 -> FAILED   (the & subtest catches it)
```

The other eight re-confirmed unchanged. Full suite **29 ok / 0 fail / exit 0**, `gofmt`/`vet` clean.

I also had to repair four stale anchors in my own mutation script after the quoting change. Worth mentioning because a stale anchor is itself a way to test nothing — the script asserts the mutation landed before trusting the result, which is what caught it.

### Thanks for confirming the two I'd flagged rather than knew

The gate reading (`--unrouted` → `FORCE=1`, otherwise one `declined:` line and a silent exit 0) and the widened argument surface. I'd raised both as things to check rather than assert, and having them independently confirmed is worth more than my own second look.

amiddavid (commented):
## Review — round 2 (e993f70). Done from my side.

Both fixes verified by re-running the mutations that were green last round:

```
delete `--dashboard \` from start-proxy.sh's launch line
  -> FAILS both subtests: "missing --dashboard … [--listen … --idle-exit=90m --dashboard-db …]"
     (was PASS before hasArgvFlag)
unquote the upstream in the printed command
  -> FAILS: --anthropic-upstream = "http://gw.example:4000/v1?tenant=acme",
     want "http://gw.example:4000/v1?tenant=acme&mode=chain"     (was PASS)
```

Plugin package `go test -count=1 -race ./...` **ok, 35.5s**, `gofmt -l` empty, `go vet` clean. Both
scripts restored byte-identical.

Two of your choices are better than what I asked for, and worth recording as such:

- **Asserting value-carrying flags by VALUE, not presence.** I only asked for a token match on
  `--dashboard`. Matching `--config`, `--listen` and `--anthropic-upstream` on their values is what makes
  the `&` mutation fail on the *content* rather than on absence — a presence check passes happily on a
  truncated URL, which is the whole defect.
- **Single quotes rather than double.** Double quotes would still expand `$` and backticks in the pasting
  user's shell, so a value containing either would be rewritten at paste time; single quotes suppress the
  lot. The apostrophe caveat is the right trade: an upstream containing `'` would produce a visibly broken
  paste rather than a silently different proxy, and loud beats silent here.

### Your #228 question, checked

You flagged one thing in #228 as unverified: whether `/context-guru:uninstall` removes
`dashboard-<port>.db`. It does not remove it automatically — **by design, and the skill already says so.**
`skills/uninstall/SKILL.md:126-135` step 3 ("Offer, do not assume, the rest") offers "Delete the state
directory — `~/.local/state/context-guru` holds the pidfile and the dashboard database … Nothing reads it
once the proxy is gone", and step 4 requires naming anything the user declined to remove, on the stated
grounds that "an uninstall that quietly leaves things behind is the reason people distrust installers".

So #228's premise needs narrowing rather than dropping: an uninstall leaves the DB behind only when the
user declines the offer, and it is never silent. What the offer does not say is **how large it might
be** — the defaults really are 7 days and 512 MiB (`main.go:110,112`, `dash/capture.go:114`), so somebody
declining is not told they may be keeping half a gigabyte. Naming the size, or the actual `du` of the
file, in that offer is the whole fix, and it is a one-line change in the skill rather than anything about
`--dashboard`'s role. Worth editing the issue to say that, since as written it reads as a leak.

### Verdict

Nothing outstanding on #227. Every flag the recovery command must carry now fails a test when dropped or
corrupted, the delegation removes the duplicate rather than repairing it, and the missing-starter branch
refuses to invent a command line. Two rounds used; I would merge this.

As before: I cannot press Approve on a PR from the same account, so read this as the approval.

PR #236, opened 2026-09-10
feat(plugin): add an escape hatch that works when Claude Code cannot

amiddavid:
## Why

`/context-guru:uninstall` is a **skill**, so it needs a working session. The failure it exists for removes that: every request goes through a local proxy, so a dead port or a credential the upstream rejects fails *every* API call — and the agent that would remove the routing cannot reach a model to be asked.

That is not hypothetical. An install left a colleague at 401 on every request, with the documented undo path unavailable for exactly the reason he needed it. His actual fault turned out to be two credential variables set at once and one line in a shell rc — which a settings-file undo would not have fixed either.

## What this adds

**A plain-`sh` hatch, installed outside the plugin.** `settings.py` now establishes three things before any routing key exists, at the single choke point every write funnels through (`save()` — `add` alone has six exits that write):

1. an `O_EXCL` copy of each settings file, taken once, never overwritten, **never pruned**. The existing `*.context-guru-backup-*` files are capped at 10 and *both* add and remove write one — so on a machine that has installed and uninstalled a few times, the backup holding the user's pre-context-guru state is the **first** one deleted. A test runs twelve cycles and asserts the copy is still byte-identical.
2. a record of which files were edited and whether each existed beforehand. A file the install *created* is put back by deleting it; nothing else can tell the difference.
3. `reset.sh`, copied **out of the plugin** into `~/.local/state/context-guru/` (and `~/.local/bin` for `PATH`). A recovery tool that lives inside the thing that broke is gone with `/plugin uninstall`, a marketplace refresh, or a wiped plugin cache.

The hatch needs no Claude, no proxy, no network, no Python, no plugin code — asserted by a test that greps its own source. Restore is `cp`. It copies the current file aside first so running it is itself reversible, verifies afterwards instead of trusting `cp`, and a second run is a real no-op.

**It reports credentials, since that was the real fault.** Whether both credential variables are set, whether `ANTHROPIC_BASE_URL` is exported in the *shell* (where no settings file can override it), and the `file:line` of every `ANTHROPIC_*` assignment in the rc files — locations only, values replaced. A test plants a secret and greps the output for it.

**The dead-proxy hook now names the hatch.** That note prints on the prompt that is about to hang, and its advice used to end in "run `/context-guru:uninstall` from a session that still works" — precisely what its reader does not have.

**`add` refuses the machine-wide file without `--user-scope`.** Project scope was already the documented default, but *only* documented: the script wrote whatever `--file` it was handed. A prompt is not a guardrail against a machine-wide lockout, and the asymmetry is the whole point — a project-scope mistake costs one project, the same mistake in `~/.claude/settings.json` takes out every session the user would use to fix it. Removal is deliberately not gated; uninstall must be able to clean every scope.

**Already-routed installs get a hatch too.** A re-run of the install on a routed project reports `result=unchanged` and never reaches `save()` — so without this, the machines that most need a hatch (everyone who installed before it existed) would never get one. It is honest there: no pre-edit copy exists, so it names the file, points at the timestamped backups, and refuses to auto-restore one that may hold a later state.

## Two defects the tests caught while being written

- The pre-edit copies were created as **dotfiles** — the parent directory of a real target is `.claude`, so the natural name is `.claude-settings.local.json.<hash>.original`. Invisible to `ls` and to every glob, in a directory whose purpose is that a panicking human can read it.
- The no-record path printed the heading `check these files by hand:` followed by **nothing at all** when it matched none of its candidates — the exact uselessness that branch exists to prevent.

## Testing

- **80 tests** in the plugin package, `go vet` clean. 15 are new. Go is not installed on the laptop this was written on, so the suite ran in a `golang:1.26` container.
- The shared `settings()` helper now pins `CONTEXT_GURU_STATE` and `HOME`. Without that, `go test` wrote into the developer's own `~/.local/state/context-guru` and rewrote the copy on their `PATH` — a test suite for a recovery tool should not be able to disturb the developer's recovery tool.
- An **isolated end-to-end run against a real Claude Code 2.1.267** (its own `CLAUDE_CONFIG_DIR`, `HOME` and state dir, verified not to touch the real config): plugin installed from the local marketplace, install steps driven in order, routing landing in the project file with the machine-wide `env` block byte-identical throughout, the scope refusal, the proxy killed so the hook fires, then recovery by pasting the hatch path. Also re-verified on macOS `sh`, since the container runs `dash`.
- **No model calls anywhere.** The proxy binary is stubbed (no Go toolchain locally), which is sufficient because everything under test keys off `/healthz`. What is therefore *not* exercised is the install skill's own prose — the model's judgment about scope and chaining — and a real request through a real proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,11 @@ needed on a Pro/Max subscription** ([details](docs/how-to/install-plugin.md)):
 /context-guru:install
 ```
 
+**If it ever breaks and Claude cannot fix it:** `~/.local/state/context-guru/context-guru-reset`
+undoes the routing from a plain terminal — no working Claude session, no proxy, no network. Routing
+every request through a local proxy means a failure there fails every request, including the ones the
+uninstall skill would need, so the way out cannot itself be a skill.
+
 `/reload-plugins` is what makes the `/context-guru:*` skills exist in this session; without it the
 last line answers `Unknown command`. A new session does the same thing.
 
--- a/context-guru-plugin/plugin_test.go
+++ b/context-guru-plugin/plugin_test.go
@@ -4184,3 +4184,190 @@ func TestRemoveIsNeverGatedByScope(t *testing.T) {
 		t.Errorf("removal damaged the machine-wide file")
 	}
 }
+
+// --- fixes from review round 1 of #236 -----------------------------------------------------
+
+// TestRemoveFirstNeverProducesAnOriginalHoldingRouting is the assertion the reviewer asked to have
+// pinned, and it names the defect exactly: record_touch() runs from save(), and save() is ALSO
+// cmd_remove's write path — so when the first recorded touch of a file was a REMOVAL, the O_EXCL
+// "copy taken before the first edit" was a copy of the ROUTED file, permanently. The hatch then
+// faithfully restored routing into an unrouted project.
+//
+// Who reaches it: every pre-hatch install that upgrades and then uninstalls, and anyone whose state
+// directory was wiped between install and remove. That is the population this whole feature is for.
+func TestRemoveFirstNeverProducesAnOriginalHoldingRouting(t *testing.T) {
+	state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+	path := filepath.Join(proj, "settings.local.json")
+	// A file as a PRE-HATCH install left it: routed, with the metadata that version recorded, and no
+	// pre-edit copy anywhere because that version did not take one.
+	writeJSON(t, path, map[string]any{
+		"env":           map[string]any{"ANTHROPIC_BASE_URL": ourURL},
+		"permissions":   map[string]any{"allow": []string{"Bash(ls:*)"}},
+		"$context-guru": map[string]any{"installed_base_url": ourURL},
+	})
+
+	if _, code := settingsIn(t, state, home, "remove", "--file", path, "--url", ourURL); code != 0 {
+		t.Fatal("remove failed")
+	}
+	// Whatever the record holds, nothing calling itself a pre-edit copy may contain routing.
+	originals, _ := filepath.Glob(filepath.Join(state, "originals", "*.original"))
+	for _, o := range originals {
+		b, err := os.ReadFile(o)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if strings.Contains(string(b), "ANTHROPIC_BASE_URL") {
+			t.Fatalf("%s is called a pre-edit copy and contains routing:\n%s", o, b)
+		}
+	}
+
+	// End to end: the hatch must not put back what the remove took out.
+	out, code := runHatch(t, state, home, proj, "--yes")
+	b, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if strings.Contains(string(b), "ANTHROPIC_BASE_URL") {
+		t.Errorf("the recovery tool re-introduced routing (exit %d):\n%s\nfile:\n%s", code, out, b)
+	}
+	if !strings.Contains(string(b), "Bash(ls:*)") {
+		t.Errorf("the user's own permission grant was lost:\n%s", b)
+	}
+}
+
+// TestEmptyPlanNeverClaimsSuccessWhenNothingCouldBeRestored. An empty plan is reached from two very
+// different states — genuinely clean, and "this file is routed and I have no copy to fix it with" —
+// and both used to end on "already back to their pre-install state". Exit 3 was correct, but nobody
+// reads an exit code; they read the last line, and it said the opposite of the truth to the one user
+// who is locked out.
+func TestEmptyPlanNeverClaimsSuccessWhenNothingCouldBeRestored(t *testing.T) {
+	state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+	path := filepath.Join(proj, "settings.json")
+	writeJSON(t, path, map[string]any{"env": map[string]any{"ANTHROPIC_BASE_URL": ourURL}})
+	// Routed already, no record: the pre-hatch install picking up a hatch on a no-op re-run.
+	if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL); code != 0 {
+		t.Fatal("fixture: add failed")
+	}
+
+	out, code := runHatch(t, state, home, proj, "--yes")
+	if code != 3 {
+		t.Errorf("exit %d, want 3", code)
+	}
+	if strings.Contains(out, "already back to their pre-install state") {
+		t.Errorf("told the user they are unrouted while the file is still routed:\n%s", out)
+	}
+	for _, want := range []string{"NOT back to their pre-install state", "ANTHROPIC_BASE_URL"} {
+		if !strings.Contains(out, want) {
+			t.Errorf("the honest branch is missing %q:\n%s", want, out)
+		}
+	}
+	if b, _ := os.ReadFile(path); !strings.Contains(string(b), ourURL) {
+		t.Error("it modified a file it has no pre-edit copy of")
+	}
+}
+
+// TestARestoreWarnsThatItRevertsTheWholeFile. The primary path is `cp`, so it reverts everything in
+// the file — and Claude Code appends permission grants to settings.local.json as the user approves
+// tools, so a months-old install means months of grants. The confirmation prompt asked "restore
+// these files?" and said nothing about that, which makes it consent to something unstated.
+func TestARestoreWarnsThatItRevertsTheWholeFile(t *testing.T) {
+	state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+	path := filepath.Join(proj, "settings.json")
+	writeJSON(t, path, map[string]any{"permissions": map[string]any{"allow": []string{"Bash(ls:*)"}}})
+	if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL); code != 0 {
+		t.Fatal("add failed")
+	}
+	// The user then approves another tool and sets a model, the way a real month of use looks.
+	cur := readJSON(t, path)
+	cur["model"] = "opus"
+	cur["permissions"] = map[string]any{"allow": []string{"Bash(ls:*)", "Bash(git push:*)"}}
+	writeJSON(t, path, cur)
+
+	out, code := runHatch(t, state, home, proj, "--dry-run")
+	if code != 0 {
+		t.Fatalf("dry run: exit %d\n%s", code, out)
+	}
+	flat := strings.Join(strings.Fields(out), " ")
+	for _, want := range []string{
+		"reverts the WHOLE file",
+		"prereset",             // and that it is undoable
+		"The two files differ", // shown as evidence, before the prompt
+	} {
+		if !strings.Contains(flat, want) {
+			t.Errorf("the plan does not warn about %q before asking:\n%s", want, out)
+		}
+	}
+	// It must not claim a filtered count of "your" changes: settings.py rewrites the file with
+	// indent=2, so a compact original differs on every line, and our own metadata spans lines that
+	// carry none of the words such a filter greps out. The first version of this reported 16 lines
+	// of user changes for a file whose only real change was one permission grant.
+	if strings.Contains(flat, "that are NOT context-guru's") {
+		t.Errorf("re-introduced a filtered line count that cannot be computed without a JSON parser:\n%s", out)
+	}
+}
+
+// TestMissingCopyIsNotRenderedAsAPath: the record carries "-" when no copy was ever taken, and
+// printing it verbatim gave "the pre-edit copy is missing (-)" — which reads as a bug in the tool
+// rather than a known limit of what it holds, on the pre-hatch path that is now the common case.
+func TestMissingCopyIsNotRenderedAsAPath(t *testing.T) {
+	state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+	path := filepath.Join(proj, "settings.json")
+	writeJSON(t, path, map[string]any{"env": map[string]any{"ANTHROPIC_BASE_URL": ourURL}})
+	if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL); code != 0 {
+		t.Fatal("add failed")
+	}
+	out, _ := runHatch(t, state, home, proj, "--dry-run")
+	if strings.Contains(out, "missing (-)") {
+		t.Errorf("rendered the empty-copy marker as a path:\n%s", out)
+	}
+	if !strings.Contains(out, "no pre-edit copy was taken") {
+		t.Errorf("did not explain why there is no copy:\n%s", out)
+	}
+}
+
+// TestDryRunAgreesWithARealRunAboutWhatIsLeft: --dry-run exited 0 unconditionally, so it and a real
+// run disagreed about whether anything was left for a human.
+func TestDryRunAgreesWithARealRunAboutWhatIsLeft(t *testing.T) {
+	state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+	path := filepath.Join(proj, "settings.json")
+	writeJSON(t, path, map[string]any{"env": map[string]any{"ANTHROPIC_BASE_URL": ourURL}})
+	if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL); code != 0 {
+		t.Fatal("add failed")
+	}
+	_, dry := runHatch(t, state, home, proj, "--dry-run")
+	_, real := runHatch(t, state, home, proj, "--yes")
+	if dry != real {
+		t.Errorf("--dry-run exited %d but the real run exited %d; they must agree about whether "+
+			"anything is left for a human", dry, real)
+	}
+}
+
+// TestAnUnusableStateDirectoryNeverFailsTheInstall. "Fail open, always" is a hard boundary in this
+// repo, and the hatch machinery is the newest thing in the write path. An install that would
+// otherwise have worked must not be broken by the recovery bookkeeping — it must report that the
+// hatch is unavailable and carry on.
+func TestAnUnusableStateDirectoryNeverFailsTheInstall(t *testing.T) {
+	home, proj := t.TempDir(), t.TempDir()
+	// A FILE where the state directory should be, so every path inside it is unusable.
+	blocked := filepath.Join(t.TempDir(), "not-a-dir")
+	if err := os.WriteFile(blocked, []byte("x"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	path := filepath.Join(proj, "settings.json")
+	writeJSON(t, path, map[string]any{"theme": "dark"})
+
+	facts, code := settingsIn(t, blocked, home, "add", "--file", path, "--url", ourURL)
+	if code != 0 || facts["result"] != "added" {
+		t.Fatalf("an unusable state dir broke the install: exit %d, %v", code, facts)
+	}
+	if facts["reset_hatch"] != "unavailable" {
+		t.Errorf("reset_hatch=%q; it must say so rather than imply a hatch exists", facts["reset_hatch"])
+	}
+	env, _ := readJSON(t, path)["env"].(map[string]any)
+	if env["ANTHROPIC_BASE_URL"] != ourURL {
+		t.Errorf("the routing itself did not get written: %v", env)
+	}
+	if readJSON(t, path)["theme"] != "dark" {
+		t.Error("the user's settings were damaged")
+	}
+}
--- a/context-guru-plugin/plugin_test.go
+++ b/context-guru-plugin/plugin_test.go
@@ -4210,7 +4210,23 @@ func TestRemoveFirstNeverProducesAnOriginalHoldingRouting(t *testing.T) {
 		t.Fatal("remove failed")
 	}
 	// Whatever the record holds, nothing calling itself a pre-edit copy may contain routing.
+	//
+	// The loop below is VACUOUS on its own for this fixture — no copy is the expected outcome, and an
+	// empty glob asserts nothing — so the expected count is asserted explicitly and the record is
+	// checked to say so. Flagged in review as the pattern worth propagating: an absence assertion
+	// needs a companion that proves the path ran.
 	originals, _ := filepath.Glob(filepath.Join(state, "originals", "*.original"))
+	if len(originals) != 0 {
+		t.Errorf("took %d pre-edit copy/copies of a file that was already routed: %v",
+			len(originals), originals)
+	}
+	rec, err := os.ReadFile(filepath.Join(state, "reset-manifest.tsv"))
+	if err != nil {
+		t.Fatalf("no record was written at all, so the rest of this test proves little: %v", err)
+	}
+	if !strings.Contains(string(rec), path) {
+		t.Errorf("the record does not mention the file, so the hatch would not know about it:\n%s", rec)
+	}
 	for _, o := range originals {
 		b, err := os.ReadFile(o)
 		if err != nil {
@@ -4569,5 +4585,223 @@ func TestTroubleshootingLeadsWithTheCommand(t *testing.T) {
 		if !strings.Contains(head, want) {
 			t.Errorf("the first Troubleshooting entry does not cover %q:\n%s", want, head)
 		}
+	}
+}
+
+// --- fixes from review round 3 of #236 -----------------------------------------------------
+
+// redactShapes is the ledger behind reset.sh's `redact` filter, one row per credential shape.
+//
+// It exists because that filter is a DENYLIST on a recovery tool, and a denylist is a list that will
+// be wrong again — it has now been wrong three times in this PR alone, in three different shapes.
+// Inverting it to an allowlist is worse on a diff of arbitrary JSON (it would redact the
+// `permissions` entries the diff exists to show), so the discipline is this table instead: ADD A ROW
+// WHEN YOU ADD A RULE, and the next miss is a row somebody forgot rather than an invisible leak.
+//
+// The must-survive rows matter as much as the leak rows. A filter that redacts everything passes
+// every absence assertion and makes the diff useless, which is the failure this table also pins.
+var redactShapes = []struct {
+	name string
+	line string
+	// secret must not appear in the output. Empty means the line must pass through UNCHANGED.
+	secret string
+}{
+	{"ANTHROPIC_API_KEY", `  "ANTHROPIC_API_KEY": "sk-ant-REALKEY0000",`, "REALKEY0000"},
+	{"ANTHROPIC_AUTH_TOKEN", `  "ANTHROPIC_AUTH_TOKEN": "REALTOKENVALUE",`, "REALTOKENVALUE"},
+	// The one that matters most here, and not a shape invented for the test: this is how a Context
+	// Guru credential is carried on this project's own dev machines, so it is the single most likely
+	// credential to appear in a context-guru user's env block — and its name contains none of
+	// key/token/secret/password/credential, which is why the first two versions of the filter missed
+	// it. HEADER is in the name class because of this row.
+	{"ANTHROPIC_CUSTOM_HEADERS", `  "ANTHROPIC_CUSTOM_HEADERS": "x-context-guru-token: cg_live_REALGURU",`, "REALGURU"},
+	{"authorization bearer JWT", `  "authorization": "Bearer eyJhbGciOiJIUzI1NiJ9.REALJWTBODY",`, "REALJWTBODY"},
+	{"github PAT", `  "GITHUB_PAT": "ghp_REALPAT00000000",`, "REALPAT00000000"},
+	{"credential in a query param", `  "ANTHROPIC_UPSTREAM": "https://gw/anthropic?api_key=REALQUERYKEY",`, "REALQUERYKEY"},
+	{"credential in URL userinfo", `  "MY_GW": "https://svc:REALURLPASS@gw.corp/anthropic",`, "REALURLPASS"},
+	{"slack token", `  "SLACK": "xoxb-REALSLACKTOKEN",`, "REALSLACKTOKEN"},
+	{"aws access key id", `  "AWS_ACCESS_KEY_ID": "AKIAREALAWSKEY0000",`, "REALAWSKEY"},
+	{"bare exported base URL with userinfo", `https://svc:REALURLPASS@gw.corp/anthropic`, "REALURLPASS"},
+
+	{"permission grant must survive", `      "Bash(git push:*)",`, ""},
+	{"model must survive", `  "model": "opus",`, ""},
+	{"theme must survive", `  "theme": "dark",`, ""},
+	{"a plain loopback base URL must survive", `  "ANTHROPIC_BASE_URL": "http://127.0.0.1:8787/anthropic",`, ""},
+}
+
+// TestRedactCoversEveryKnownCredentialShape drives the real filter, lifted out of reset.sh, rather
+// than a copy of it — a copy would drift from the thing that ships.
+func TestRedactCoversEveryKnownCredentialShape(t *testing.T) {
+	requireTool(t, "bash")
+	body, err := os.ReadFile(filepath.Join(scriptsDir(t), "reset.sh"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Slice out REDACT_NAME through the end of the redact() function.
+	_, rest, ok := strings.Cut(string(body), "REDACT_NAME=")
+	if !ok {
+		t.Fatal("reset.sh no longer defines REDACT_NAME; this test is extracting the wrong thing")
+	}
+	fnEnd := strings.Index(rest, "\n}\n")
+	if fnEnd < 0 {
+		t.Fatal("could not find the end of redact(); extraction would be silently partial")
+	}
+	fn := "REDACT_NAME=" + rest[:fnEnd+3]
+	if !strings.Contains(fn, "redact()") || !strings.Contains(fn, "sed") {
+		t.Fatalf("extracted fragment does not look like the filter:\n%s", fn)
+	}
+	fnPath := filepath.Join(t.TempDir(), "redact.sh")
+	if err := os.WriteFile(fnPath, []byte(fn), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	for _, tc := range redactShapes {
+		cmd := exec.Command("sh", "-c",
+			`. "$1"; printf '%s\n' "$2" | redact`, "_", fnPath, tc.line)
+		b, err := cmd.CombinedOutput()
+		if err != nil {
+			t.Fatalf("%s: running redact: %v (%s)", tc.name, err, b)
+		}
+		got := strings.TrimRight(string(b), "\n")
+		if tc.secret == "" {
+			if got != tc.line {
+				t.Errorf("%s: over-redacted, so the diff loses what it exists to show\n got:  %q\n want: %q",
+					tc.name, got, tc.line)
+			}
+			continue
+		}
+		if strings.Contains(got, tc.secret) {
+			t.Errorf("%s: LEAKED %q\n got: %q", tc.name, tc.secret, got)
+		}
+		if got == tc.line {
+			t.Errorf("%s: line passed through untouched, so no rule matched it at all\n got: %q",
+				tc.name, got)
+		}
+	}
+}
+
+// TestNoPrinterOfContentEscapesTheFilter walks every site in reset.sh that prints real content and
+// plants a credential in each. The round-2 fix covered the plan diff; the review then found a FOURTH
+// site inside report_environment itself — the function whose own header promises values are never
+// printed — because that test only exercised the diff. So this one is organised by SITE.
+func TestNoPrinterOfContentEscapesTheFilter(t *testing.T) {
+	const secret = "REALSECRET-do-not-print-me"
+
+	t.Run("the verify pass", func(t *testing.T) {
+		// A user's own gateway carrying credentials in the URL, taken over with --force and then
+		// handed back. Verify greps the restored file and prints the matching lines.
+		state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+		path := filepath.Join(proj, "settings.json")
+		theirs := "https://svc:" + secret + "@gw.corp.example/anthropic"
+		writeJSON(t, path, map[string]any{"env": map[string]any{"ANTHROPIC_BASE_URL": theirs}})
+		if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL,
+			"--force"); code != 0 {
+			t.Fatal("add --force failed")
+		}
+		out, _ := runHatch(t, state, home, proj, "--yes")
+		if strings.Contains(out, secret) {
+			t.Errorf("the verify pass printed a credential:\n%s", out)
+		}
+		if !strings.Contains(out, "still mentions") {
+			t.Fatalf("the verify branch never ran, so this asserted nothing:\n%s", out)
+		}
+	})
+
+	t.Run("the environment report", func(t *testing.T) {
+		state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+		path := filepath.Join(proj, "settings.json")
+		writeJSON(t, path, map[string]any{"theme": "dark"})
+		if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL); code != 0 {
+			t.Fatal("add failed")
+		}
+		cmd := exec.Command(hatchPath(t, state), "--dry-run")
+		cmd.Dir = proj
+		cmd.Env = []string{
+			"CONTEXT_GURU_STATE=" + state, "HOME=" + home, "PATH=" + os.Getenv("PATH"),
+			"ANTHROPIC_BASE_URL=https://svc:" + secret + "@gw.corp.example/anthropic",
+		}
+		b, _ := cmd.CombinedOutput()
+		out := string(b)
+		if strings.Contains(out, secret) {
+			t.Errorf("report_environment printed a credential:\n%s", out)
+		}
+		if !strings.Contains(out, "exported in this shell") {
+			t.Fatalf("the branch that prints $base never ran:\n%s", out)
+		}
+	})
+
+	t.Run("the no-record grep", func(t *testing.T) {
+		state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+		if err := os.MkdirAll(filepath.Join(proj, ".claude"), 0o755); err != nil {
+			t.Fatal(err)
+		}
+		path := filepath.Join(proj, ".claude", "settings.local.json")
+		writeJSON(t, path, map[string]any{"theme": "dark"})
+		if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL,
+			"--upstream", "https://gw.corp/anthropic?api_key="+secret); code != 0 {
+			t.Fatal("add failed")
+		}
+		hatch := hatchPath(t, state)
+		if err := os.Remove(filepath.Join(state, "reset-manifest.tsv")); err != nil {
+			t.Fatal(err)
+		}
+		cmd := exec.Command(hatch, "--yes")
+		cmd.Dir = proj
+		cmd.Env = []string{"CONTEXT_GURU_STATE=" + state, "HOME=" + home, "PATH=" + os.Getenv("PATH")}
+		b, _ := cmd.CombinedOutput()
+		out := string(b)
+		if strings.Contains(out, secret) {
+			t.Errorf("the no-record grep printed a credential:\n%s", out)
+		}
+		if !strings.Contains(out, "ANTHROPIC_UPSTREAM") {
+			t.Fatalf("the grep never printed the key line, so this asserted nothing:\n%s", out)
+		}
+	})
+}
+
+// TestASuccessfulRestoreSaysSoEvenWhenTheShellIsRouted. The final summary tested INCOMPLETE for a
+// question about files — the same bug fixed in the empty-plan branch, one branch further down. So a
+// COMPLETELY successful restore, verified unrouted, run from a shell with an exported base URL (a
+// hosted agent, or simply the shell they installed from) printed "Finished with something left for
+// you" and never printed the count: $RESTORED was thrown away on the one run where it is the good
+// news, and the user who had just recovered was told the run did not finish.
+//
+// It survived three rounds because NO test grepped for "Done." at all.
+func TestASuccessfulRestoreSaysSoEvenWhenTheShellIsRouted(t *testing.T) {
+	state, home, proj := t.TempDir(), t.TempDir(), t.TempDir()
+	path := filepath.Join(proj, "settings.json")
+	writeJSON(t, path, map[string]any{"permissions": map[string]any{"allow": []string{"Bash(ls:*)"}}})
+	if _, code := settingsIn(t, state, home, "add", "--file", path, "--url", ourURL); code != 0 {
+		t.Fatal("add failed")
+	}
+
+	// Non-empty plan (a real restore) AND a routed shell: the two conditions together.
+	cmd := exec.Command(hatchPath(t, state), "--yes")
+	cmd.Dir = proj
+	cmd.Env = []string{
+		"CONTEXT_GURU_STATE=" + state, "HOME=" + home, "PATH=" + os.Getenv("PATH"),
+		"ANTHROPIC_BASE_URL=" + ourURL,
+	}
+	b, err := cmd.CombinedOutput()
+	code := 0
+	if ee, ok := err.(*exec.ExitError); ok {
+		code = ee.ExitCode()
+	} else if err != nil {
+		t.Fatal(err)
+	}
+	out := string(b)
+	if !strings.Contains(out, "restored:") {
+		t.Fatalf("no restore happened, so this test asserted nothing:\n%s", out)
+	}
+	if !strings.Contains(out, "Done. 1 file(s) put back.") {
+		t.Errorf("a fully successful restore never reported its count:\n%s", out)
+	}
+	if strings.Contains(out, "Finished with something left for you") {
+		t.Errorf("told the user the run did not finish after a verified-clean restore:\n%s", out)
+	}
+	if code != 3 {
+		t.Errorf("exit %d, want 3: the exported base URL is still left for a human", code)
+	}
+	if !strings.Contains(out, "environment report above") {
+		t.Errorf("did not point at what is actually left:\n%s", out)
 	}
 }
--- a/context-guru-plugin/plugin_test.go
+++ b/context-guru-plugin/plugin_test.go
@@ -4608,11 +4608,18 @@ var redactShapes = []struct {
 }{
 	{"ANTHROPIC_API_KEY", `  "ANTHROPIC_API_KEY": "sk-ant-REALKEY0000",`, "REALKEY0000"},
 	{"ANTHROPIC_AUTH_TOKEN", `  "ANTHROPIC_AUTH_TOKEN": "REALTOKENVALUE",`, "REALTOKENVALUE"},
-	// The one that matters most here, and not a shape invented for the test: this is how a Context
-	// Guru credential is carried on this project's own dev machines, so it is the single most likely
-	// credential to appear in a context-guru user's env block — and its name contains none of
-	// key/token/secret/password/credential, which is why the first two versions of the filter missed
-	// it. HEADER is in the name class because of this row.
+	// The one that matters most here, and not a shape invented for the test.
+	//
+	// PROVENANCE, recorded because it is the strongest argument for this whole table: this leak was
+	// LIVE, not theoretical. `ANTHROPIC_CUSTOM_HEADERS` carrying a `<header>: <token>` string is how a
+	// Context Guru credential is set on this project's own development machines — set for interactive
+	// sessions and explicitly unset for benchmark runs — and it was present in the real
+	// ~/.claude/settings.json of the machine this filter was written on, where the author had read
+	// that file earlier the same day and not connected it to the filter.
+	//
+	// So the single most likely credential to appear in a context-guru user's env block was the one
+	// the first two versions of the filter did not catch, and its name contains none of
+	// key/token/secret/password/credential. HEADER is in the name class because of this row.
 	{"ANTHROPIC_CUSTOM_HEADERS", `  "ANTHROPIC_CUSTOM_HEADERS": "x-context-guru-token: cg_live_REALGURU",`, "REALGURU"},
 	{"authorization bearer JWT", `  "authorization": "Bearer eyJhbGciOiJIUzI1NiJ9.REALJWTBODY",`, "REALJWTBODY"},
 	{"github PAT", `  "GITHUB_PAT": "ghp_REALPAT00000000",`, "REALPAT00000000"},
@@ -4626,6 +4633,13 @@ var redactShapes = []struct {
 	{"model must survive", `  "model": "opus",`, ""},
 	{"theme must survive", `  "theme": "dark",`, ""},
 	{"a plain loopback base URL must survive", `  "ANTHROPIC_BASE_URL": "http://127.0.0.1:8787/anthropic",`, ""},
+	// These three are what the verify pass and the no-record grep exist to SHOW the user — they are
+	// the routing itself, not a credential. AUTH and HEADER in the name class match more broadly than
+	// the original five did, so each is pinned here: over-redacting these would leave a locked-out
+	// user reading "<value not shown>" where they need to see which port they are pointed at.
+	{"a corporate base URL must survive", `  "ANTHROPIC_BASE_URL": "https://gateway.corp.example/anthropic",`, ""},
+	{"ANTHROPIC_UPSTREAM must survive", `  "ANTHROPIC_UPSTREAM": "https://gateway.corp.example",`, ""},
+	{"CONTEXT_GURU_BIN must survive", `  "CONTEXT_GURU_BIN": "/home/user/.local/bin/context-guru-proxy",`, ""},
 }
 
 // TestRedactCoversEveryKnownCredentialShape drives the real filter, lifted out of reset.sh, rather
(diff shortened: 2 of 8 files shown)


amiddavid:
Handing this to review. A few pointers on where I think the risk actually is, rather than a re-summary of the description:

**Weakest points, in my own estimation**

1. **`record_touch()` is called from inside `save()`** (`settings.py`). That was deliberate — `add` has six exits that write, and one of them forgetting to record would produce a routed machine with no recorded way back — but it means a function whose job is "write the user's settings atomically" now also writes two files elsewhere. Everything in it is wrapped best-effort so it cannot fail an install, and that is worth checking rather than trusting.
2. **`existed_before` is decided once, on first touch, and never revised.** A wrong value here is the one way the hatch could *delete* a file the user wrote themselves. The first record wins by design (a later run must not claim "it existed" about a file we created), so the question is whether any path can reach `save()` with `os.path.exists(real)` already false for a file that did pre-exist.
3. **`ensure_hatch()` on `result=unchanged`** is the one place that records a path *without* having edited it in that call. It is gated on `rc == 0` so a refused `result=conflict` (somebody else's gateway) is never recorded — `TestAConflictIsNeverRecordedAsOurEdit` covers exactly that, and it is the assertion I would try hardest to break.
4. **`_render_tsv()` refuses a path containing a tab or newline** and records it as a comment instead, which means the hatch silently will not restore that one. I chose a named gap over a line that could be misparsed into the wrong filename, but it is a real gap.

**Known gaps, already stated in the description but worth repeating here**

- No model calls in any of the testing, so the install skill's *prose* — the model's judgment about scope and chaining — is unexercised. The Go tests and the sandbox both drive the scripts directly.
- The proxy is stubbed (no Go toolchain on the machine this was written on), so nothing here proves a real request survives a real proxy. Everything under test keys off `/healthz`.

CI is green on all checks, including `build-test` with `-race` on ubuntu-latest.

amiddavid:
## Review round 1 of 3

Reviewed the whole diff and drove `settings.py` + `reset.sh` directly on this laptop (python3 + `sh`, no Go needed) rather than reading only. Four findings; **two are severe and both land on the population this PR is explicitly for** — machines that installed before the hatch existed. Repros below are copy-pasteable.

Answering your own four first, since three of them hold up:

1. **`record_touch()` inside `save()`** — the placement is right, and it is *nearly* best-effort. Two gaps: the `except OSError` clauses do not cover everything the body can raise (`_render_tsv`/`json.dumps` do `.encode("utf-8")` on a path, which raises `UnicodeEncodeError` — not an `OSError` — for a filename carrying surrogates from `surrogateescape` on Linux). Given `CLAUDE.md`'s "fail open, always", the hatch machinery should be `except Exception` at the top of both `record_touch` and `ensure_hatch`, not `except OSError` at three inner sites. Cheap, and it makes the invariant true rather than nearly true. The second gap is finding 1 below.
2. **`existed_before` decided once** — I could not find a path that reaches `save()` with `os.path.exists(real)` false for a file that did pre-exist. `load()` and `save()` compute existence independently but both follow symlinks, and every writing branch runs `load()` first. The one asymmetry is a *dangling* symlink at `--file`: `real` is the (absent) target, so `existed_before=0` and the hatch deletes the target — which is correct — but it leaves the dangling link in `.claude` and the manifest never mentions the link. Not a defect, just noting I looked.
3. **`ensure_hatch()` on `result=unchanged`** — the `rc == 0` gate holds. I tried to reach `result=unchanged` on a file context-guru never wrote: the routing branch needs `env[KEY] == args.url`, the statusline-only branch needs `statusLine` to equal our own recorded value. Neither is reachable for a stranger's file. `conflict` exits 2 and is excluded. I agree this is the assertion worth defending and I could not break it.
4. **`_render_tsv()` tab/newline refusal** — agreed, a named gap beats a misparsed filename. Not raising it.

---

### 1. severe — the hatch can *re-introduce* routing, on exactly the pre-hatch installs this PR targets

`record_touch()` is called from `save()`, and `save()` is also `cmd_remove`'s write path. So when the **first** recorded touch of a file is a *removal*, the `O_EXCL` "copy taken before the first edit" is a copy of the **routed** file — and because it is `O_EXCL` and never overwritten, that wrong copy is now permanent.

Who hits it: anyone who installed with a pre-hatch plugin (i.e. every current install), upgrades, then runs `/context-guru:uninstall`. Also anyone whose state dir was wiped between install and remove. Same thing reaches it through `ensure_hatch()`: a pre-hatch user re-runs the install (`result=unchanged`, `original: ""`), and the next `save()` backfills `e["original"]` with a copy of the routed file.

Reproduced — the recovery tool wrote `ANTHROPIC_BASE_URL` back into an unrouted project:

```sh
T=$(mktemp -d); cd "$T"; export HOME="$T/home" CONTEXT_GURU_STATE="$T/state"; mkdir -p "$HOME" proj/.claude
cat > proj/.claude/settings.local.json <<'JSON'
{"env": {"ANTHROPIC_BASE_URL": "http://127.0.0.1:8787/anthropic"},
 "permissions": {"allow": ["Bash(ls:*)"]},
 "$context-guru": {"installed_base_url": "http://127.0.0.1:8787/anthropic"}}
JSON
cd proj
python3 .../scripts/settings.py remove --file .claude/settings.local.json --url http://127.0.0.1:8787/anthropic
cat "$CONTEXT_GURU_STATE"/originals/*.original   # <-- the "pre-edit" copy is the ROUTED file
sh .../scripts/reset.sh --yes
cat .claude/settings.local.json                  # <-- routed again
```

Output: `restored: .../settings.local.json`, then the file contains the base URL it had removed.

To be fair to the design: the verify pass does flag it (`? ... points at a local proxy`) and the run exits 3, and the `.context-guru-prereset-*` copy makes it recoverable. But the user was told `restored:` on a line above, the tool's whole promise is "run this and you are unrouted", and a recovery tool re-applying the thing it exists to remove is the one behaviour it cannot have. It is also worst in the state where the user is already panicking.

Suggested fix, smallest first: take the original copy **only when the current file does not already carry our keys** (`KEY`/`UPSTREAM_KEY`/`BIN_KEY` in `env`, or a `META` record present). If it does, record the path with `original: ""` and let the honest missing-copy branch handle it — which is what `ensure_hatch` already does for this exact situation. Alternatively thread intent into `save()` so `remove` records without claiming an original. Either way a test asserting *"remove-first never produces an original that contains `ANTHROPIC_BASE_URL`"* is the one I would want pinned.

### 2. severe — "already back to their pre-install state" is printed about a file that is still routed

`reset.sh:264`. The missing-pre-edit-copy branch (lines 245-257) deliberately adds nothing to `$PLAN`, so an empty plan is reached in two very different states: genuinely clean, and *"I know this file is routed and I have no copy to fix it with"*. Both print:

```
Nothing to restore — your settings files are already back to their pre-install state.
```

Reproduced: routed project, `originals/` copy absent, `reset.sh --dry-run` prints the correct `! the pre-edit copy is missing` block **and then** that sentence. Exit status is 3, which is right, but nobody reads an exit code — they read the last line, and it says the opposite of the truth to the one user who is locked out. This is the `ensure_hatch` population by construction: those entries always have no original.

Fix: gate that sentence on `INCOMPLETE = 0`. When `INCOMPLETE` is set, say what is actually true — "nothing could be restored automatically; see the `!` lines" — and repeat the manual `cp` line.

### 3. medium — a whole-file restore silently discards every unrelated later change to the settings file

The primary path is `cp "$original" "$path"`, restoring the file as it was *before the install*. Everything the user changed in that file since is gone. For `.claude/settings.local.json` that is not hypothetical: Claude Code appends permission grants to it as the user approves tools, so a months-old install means months of accumulated grants.

Reproduced — install, then add `model`, `theme` and two permission rules, then run the hatch:

```
before: {"permissions": {"allow": ["Bash(ls:*)","Bash(git push:*)","Read(...)"]}, "model": "opus", "theme": "dark"}
after:  {"permissions": {"allow": ["Bash(ls:*)"]}}
```

The plan output before the prompt says only `RESTORE from the copy taken before the first edit:` — nothing warns that this reverts unrelated settings, and the user has no way to know anything was lost. The `.context-guru-prereset-*` copy means it is recoverable, which is why this is medium and not severe.

Two options, and I do not think this needs to block:
- **Minimum:** one line in the plan, before the prompt — "this restores the whole file as it was before the install; anything you changed in it since (permission grants, model, theme) goes back too. Your current version is copied to `*.context-guru-prereset-*` first." That turns silent loss into an informed choice, which is all the confirmation prompt is for.
- **Better, if you want it:** make key-scoped removal the primary and whole-file restore the fallback. `sed`-ing JSON is exactly the cleverness this script is right to refuse — but `cmp`-ing against the original to confirm the *only* difference is our keys, and otherwise falling back to the whole-file copy, would get most of it without a parser.

### 4. minor

- `_render_tsv()` writes `e.get("original") or "-"`, and `reset.sh:247` prints it as a path: `! the pre-edit copy is missing (-)`. Every `ensure_hatch` entry renders that way — i.e. on the "already-routed installs get a hatch too" path the description leads with. Special-case `-` to something like `no pre-edit copy was taken (this project was already routed when the record was created)`.
- `--dry-run` exits 0 even when `report_environment` set `INCOMPLETE=1` (`reset.sh:272-277`), so `--dry-run` and a real run disagree about whether anything is left for a human. Either honour `INCOMPLETE` there or document that `--dry-run` only ever reports plan validity.

---

Nothing to say against the rest: the `O_EXCL`-never-pruned reasoning, installing the hatch outside the plugin, the scope gate living in the script instead of in prose, `</dev/tty` for the prompt, verifying instead of trusting `cp`, and the values-never-printed environment report are all right, and the scope gate behaves as documented (`reason=user_scope_needs_flag`, exit 2 — checked). Both known gaps you listed are fine to ship with, stated as they are.

Findings 1 and 2 are the ones I would want fixed before merge; 3 and 4 are yours to weigh. Two review rounds left.

amiddavid:
Round 1 addressed in 238634e. All four fixed; finding 1 was as severe as you called it, and I reproduced it with your snippet before touching anything.

**1 — the hatch re-introducing routing.** Confirmed, and thank you for finding it: `save()` being `cmd_remove`'s write path is exactly the connection I missed when I argued for the single choke point. Your first suggestion is what shipped — no original is taken when the file already carries our keys — but decided from **content, not intent**, because threading "this is an install" through `save()` still takes a routed copy when an *add* runs against a file we had already routed (wiped state dir, `--force` re-run). The predicate is: our metadata key present, or `ANTHROPIC_UPSTREAM`/`CONTEXT_GURU_BIN` present, or the base URL is loopback.

That last clause is the ambiguous one, so to be explicit about the trade: a user's own loopback gateway (litellm on 4000) installing us would now be recorded with no original, costing them a whole-file restore they could have had. I took that deliberately — a false positive is declined honestly by the missing-copy branch, a false negative re-applies the routing they ran the hatch to escape. A base URL that is *not* ours is not a signal at all, since a corporate gateway is the value most worth having a copy of.

`TestRemoveFirstNeverProducesAnOriginalHoldingRouting` pins it under the name you asked for, and asserts both halves: nothing calling itself a pre-edit copy contains `ANTHROPIC_BASE_URL`, and the hatch run end-to-end leaves the file unrouted with the user's permission grant intact.

**2 — "already back to their pre-install state".** Gated on `INCOMPLETE`, as you suggested. The honest branch now names the two manual routes (copy the newest timestamped backup, or delete the three keys from the `env` block) rather than only pointing upward at the `!` lines.

**3 — whole-file restore.** Took the minimum, and my first attempt at it was wrong in a way worth recording: I counted "lines that are NOT context-guru's" by grepping our key names out of a `diff`, and that number is false precision twice over — `settings.py` rewrites the file with `indent=2` so a compact original differs on every line, and our own metadata spans lines carrying none of the filtered words. It reported **16 lines of "your" changes for a file whose only real change was one added permission grant.** Since `sh` cannot compare JSON semantically and must not try, the warning is now unconditional on a restore and the diff is shown as *evidence*, labelled as including our keys and our reformatting, with no claim about whose side is whose. There's a test asserting the filtered count does not come back.

I did not take the key-scoped-removal option. `cmp`-ing to confirm the only difference is our keys can't be done without a JSON parser for the same reformatting reason, and I'd rather ship a warning that is true than a comparison that is subtly wrong.

**4 — both.** `-` now reads "no pre-edit copy was taken for this file — it already carried context-guru's keys when the record was created"; `--dry-run` honours `INCOMPLETE`, with a test asserting the two exit codes agree.

**Your note on my #1 (the `except OSError` gap).** You're right, and it's a `CLAUDE.md` hard-boundary violation rather than a nicety. Both entry points are now thin fail-open wrappers with `except Exception`, and `TestAnUnusableStateDirectoryNeverFailsTheInstall` asserts an install still succeeds and reports `reset_hatch=unavailable` when the state directory is unusable.

On your #2 and #3: agreed on both, nothing changed. The dangling-symlink asymmetry you noted is real and I'm leaving it — the hatch deletes the target, which is correct, and the leftover link is inert.

**State:** 86 tests, `go vet` and `gofmt` clean, sandbox run against the real Claude Code install still 20/20.

One thing for the humans, unrelated to the review: pushing this commit reported `Bypassed rule violations for refs/heads/feat/escape-hatch-and-scope-guard: Cannot update this protected ref`. The push landed (remote is at 238634e), but a branch-protection rule appears to match feature branches on this repo and my push bypassed it rather than being blocked.

amiddavid:
## Review round 2 of 3

Re-ran every round-1 repro against 238634e, then attacked the new code. **Findings 1, 2 and 4 are genuinely fixed** — I could not re-break any of them. Finding 3's fix, however, introduced a severe defect of its own, and the fix for 2 is not quite complete.

Verified fixed:
- **1** — my repro now records `original: ""` and the hatch declines honestly instead of re-routing. A fresh install still takes a real, unrouted original (checked the copy's content, not just its existence). The `_looks_routed_by_us` predicate does what it says.
- **2** — the reassuring sentence is now gated, and the two named manual routes are the right ones.
- **4** — both. The `-` message reads as a known limit rather than a tool bug.
- **the `except OSError` gap** — both entry points are now genuinely fail-open.

Three new findings, and then the two judgement calls you asked me to attack.

---

### 1. severe — the new plan diff prints the user's API key to the terminal

`reset.sh:265-274`. The whole-file warning is right and the diff-as-evidence choice is right, but the diff is unfiltered file content, and `env` is exactly where people keep credentials. Reproduced with a real-shaped settings file:

```sh
cat > .claude/settings.local.json <<'JSON'
{"env": {"ANTHROPIC_API_KEY": "sk-ant-SUPERSECRET-do-not-print-me"},
 "permissions": {"allow": ["Bash(ls:*)"]}}
JSON
python3 .../settings.py add --file .claude/settings.local.json --url http://127.0.0.1:8787/anthropic
# ... user adds a permission grant ...
sh .../reset.sh --dry-run
```

Output, twice, once per side of the diff:

```
          <   "env": {"ANTHROPIC_API_KEY": "sk-ant-SUPERSECRET-do-not-print-me"},
          >     "ANTHROPIC_API_KEY": "sk-ant-SUPERSECRET-do-not-print-me",
```

This contradicts the script's own stated principle 170 lines above it — *"a recovery tool that echoes a live API key into a terminal buffer is not a recovery tool"* — and it fires on `--dry-run`, the invocation the docs recommend running first. It matters more than the usual "secret in a log" because of who runs this: somebody debugging a 401 who is very likely to paste the output into a chat, an issue, or a screenshot, precisely because the output is designed to be read and acted on.

Fix is the treatment you already use for the rc-file grep — redact the value half rather than dropping the line, so the diff still shows *that* the line changed:

```sh
diff "$original" "$path" 2>/dev/null | grep '^[<>]' | head -14 \
  | sed -E 's/("[A-Za-z_]*(KEY|TOKEN|SECRET|PASSWORD|CREDENTIAL)[A-Za-z_]*"[[:space:]]*:[[:space:]]*")[^"]*/\1<value not shown>/' \
  | sed 's/^/          /'
```

Case-insensitivity is worth having too (`apiKey`, `authToken`). A test planting a secret and grepping the plan output for it would pin this the same way `TestTheEnvironmentReportNeverPrintsAValue` pins the other half — and note that existing test does not cover this path, which is how it got through.

### 2. medium — `reset_original=unavailable` is now reported when a perfectly good original exists

`_record_touch`, the new `if existed and _looks_routed_by_us(real)` branch, sets `reset_original`/`reset_original_reason` unconditionally — without checking whether the manifest entry already *has* an original. So the normal life-cycle of a normally-installed project now reports the alarming thing on uninstall:

```
$ settings.py add    --file .claude/settings.local.json --url http://127.0.0.1:8787/anthropic
result=added                                          # original taken correctly, content verified clean
$ settings.py remove --file .claude/settings.local.json --url http://127.0.0.1:8787/anthropic
result=removed
reset_original=unavailable
reset_original_reason=file already carried context-guru's keys
$ # ...but the manifest entry's original is right there, and valid:
.../state/originals/claude-settings.local.json.084f7398a047.original
```

The predicate is behaving correctly — a copy of the file *at that moment* would indeed be meaningless. The reporting is what's wrong: `reset_original=unavailable` is a statement about what the hatch holds, and the hatch holds a good copy. `install/SKILL.md` tells the model this fact means "the hatch can then unroute but not restore", so the skill will now tell users their original content is unrecoverable on the ordinary uninstall path, when it is sitting on disk. That is the same class of false-reassurance-inverted as round-1 finding 2, just pointing the other way.

Fix: set the fact only when the entry that ends up in the manifest has no `original` — i.e. decide it after the entry lookup below, not in the branch. Roughly, drop the two `HATCH_FACTS` lines from the predicate branch and let the existing `if existed and not original:` tail handle it, extended to consult the matched entry's stored `original`.

### 3. minor — the exit codes still disagree, one branch further along

Round 1's finding 4b is fixed for the `--dry-run` path, but the empty-plan branch tests `INCOMPLETE` *before* `report_environment` runs, so the one condition no file restore can fix still exits 0:

```sh
export ANTHROPIC_BASE_URL=http://127.0.0.1:8787/anthropic   # exported in the shell
sh reset.sh --yes
# Nothing to restore — your settings files are already back to their pre-install state.
#   ! ANTHROPIC_BASE_URL is exported in THIS SHELL and points at a local proxy:
#     ... this shell stays routed until you unset it.
# exit=0
```

Exit 0 while printing a `!` that says the user is still routed, and the identical condition on a non-empty plan exits 3. Fix: call `report_environment` before the `INCOMPLETE` test in that branch (or re-test after it). The sentence itself is defensible — the *files* are back — but the status code and the `!` should not contradict each other.

### 4. minor — docs drift from 238634e

`install/SKILL.md` still describes `reset_original=unavailable` as *"Rare (an unreadable settings file)"*. After this commit it is the routine already-routed case, and the new `reset_original_reason=` fact is undocumented, so the model has the explanation available and no instruction to use it. One line each. (Moot for the common path if you take finding 2.)

---

### The two calls you asked me to attack

**The loopback clause — accept, and it costs you less than you think.** Your reasoning on the asymmetry is right, but I think you have over-priced the false positive: the clause is only reached when `META` is absent *and* neither `ANTHROPIC_UPSTREAM` nor `CONTEXT_GURU_BIN` is present. `meta["installed_base_url"]` is written by every `add` branch in `main` — it predates this PR — so **every context-guru install that has ever existed already trips the `META` check**, including the whole pre-hatch population the clause was added for. What's left for the loopback clause is a hand-edited settings file whose base URL happens to be loopback. A user's litellm-on-4000 with no context-guru metadata is in that residue, but so is nothing else you need. Keep it as belt-and-braces; the trade is cheaper than your comment claims.

One nit if you keep it: `startswith()` misses `http://127.0.0.1/anthropic` (no port), `http://0.0.0.0:`, `http://[::1]/` without a port, and `https://localhost:`. Each is a false *negative*, which is the direction that re-introduces finding 1. Low likelihood given the above, cheap to widen.

**The unconditional warning + unfiltered diff — yes, honest enough**, and dropping the filtered line count was the right call for exactly the reason you give. Showing the difference and labelling what it includes is the honest move where a number would have been a claim you can't support in `sh`. It just has to stop printing secrets (finding 1). Not taking key-scoped removal is also the right call; I raised it as an option, not a requirement, and "a warning that is true over a comparison that is subtly wrong" is the correct ordering.

---

### The two things the user asked me to check specifically

**Project scope must be the default, so a broken project leaves other projects usable.** Verified end-to-end, and this holds:

- `settings.py add` is the **only** thing in the plugin that writes the routing key into a settings file — `start-proxy.sh`, `check-proxy.sh` and `statusline.py` only *read* `$ANTHROPIC_BASE_URL`. So one gate is a complete gate.
- Four bypass attempts, all refused with `reason=user_scope_needs_flag`, exit 2: the absolute path, a relative `.claude/settings.json` from `$HOME`, a `..`-laundered path, and a symlink pointing at it. `realpath` on both sides is doing its job.
- With `--user-scope` it writes; `remove` is ungated in every case. Both correct.
- The user's rationale is now structurally guaranteed rather than advisory: a mistake can only land in a project file, so a session in any other project still works and can run the uninstall skill.

One structural note, not a defect: step 6's command is `--file <target>`, and *which* project file (`.claude/settings.local.json` vs `.claude/settings.json`) is still chosen by the step-3 prose. That is fine — both are project scope, so the property the user cares about holds either way — but it is worth knowing that what the script enforces is "not machine-wide", not "local, not team".

**The reset must work outside the plugin, with clear docs and one single bash command.** The mechanism is right — plain `sh`, installed to `~/.local/state/context-guru/` plus `~/.local/bin`, no plugin/Python/network/Claude dependency, and a single command with no flags is the documented primary form. The gap is **discoverability at the moment of failure**:

- `docs/how-to/install-plugin.md`'s **Troubleshooting** section — the section a stuck user actually opens — never names the hatch. The hatch has its own section *above* it, and nothing in Troubleshooting cross-references it.
- No entry anywhere for the symptom that motivated the whole PR. There is no "every request fails with 401" or "every request hangs" entry whose answer is the one command.
- No mention in the top-level `README.md`.

Please add a Troubleshooting entry that leads with the literal command on its own line, e.g. **"Every request fails or hangs, and `/context-guru:uninstall` can't run."** → `~/.local/state/context-guru/context-guru-reset`. A user who has to find this will be searching the docs with a Claude that cannot answer questions, and they should not have to read past a heading to get the command.

---

Finding 1 is the only one I would block on; 2 is worth fixing because it makes the install skill say something false on the ordinary path. 3 and 4 are cheap. One review round left.

amiddavid:
Round 2 addressed in d752e4d. Finding 1 was mine and you're right that it was the blocker — I reproduced it before touching anything, two leaked lines on `--dry-run`.

**1 — the diff printing credentials.** Fixed, and I took your redaction approach (keep the line, redact the value) but applied it at **every site that prints file content**, not only the one you found: the plan diff, the verify pass, and the no-record grep all echo lines from the user's settings. "Remember to filter this one too" is how the first leak happened, so there is now one `redact` filter and three call sites through it.

Three shapes rather than one: key names containing key/token/secret/password/credential in any case, an `sk-…` value whatever the field is called (an oddly-named field holding a real key was in my fixture), and credentials embedded in a URL (`https://alice:pw@host`). Case-insensitivity is spelled with bracket classes because BSD sed — macOS, where this runs most — has no `s///I`.

`TestThePlanNeverPrintsACredential` covers both `--dry-run` and a real run, and also asserts the diff still shows the change it exists to show. Worth recording that **the test caught its own fixture being wrong**: I first built it with `writeJSON`, which indents — and since `settings.py` rewrites with the same indent, the credential lines came out byte-identical on both sides and never entered the diff at all. The secret-absence assertion was passing for the wrong reason, and only the "no redaction marker, so the diff may simply not have run" check exposed it. That check was defensive padding when I wrote it; it turned out to be the assertion that mattered.

**2 — `reset_original=unavailable` on a normal uninstall.** Fixed exactly as you suggested: decided after the entry lookup, from what the manifest ends up holding, with `reset_original_reason` distinguishing "already carried our keys when first recorded" from "no copy could be taken". You were right that this is the same class as round 1's finding 2 pointing the other way, and the `install/SKILL.md` consequence is what made it worth fixing rather than deferring.

**3 — the exit codes.** Fixed, and fixing it exposed a wording bug in **my own round-1 change** that you didn't see because it needed this fix to become reachable. `INCOMPLETE` now covers an environment condition, so wiring it into the empty-plan branch made that branch tell a user with clean files that their "settings files are NOT back to their pre-install state" and offer to repair a file — two lines under a line saying that file already matched its pre-install copy. Which sentence to print is a question about *files*; the exit code is a question about whether anything is left for a human. Two questions, two flags (`FILES_UNFIXED` and `INCOMPLETE`). Clean files with a routed shell now says "Your FILES are fine" and still exits 3.

**4 — docs drift.** Both done: `reset_original=unavailable` is described as the routine already-routed case rather than "rare", and `reset_original_reason` is documented with an instruction to pass it on rather than guess.

**The loopback nit — you're right, and thank you for pricing it properly.** `meta["installed_base_url"]` being written by every `add` branch is the part I hadn't traced; it means the `META` check already covers the entire pre-hatch population and the loopback clause's residue is only hand-edited files. My comment over-priced the false positive and I've corrected it. I also widened the test as you suggested — no-port, `0.0.0.0`, bare `[::1]`, `https://localhost` — since every miss was a false *negative*, the direction that re-introduces finding 1. `TestLoopbackDetectionCoversTheShapesThatMatter` pins all of them plus two must-not-match cases including `not-localhost.example.com`.

**Discoverability — taken in full, and you were right to treat it as part of the feature.** A Troubleshooting entry leading with the literal command on its own line, keyed to the symptom that motivated the PR; the same command in `README.md`; and `TestTroubleshootingLeadsWithTheCommand` asserts it stays in the *first* entry rather than drifting down the section. That last one exists because docs rot silently and this particular line is the one a locked-out user needs.

One more change of my own while I was in there: the diff cap went 14 → 30. A compact original against a pretty-printed rewrite spends its first dozen lines on reformatting, and at 14 the cut landed *before* the user's own added permission grant — the one line they needed. Bounded, still.

On your structural note about step 6 choosing between `settings.local.json` and `settings.json` in prose: agreed, and I'm leaving it. What the script enforces is "not machine-wide", which is the property the blast-radius argument rests on; local-vs-team is a genuine judgement about whether a team shares the routing, and both answers are project scope.

**State:** 91 tests, `go vet` and `gofmt` clean, sandbox run against the real Claude Code install still 20/20.

amiddavid:
## Review round 3 of 3

Re-ran every earlier repro against d752e4d. **All of round 2 is fixed and none of it re-breaks.** The `FILES_UNFIXED`/`INCOMPLETE` split is the right decomposition, `_is_loopback` closes all four shapes I listed (checked each), `held_original` fixes the false `reset_original=unavailable` on a normal uninstall, and the Troubleshooting entry leads with the bare command as asked.

You pointed me at two things. Both were the right places to look — I found something in each.

---

### 1. medium — the filter misses four credential shapes, one of which this repo's own workflow uses

You asked for exactly this. Feeding the shapes through `redact` itself:

```
>   "ANTHROPIC_API_KEY": "<value not shown>",                                   ok
>   "ANTHROPIC_AUTH_TOKEN": "<value not shown>",                                ok
>   "ANTHROPIC_BASE_URL": "https://<credentials not shown>@gw/anthropic",       ok
>   "ANTHROPIC_CUSTOM_HEADERS": "x-guru-auth: REAL-GURU-CREDENTIAL",            LEAKED
>   "authorization": "Bearer eyJhbGciOiJIUzI1NiJ9.REALJWT",                     LEAKED
>   "GITHUB_PAT": "ghp_REALPAT0000000000",                                      LEAKED
>   "ANTHROPIC_UPSTREAM": "https://gw.corp/anthropic?api_key=REALQUERYKEY"      LEAKED
```

`ANTHROPIC_CUSTOM_HEADERS` is the one that matters most, and it is not a hypothetical shape I invented: it is how a Context Guru credential is carried on this project's own dev machines — it is set for interactive sessions and explicitly unset for benchmark runs. So the single most likely credential to appear in a `context-guru` user's `env` block is the one the filter does not catch, and the name contains none of `key`/`token`/`secret`/`password`/`credential`. `Authorization: Bearer <JWT>` has the same problem, and it is the other natural way to put a credential in an `env` block.

Concretely, widen both halves:

```sh
REDACT_NAME='...|[Aa][Uu][Tt][Hh]|[Hh][Ee][Aa][Dd][Ee][Rr]|[Ss][Ee][Ss][Ss][Ii][Oo][Nn]|[Cc][Oo][Oo][Kk][Ii][Ee]|[Ss][Ii][Gg][Nn][Aa][Tt][Uu][Rr][Ee]'
# value shapes, alongside the existing sk- rule:
-e 's/(ghp_|github_pat_|xox[baprs]-|AKIA|eyJ)[A-Za-z0-9_.\/+-]{6,}/\1<value not shown>/g'
-e 's/([?&][A-Za-z0-9_-]*([Kk][Ee][Yy]|[Tt][Oo][Kk][Ee][Nn]|[Ss][Ee][Cc][Rr][Ee][Tt])=)[^&"[:space:]]*/\1<value not shown>/g'
```

`[Aa][Uu][Tt][Hh]` also subsumes `authToken` and catches `ANTHROPIC_AUTH_TOKEN` a second way, which is fine — belt and braces is the right posture for the only thing standing between raw settings content and a terminal.

The structural point behind it: a denylist on a recovery tool is a list that will be wrong again. I am not asking you to invert it to an allowlist — on a diff of arbitrary JSON that would redact the `permissions` entries the diff exists to show, which is worse. But it does mean the list needs a **table-driven test, one row per shape**, so the next miss is a row somebody forgot to add rather than an invisible leak. `TestThePlanNeverPrintsACredential` currently pins three shapes inline.

### 2. medium — a fourth site prints file/variable content, and it is inside `report_environment`

`reset.sh:158` and `:163` print `$base` raw:

```sh
export ANTHROPIC_BASE_URL='https://svc:S3CR3T-in-the-url@gw.corp.example/anthropic'
sh reset.sh --dry-run
#   - ANTHROPIC_BASE_URL is exported in this shell: https://svc:S3CR3T-in-the-url@gw.corp.example/anthropic
```

This is the answer to "is there a fourth site" — and it is the worst possible one, because it sits in the function whose own header comment is *"Values are never printed… a recovery tool that echoes a live API key into a terminal buffer is not a recovery tool."* The function honours that for the credential variables and for the rc-file grep, and then prints a base URL verbatim. `redact` already has the rule that fixes it; the site just does not go through it:

```sh
say "  - ANTHROPIC_BASE_URL is exported in this shell: $(printf '%s' "$base" | redact)"
```

Both branches of the `case` need it (`:158` prints `$base` on its own line too). Embedded `user:pass@` is the common shape; `?api_key=` needs finding 1's rule.

### 3. medium — yes, one other site conflates the two flags, and it is the main success path

Your question 1. `reset.sh:490-499`, the final summary, still tests `INCOMPLETE` for a question about files. A **completely successful** restore, verified unrouted, with a routed shell:

```sh
python3 settings.py add --file .claude/settings.local.json --url http://127.0.0.1:8787/anthropic
export ANTHROPIC_BASE_URL=http://127.0.0.1:8787/anthropic     # hosted agent, or the shell they installed from
sh reset.sh --yes
#   restored: .../settings.local.json
#   ok  .../settings.local.json (no context-guru keys)
#   Finished with something left for you — see the ! lines above.      <- exit 3
```

`Done. 1 file(s) put back.` is never printed. `$RESTORED` is computed and then thrown away on precisely the run where it is the good news, and the user who just successfully recovered is told the run did not finish — the same wording bug you fixed in the empty-plan branch, one branch further down. Same fix, same shape as what you already wrote there: `FILES_UNFIXED` chooses the sentence (and keeps the `Done. N file(s)` count), `INCOMPLETE` chooses the exit code, with a line pointing at the environment report when they disagree.

**Why it survived:** no test anywhere greps for `Done.` — `grep -n '"Done\.' plugin_test.go` returns nothing. `TestAnExportedBaseURLIsNotCalledAFileProblem` covers this exact condition but only on the **empty-plan** path (it runs after a clean restore, so the plan is empty by then). One more case in that test, with a non-empty plan, closes it.

### 4. minor — on your vacuous-test question

The fix you made is the right pattern and worth propagating: `if !strings.Contains(out, "value not shown")` is a *proof the code path ran*, sitting next to the absence assertion, and that guard is what caught your own vacuous pass. Two places to copy it to:

- `TestRemoveFirstNeverProducesAnOriginalHoldingRouting` — the first half iterates `filepath.Glob(originals/*.original)` and asserts nothing when the glob is empty, which is the *expected* state for this fixture. So that half is structurally vacuous and the end-to-end half is doing all the work. It is still a good test because of the second half, but an explicit `len(originals) == 0` branch asserting the record says so would make the first half mean something.
- `TestThePlanNeverPrintsACredential` covers only the **diff** site. Extending it over the verify pass and the environment report is what would have caught finding 2 — the absence assertions there would have failed on `$base`.

Nit, unrelated: `ENV_TMP` is not in the `trap` that cleans `$PLAN`, so an interrupt between `mktemp` and either `rm -f` leaks a temp file. One-line addition to the existing trap.

---

## Where this leaves the PR

Nothing from rounds 1 or 2 re-broke, and every fix in this commit is a real fix rather than a patch over the symptom — the `FILES_UNFIXED` split in particular is the decomposition I would have asked for. The design I said was right in round 1 has held up under three rounds of attack: `O_EXCL`-never-pruned, the hatch outside the plugin, the scope gate in the script instead of in prose, verify-instead-of-trust, and content-not-intent for `_looks_routed_by_us`.

Findings 1 and 2 are the same class as the round-2 blocker — a recovery tool printing a live credential to a terminal whose output its own docs invite the user to paste into a bug report — so I do not want to sign off on them unverified. All three are small and local: two `sed` rules plus a name-class widening, one `| redact`, and one flag swap in a five-line block.

That is my third round, so I am out of rounds and will ask the user for a fourth to verify these three specifically. **Push the fixes; I will re-run the same repros the moment I have the round.** If the user declines the extra round, my position is: mergeable **only** with findings 1, 2 and 3 landed, and the credential-shape table test is the one I would not want to skip, because this is now the third distinct way the same leak has appeared.

amiddavid:
Round 3 addressed in c81d1d1 — all three findings plus the trap nit. Both places you looked had something, and finding 1's `ANTHROPIC_CUSTOM_HEADERS` row is sharper than you put it: that exact shape is in the `~/.claude/settings.json` of the machine I am writing this from, which I had already seen earlier in this session and did not connect to the filter. So the most likely credential in a context-guru user's env block was one I had direct evidence of and still missed.

**1 — the four shapes.** Name class widened with AUTH, HEADER, SESSION, COOKIE, SIGNATURE; value rules added for `ghp_`/`github_pat_`/`xox*`/`AKIA`/`eyJ`, `Bearer <token>`, and credentials in a query parameter.

Your structural point is the one I took most seriously, and `TestRedactCoversEveryKnownCredentialShape` is table-driven with one row per shape — including **rows for what must SURVIVE** (a permission grant, `model`, `theme`, a plain loopback base URL). That direction matters as much as the leak rows: a filter that redacts everything passes every absence assertion and quietly makes the diff useless, so the table pins both failure modes. It lifts the real `redact` out of `reset.sh` rather than copying it, and fails loudly if the extraction stops looking like the filter — a copy would drift from the thing that ships. There's a comment on the filter saying *add a row when you add a rule*.

I agree with your reasoning for not inverting to an allowlist, and I've written that trade-off into the code rather than leaving it in this thread.

**2 — the fourth site.** Fixed, both branches of the `case`. You're right that it was the worst possible one; the function's own header promises values are never printed and it honoured that for the credential variables and the rc grep while echoing a base URL verbatim two lines later.

The lesson I took beyond the one-line fix: my round-2 test was organised by *shape*, which is why a whole site could be missing from it. `TestNoPrinterOfContentEscapesTheFilter` is organised by **site** — plan diff, verify pass, environment report, no-record grep — and each subtest asserts both that the secret is absent **and that the branch actually ran**. Without that second half the environment-report subtest would pass on a build where the branch never executes, which is the vacuity you flagged.

**3 — yes, and it was the main success path.** Fixed with the same split, and your diagnosis of why it survived was exactly right: nothing grepped for `Done.`. `TestASuccessfulRestoreSaysSoEvenWhenTheShellIsRouted` runs a non-empty plan with a routed shell together, and asserts the count is printed, that "Finished with something left for you" is *not*, that the exit code is still 3, and that the output points at the environment report. It also fails fast if no restore actually happened, so it cannot pass vacuously.

That is twice now that this same conflation shipped in a branch I had already fixed elsewhere. I've written the reason into the comment at both sites rather than just the fix, since the next person adding a branch here will face the same choice.

**4 — the vacuous half.** Fixed as suggested: explicit expected count with a message, plus an assertion that the record actually mentions the file. And I've propagated the marker-guard pattern to every new absence assertion in this commit.

**Nit** — `ENV_TMP` is in the trap.

**State:** 94 tests, `go vet` and `gofmt` clean, sandbox against the real Claude Code install still 20/20, and all 14 shape rows behave.

On the fourth round: that is the user's call, not mine, and I've asked them. The fixes are pushed and ready for the same repros whenever you have it. For what it's worth I agree with your position — two of these three are the same class as the round-2 blocker, and "the third distinct way the same leak has appeared" is the right reason not to sign off unverified. If the answer is no extra round, my own read is the same as yours: the three are landed, but landed-and-unverified is not the same as verified, and I would rather that be stated plainly on the PR than implied by silence.

amiddavid:
Two follow-ups to round 3 in the commit above, neither a defect — both are the checks you said you would run, landed so a fourth round verifies rather than discovers.

**The over-redaction risk you named.** You were right to want it checked: `AUTH` and `HEADER` match more broadly than the original five fragments, and `ANTHROPIC_BASE_URL` / `ANTHROPIC_UPSTREAM` / `CONTEXT_GURU_BIN` are precisely what the verify pass and the no-record grep exist to *show* — they are the routing, not a credential. Over-redacting them would leave a locked-out user reading `<value not shown>` where they need to see which port they are pointed at, which would be a quieter version of the same class of failure. Three must-survive rows now pin it, and they pass: none of the three names contains any fragment in the class, so the widening is safe as written rather than safe by luck.

**The provenance, recorded durably** as you suggested, in the table itself rather than in this thread: the `ANTHROPIC_CUSTOM_HEADERS` leak was **live, not theoretical** — that variable carrying a `<header>: <token>` string is how a Context Guru credential is set on this project's own development machines, and it was in the real `~/.claude/settings.json` of the machine the filter was written on, read earlier the same day and not connected to the filter. Shape only; no value is in the repo. Agreed that this is the strongest argument for the table existing at all, and it belongs where the next person maintaining the filter will read it.

Nothing else outstanding. 94 tests, `go vet` and `gofmt` clean, 17 shape rows, sandbox 20/20.

amiddavid:
## Review round 4 — verified, and done

The user granted a fourth round specifically to verify the two credential-leak fixes rather than take them on trust. I re-ran every repro from all three earlier rounds against c81d1d1 (`191ec88` with the follow-up test). **All three round-3 findings are fixed, nothing from rounds 1 or 2 regressed, and I found nothing new.**

### The three fixes, verified

**Finding 1 — credential shapes.** All eight leaking shapes now redact, and all five must-survive lines survive, run through the real `redact` lifted out of `reset.sh` on macOS BSD `sed`:

```
"ANTHROPIC_CUSTOM_HEADERS": "<value not shown>"          was LEAKED
"authorization": "<value not shown>"                     was LEAKED
"GITHUB_PAT": "ghp_<value not shown>"                    was LEAKED
"ANTHROPIC_UPSTREAM": "...?api_key=<value not shown>"    was LEAKED
"AWS": "AKIA<value not shown>" · "slack": "xoxb-<value not shown>"
"ANTHROPIC_BASE_URL": "http://127.0.0.1:8787/anthropic"  survives
"ANTHROPIC_UPSTREAM": "https://gateway.corp.example/..." survives
"CONTEXT_GURU_BIN": "/Users/me/.local/bin/..."           survives
"permissions": {"allow": ["Bash(git push:*)", ...]}      survives
```

The last four matter as much as the first four: `AUTH` and `HEADER` match far more broadly than the original five classes, and over-redacting the routing keys would leave a locked-out user reading `<value not shown>` where they need to see which port they are pointed at. Pinning those as must-survive rows was the right instinct.

On portability, which I flagged to myself as a risk and can now retire: the `Bearer` rule puts an unescaped `/` inside a bracket expression while `/` is the `s///` delimiter, which POSIX leaves to the implementation. My run above proves BSD `sed`; the `authorization bearer JWT` row in `redactShapes` means CI's GNU `sed` on ubuntu proves the other. Both platforms this script runs on are covered by something, which is the only reason I am not asking for a change.

**Finding 2 — the fourth site.** Both branches redact now, including the query-param shape the first fix would not have caught:

```
  - ANTHROPIC_BASE_URL is exported in this shell: https://<credentials not shown>@gw.corp.example/anthropic
  ! ANTHROPIC_BASE_URL is exported in THIS SHELL and points at a local proxy:
      http://127.0.0.1:8787/anthropic?api_key=<value not shown>
```

I went looking for a fifth site and did not find one. `check-proxy.sh`'s `Log from the last attempt: ${LOG}` is a path, not contents, so the hook note is not one. The rc-file grep uses its own stronger `=.*` filter. That accounts for every place either script emits something read off disk or out of the environment.

**Finding 3 — the success summary.** Both directions now behave:

```
routed shell:  restored: .../settings.local.json
               Done. 1 file(s) put back.
               The files are done. What is left is in the environment report above — …
               exit=3, file verified unrouted
clean shell:   exit=0
```

### Regression sweep on c81d1d1

| | |
|---|---|
| R1-1 hatch re-introducing routing | ok — remove-first still records no original |
| R1-2 false "back to pre-install state" | ok |
| R2-1 secret in the plan diff | ok — and the diff still shows `git push`, the change it exists to show |
| R2-2 false `reset_original` on a normal uninstall | ok |
| scope gate | ok — machine-wide file still refused |

### On the vacuous-test question, now that I can read the fixes

`TestRedactCoversEveryKnownCredentialShape` is the strongest test in this PR, and for a reason worth naming so it survives future edits: `if got == tc.line { … "no rule matched it at all" }` means a shape no rule touches **fails even when the secret is coincidentally absent**. That is the general form of the vacuous pass you hit, not a patch over the one instance of it. The extraction guards (`Cut` on `REDACT_NAME`, the `\n}\n` bound, the "does this look like the filter" assertion) mean it cannot silently drift into testing a fragment either.

Reorganising `TestNoPrinterOfContentEscapesTheFilter` by site was the correct inversion. Each of the three subtests asserts the branch ran (`still mentions`, `exported in this shell`, `ANTHROPIC_UPSTREAM`) alongside the absence, and the fourth site — the plan diff — is pinned by `TestThePlanNeverPrintsACredential`. All four printers are covered, each with a positive proof of execution.

Recording the provenance of the `ANTHROPIC_CUSTOM_HEADERS` row in the table comment was the right call. The next person to read that list will treat it as a live constraint rather than a paranoid one.

---

## Verdict: approve

Four rounds, seven findings, three of them severe — a recovery tool that re-applied the routing it existed to remove, a tool that told a locked-out user their files were fine, and a credential leak that appeared in three distinct shapes. Every one is fixed at the cause rather than at the symptom, and each fix came back with a test that pins the general property rather than the instance: content-not-intent in `_looks_routed_by_us`, the `FILES_UNFIXED`/`INCOMPLETE` split applied to both branches that needed it, `held_original` judging the record rather than the call, and a shape table that fails on an unmatched line.

The design I said was right in round 1 held up under all four: `O_EXCL`-never-pruned, the hatch installed outside the plugin, the scope gate in the script instead of in prose, verify-instead-of-trust, and values-never-printed — the last of which is now actually true everywhere rather than in the places it was remembered.

The two known gaps stand as stated and are fine to ship: no model calls, so the install skill's own prose is unexercised, and a stubbed proxy, so nothing here proves a real request survives a real proxy. Both are named in the description, which is the right place for them.

Nothing outstanding from me. Ship it.

OsherElhadad:
## Review — the hatch design is sound, and most of its safety claims survive attack. Three blockers, ten should-fix.

Reviewed against `origin/main` @ `e66f255`. **Every invocation ran under a sandboxed `HOME`**; HOME-honoring was audited by source grep *before* the first execution. The reviewer's real `~/.claude/settings.json` was `cmp`'d three times (start, mid, end) — **byte-identical throughout**, and its contents were never printed. `shellcheck`, `dash`, `busybox sh` and `bash --posix` all exercised.

**What holds up.** The choke-point argument is correct — all six `cmd_add` write exits do funnel through `save()`. The `O_EXCL` copies are `0600` and **not** umask-dependent (verified under `umask 000`). Functional restore is byte-identical under `busybox sh`, `dash` and `bash --posix`. The rc-file scanner leaks in **none** of the adversarial shapes tested (multiple assignments per line, continuations, heredocs, `=$(...)`, embedded delimiters, trailing comments) because it truncates everything after the first `=` rather than matching names — a genuinely better design than a name matcher. And **the catastrophic case I went looking for does not exist**: across six cycles on a pre-existing file, six on a created file, four with the state dir wiped, and a corrupted manifest, `existed_before` **never** flips to `0` for a file the user had. That claim is REFUTED.

---

### BLOCKER B1 — An interrupted first install permanently poisons the recovery path, and `reset.sh` then destroys the user's settings while printing "Done" · CONFIRMED

`settings.py` `_record_touch` (the `elif existed:` branch) + `reset.sh:290`.

Three individually-reasonable links compose into silent data loss:

1. Only `except OSError` unlinks a half-written copy. **`KeyboardInterrupt` is a `BaseException`** — Ctrl-C during the first install propagates past that handler *and* past `record_touch`'s `except Exception`, leaving a fragment. `O_EXCL` then guarantees **by design** it is never replaced, so the damage is permanent.
2. `reset.sh:290` gates on `[ -f "$original" ]` — exists, not non-empty.
3. The advertised "verification instead of trusting `cp`" is a **grep for key names, not a byte compare** — and a *destroyed* file has no `ANTHROPIC_*` keys, so it verifies **clean**.

```
propagated: KeyboardInterrupt
  left behind: claude-settings.local.json.abfd04989ed9.original  8 bytes -> {"model"

# then, with a zero-length original, against a file holding model + permissions + theme:
  restored / verified:   ok  …/settings.local.json (no context-guru keys)
  Done. 1 file(s) put back.      exit=0
  --- FILE AFTER RESET ---  0 bytes
```

Recoverable only via `*.context-guru-prereset-*`, which the summary gives the user no reason to look for.

The irony is precise: **the never-overwrite property that makes the hatch trustworthy is exactly what converts a transient Ctrl-C into permanent unrecoverable state.**

**Fix (both cheap).** In `settings.py`, write to a temp name then `os.link(tmp, dest)` — atomic, and still fails if `dest` exists, so never-overwrite semantics are preserved. In `reset.sh`, use `[ -s "$original" ]` and `cmp -s "$original" "$path"` after the `cp`, incrementing `RESTORED` only on success.

---

### BLOCKER B2 — The scope gate breaks the shipped `/context-guru:statusline` skill and two commands in a doc this PR edits · CONFIRMED

The gate is the **first statement in `cmd_add`**, above the statusline-only early return, so a `--statusline`-only call — which touches no routing whatsoever — is refused:

```
### documented command, verbatim (skills/statusline/SKILL.md:54)
PR 236:  result=error  reason=user_scope_needs_flag   exit=2
main:    result=added                                 exit=0
```

Six documented invocations break: `skills/statusline/SKILL.md:54,81,85,89` (on `main`, untouched here) and **`docs/how-to/install-plugin.md:350,356` — a file this PR edits and did not fix.**

The refusal text is also wrong for this call: *"this file **routes** EVERY project on the machine"* / *"project scope … is the default"*. A statusline is machine-wide by design and routes nothing.

**Fix:** gate on `args.url` being present, or move the gate below the `if not args.url:` branch.

---

### BLOCKER B3 — The state and `originals/` directories are created with the default mode, so under a permissive umask a local user can substitute the file `reset.sh` restores · CONFIRMED

Surfaced by the `0600` check that the copies themselves passed. The **files** are hardened; the **directories** are not. Under a permissive umask they are world-writable, so another local account can replace an `originals/` entry and have `reset.sh` write attacker-chosen content into the victim's `~/.claude/settings.json` — a config file that controls routing and holds a credential.

**Fix:** `os.makedirs(..., mode=0o700)` plus an explicit `os.chmod` after creation (mkdir's mode is umask-masked), and verify the mode before trusting an existing directory.

---

### Should-fix

- **S1 — `reset.sh:171` misses 4 of 6 loopback shapes** (shellcheck SC2102): `[::1]` is an unquoted glob bracket expression, so it never matches. `http://[::1]:…`, `0.0.0.0`, no-port, and `https` all report clean, **exit 0**, while the user's shell stays routed at a dead proxy. `settings.py:_is_loopback` handles all six correctly — and `TestLoopbackDetectionCoversTheShapesThatMatter` tests *that Python function*, not this glob.
- **S2 — the rc scanner covers 4 variable names, not `ANTHROPIC_*`.** A credential in `ANTHROPIC_CUSTOM_HEADERS` (or via `declare -x`) yields *"No ANTHROPIC_\* assignments found"* — a false all-clear on the exact diagnostic the PR added.
- **S3/S9 — the settings-file diff prints values the denylist does not recognize**, in the `--dry-run` output the docs tell users to run first. `redact()` is a **denylist**; independently reproduced with planted canaries, 5 of 12 escaped:
  - `sk_live_…` / `sk_test_…` — the value rule is literally `sk-`, so underscore variants never match
  - secrets in a URL **path** (`hooks.slack.com/services/T/B/<secret>`) — rules cover only `user:pass@` and `?key=`/`?token=`/`?secret=`
  - `?auth=<secret>` on `ANTHROPIC_BASE_URL` — no denylist word in the name, `auth` not in the query rule
  - non-string JSON values on credential-named keys — the rule requires a `"` immediately after the colon
  
  Realistic path: a developer debugging a 401 runs `--dry-run` (documented step 1) and pastes the output into an issue — which the script's own header says is the expected next move — shipping their Stripe key, Slack webhook and gateway token. **The last row prints on every path, including "nothing to restore."**
  
  `plugin_test.go:4645` is the right shape (it drives the real filter) and just needs rows. Suggested: `sk[-_][A-Za-z0-9_-]{6,}`; add `auth|credential|password|pass|sig` to the query rule; fire the JSON-name rule on non-string values; treat a URL field as opaque beyond the host when the path has ≥3 segments. **`ANTHROPIC_BASE_URL` must stay partly visible** — showing the localhost redirect is the diagnostic's whole point — so redact only its userinfo, query, and path beyond `/anthropic`.
  
  **Design note beyond the rows:** a denylist in a tool whose output is meant to be pasted publicly will leak every future secret shape until someone remembers to add it. An allowlist (print only keys known to be safe) inverts the failure mode.
- **S4 — `settings.py`'s `emit()` prints the user's previous base URL verbatim, credentials and all.** Pre-existing, but it contradicts this PR's own stated principle.
- **S5 — `go test` is not hermetic.** It fails where `CLAUDE_CONFIG_DIR` is set (as it is on this box), and **it writes into the developer's real `~/.local/state/context-guru/`** — reproduced live; four stale pidfiles were left there and had to be cleaned up by hand. Cannot be fixed correctly without **S10** (`start-proxy.sh:252` ignores `CONTEXT_GURU_STATE`, unlike its three siblings).
- **S6 — `*.context-guru-prereset-*` is a new family of full settings-file copies, not gitignored, written inside project working trees.** `git status` lists them, and that visibility is the only control between this and a committed credential.
- **S7 — `TestTheHatchNeedsNothingButPOSIXSh` asserts a shebang string and five banned words, not the condition.** This repo has commits titled *"assert the fallback CONDITION, not one sentence that expressed it"*; running the script under `dash`/`busybox` is the actual assertion.
- **S8 — `TestHatchReportsCredentialsByLocationAndNeverByValue` covers one easy shape.** It passes while S3/S9 leak.

---

### Summary

The design intent is right and several of its claims are stronger than they needed to be. But a recovery tool has an unusually high bar — it runs when everything else is broken, on a machine whose owner is already in trouble — and on that bar B1 (silent destruction of the file it exists to protect) and S3/S9 (credential disclosure on the documented first command) both fire on the *happy* documented path, not an exotic one. Those two plus B3 are worth fixing before merge; the rest can follow.

OsherElhadad:
## Follow-up: S5 reproduced live during this review, and the fix the PR applied is insufficient

Two additions after the full audit, plus a sharper read on one of the redaction shapes.

### S5 — the test-isolation defect this PR claims to have closed, reproduced by running the PR's own tests

While running `go test ./context-guru-plugin/...`, the hook tests spawned `start-proxy.sh`, which resolved the **real** `~/.local/state/context-guru/` and wrote 4 pidfiles into it. Every invocation of `settings.py` and `reset.sh` in this review was sandboxed under an overridden `HOME`; this came from the test suite itself.

The cause is `plugin_test.go:285` — `cmd.Env = append(os.Environ(), …)` — so the hook tests inherit the real `HOME`. The PR's notes state this was fixed, on the grounds that *"a test suite for a recovery tool should not be able to disturb the developer's recovery tool."* **It was fixed for the `settings()` helper only.** The 4 pidfiles are a live reproduction of the defect the PR claims to have closed, produced by the PR's own test suite on a developer machine.

Cleanup, for completeness: all 4 removed after confirming each named a dead process; nothing pre-existing was deleted (27 entries, all dated Sep 6–7, `dashboard-8787.db{,-shm,-wal}` intact, zero files dated Sep 14). `~/.local/bin` was never touched. Net change to that directory: zero files added, changed or lost.

### S10 — `start-proxy.sh` ignores `CONTEXT_GURU_STATE`, so the applied fix will not isolate the hook tests

`settings.py`, `reset.sh` and `check-proxy.sh` all honor `CONTEXT_GURU_STATE`. `start-proxy.sh:252` does not — it honors `HOME` and `XDG_STATE_HOME` correctly, but not that variable. So **pinning `CONTEXT_GURU_STATE`, which is the fix this PR applies, does not isolate the hook tests**, and a user who relocates their state directory gets the pidfile written somewhere else — which is one way uninstall loses track of a running proxy.

`start-proxy.sh` is not in this PR's diff, but S5 cannot be fixed correctly without it. Worth pulling into scope or filing alongside.

### On the `redact()` shapes — shape 3 is a one-word fix

Independent verification by lifting `redact()` out of the script and driving it directly, rather than taking the earlier findings on trust: **all four shapes confirmed.** One refinement worth acting on first:

**`?auth=<secret>` prints in the clear because the query-parameter rule carries its own narrower alternation (`key|token|secret`) instead of reusing the file's own `REDACT_NAME` — which already includes `AUTH`.** That is an internal inconsistency rather than a missing case, so it is a one-word fix and the cheapest of the four.

Also worth noting for shape 4 (non-string value on a credential-named key): the leaking fixture was **`"apiKeyHelper": {"cmd": "…"}`** — and `apiKeyHelper` is a real Claude Code settings key whose entire purpose is producing the credential. So that shape is not hypothetical on the exact file this tool prints.

amiddavid:
Thank you — this is the review this PR needed. All three blockers and all ten should-fix items are addressed in d189552. I reproduced every finding before touching it; two of them were worse than I would have guessed from reading the diff.

## Blockers

**B1 — confirmed, and it is the worst defect this hatch has had.** Your chain is exactly right, and I reproduced the whole thing: 8-byte fragment, permanently locked in by `O_EXCL`, restored over a real settings file, `Done. 1 file(s) put back.`, verified "clean", exit 0. Your framing of the irony is the part I want to keep in the code: the never-overwrite property that makes the copy trustworthy is precisely what turns a transient Ctrl-C into permanent unrecoverable state. It is in the docstring now.

Both halves of your fix, taken as given: `settings.py` streams into a temp, `fsync`s, then `os.link`s into place — atomic, still fails if the destination exists, so never-overwrite is preserved without a partial destination ever being reachable — and the temp is removed on every path including `BaseException`, which is allowed to propagate because a user pressing Ctrl-C means it. `reset.sh` uses `-s` and `cmp -s` after the `cp`, and nothing counts toward `$RESTORED` without that compare passing. I also fixed `backup()`, which had the identical hole and whose partial the install skill reports to the user as "that is their undo".

**B2 — confirmed.** Gated on `args.url` now, which is the honest predicate: the blast-radius argument is entirely about where model traffic is pointed, and a statusline is machine-wide by design and routes nothing. Your point about the refusal *text* being wrong for that call was the tell that the gate was in the wrong place, not just too broad.

**B3 — confirmed under `umask 000`, both directories `drwxrwxrwx`.** `0700` with an explicit `chmod` (mkdir's mode is umask-masked, as you noted), and an existing directory is tightened rather than trusted.

## Should-fix

**S3/S9 — inverted to an allowlist.** You were right and the earlier reviewer was right, and the two constraints turned out to be compatible because the cases have different shapes: `"key": value` lines print the value only for keys known to be safe, while bare array elements print only if they match the permission-grant shape `Tool(...)`. So an unknown credential key is covered the day it is invented, and the diff still shows the grants it exists to show. All four of your shapes fixed: `sk[-_]`, URL paths, `?auth=` (your one-word diagnosis was right — the query rule carried its own narrower copy of a word list that already contained `auth`), and non-string values. `apiKeyHelper` is now a test row with your observation about what that key is *for* recorded next to it.

The routing keys stay partly visible exactly as you specified — scheme, host and first path segment, losing userinfo, query and anything deeper.

**S1** — fixed, and `shellcheck --severity=warning` is clean across all four scripts now. **S2** — `ANTHROPIC_[A-Z0-9_]+` plus `declare -x`/`typeset -x`; widening the name side costs nothing because the `=.*` truncation you praised is what actually keeps it safe. **S4** — emitted URL facts go through a redactor; the file keeps the real value so uninstall can still restore it. **S6** — `prereset/` copies live under the `0700` state directory now, not in the user's git tree; the uninstall skill also stops describing that directory as leftovers, since deleting it throws away the user's settings copies. **S7** — the POSIX claim is now asserted by RUNNING the hatch under `sh`, `dash`, `bash` and `busybox` and checking it recovers; the word-list test stays as a cheap static guard. **S8** — covered by the new rows plus tests organised per printing site.

**S5 + S10 — fixed together, as you said they had to be.** One `sandboxEnv` helper that every subprocess goes through (29 call sites), and `start-proxy.sh` now honours `CONTEXT_GURU_STATE` like its three siblings. Sorry about the four pidfiles.

## Two things I got wrong while fixing this

Both caught by tests, both worth recording:

1. My wholesale replacement of `redact()` **deleted the `is_loopback_url` function I had just added for S1**, leaving a call site with no definition — which in `sh` evaluates false silently. That is S1's own failure mode, reintroduced by the fix for S1.
2. My first allowlist assumed the indented one-key-per-line shape `settings.py` writes. A **compact one-line settings file** — which is the `<` side of every diff this filter runs on — never engaged it, and an `authToken` printed in the clear. It scans every `"key": value` on a line now, which also handles nesting for free.

## On your refutation

You went looking for `existed_before` flipping to `0` for a file the user had, across six cycles on an existing file, six on a created one, four with the state dir wiped, and a corrupted manifest — and refuted it. That was my own stated worry, so having it attacked properly and found absent is more useful than the finding would have been.

## Caveats I want on the record

- **I cannot run the Go suite natively** (no Go toolchain on this machine), so it runs in a `golang:1.26` container — which means I could not reproduce S5's native symptom, only fix its two mechanisms. I verified S10 natively instead: with a pristine `env -i`, the pidfile lands under `CONTEXT_GURU_STATE` and nothing appears in `HOME`'s state directory or the real one.
- **The allowlist has a residual gap you may want to look at:** a bare array element inside a *compact* one-line document, under an unsafe key, is not classified by key — the value-shape rules are the only layer there. Judging it correctly needs nesting state, which means a JSON parser, which is the cleverness this script should not contain. Named rather than hidden.

95 tests, `go vet`/`gofmt`/`shellcheck` clean, sandbox run against a real Claude Code install still 20/20, and both interrupt paths verified to leave no fragment.

PR #248, opened 2026-09-14
docs(plugin): design a deterministic install, and name what makes it one

amiddavid:
## Why

`skills/install/SKILL.md` is 423 lines, 39 of them some form of *"do not improvise this"*, *"this was a real defect"*, or *"observed on a hosted agent"*. Every defect in its history is the same shape: **the prose was right, and the emitted command differed from it.**

On 2026-09-14 a colleague installed the plugin on a hosted setup. He hit both auto-mode gates the docs predict, and finished the install by pasting two commands, ~150 characters each, that a model composed and a human could not check:

```
! ".../scripts/start-proxy.sh" --unrouted --upstream "ete-litellm.ai-models.vpc-int.res.ibm.com"
! ".../scripts/settings.py" add --file /Users/.../.claude/settings.local.json --url "..." --upstream "..."
```

Neither gate is a bug — `docs/how-to/install-plugin.md:189` predicts both, and line 184 records that a pre-granted permission rule clears them. He never saw that advice, because the skill surfaces it only *after* a denial has already stopped the install.

This is a design document. **No code changes.**

### Today, as it actually went

```
/plugin marketplace add rossoctl/context-guru
/reload-plugins
/plugin install context-guru@context-guru
/reload-plugins
/context-guru:install
        ↓ blocked by auto mode
! ".../scripts/start-proxy.sh" --unrouted --upstream "ete-litellm...ibm.com"
continue
        ↓ blocked again
! ".../scripts/settings.py" add --file /Users/.../.claude/settings.local.json --url "..." --upstream "..."
```

### Today's documented best case, if you read the how-to first

```
/permissions
        ↓ paste: Bash(/Users/you/.claude/plugins/cache/context-guru/**)
/plugin marketplace add rossoctl/context-guru
/reload-plugins
/plugin install context-guru@context-guru
/reload-plugins
/context-guru:install
```

### After

```
/plugin marketplace add rossoctl/context-guru
/reload-plugins
/plugin install context-guru@context-guru
/reload-plugins
/context-guru:install
```

Then **one** question in plain English, because `--plan` has already run and the script knows the real values:

> You're already pointed at `ete-litellm.ai-models.vpc-int.res.ibm.com`. I'd put context-guru in front of it, so that gateway keeps handling your login. Keep-alive will be on as `5-min-ping`, which spends a little of your own quota on idle turns to hold the cache warm. Routing this project only. OK?

You type `yes`. Then **at most one** approval prompt on a fixed command string that takes no arguments.

### DAM / attach

```
/context-guru:install --mode attach --base-url https://gw.dam-internal/anthropic
```

Nothing installed, nothing started, one key written.

### Fully deterministic, no model in the path at all

```
! cg-install
```

A `!` line in the REPL is the user's own shell command: no model composes it, no model reads it before it runs, and no classifier gates it — which is precisely why the two denied steps completed the moment they were pasted by hand. The repo already ships this pattern for the reset hatch, which lives at a plain path outside the plugin so it works with "no Claude, no network, no proxy and no plugin".

### Cache strategy

| | Today | After |
|---|---|---|
| turn keep-alive on | `/context-guru:keepalive`, then stop and restart the proxy by hand | on by default as `5-min-ping` |
| change it | edit YAML, restart | `/context-guru:cache-strategy-picker` |
| set at install | not possible | `/context-guru:install --cache-strategy split` |
| go back | remember which four tuning numbers you had | `5-min-ping` |

### Scoreboard

| | Slash commands | Pasted bash | Questions |
|---|---|---|---|
| Colleague's install | 5 | **2** (~150 chars each) | 1 |
| Today, best case | 6 + a pasted rule | 0 | 0 |
| After | 5 | **0** | 1 plain-English, maybe one `y` |

The win is not keystrokes — today's best case is already competitive. The win is that **no human ever pastes a command a model composed.**

## What the design proposes

1. **`install.sh --route`** — steps 1–7 as one script. Ordering becomes structural rather than a numbered paragraph (writing settings before starting the proxy once killed the installing session). Auto-rollback if the post-write health check fails, instead of asking the model to *offer* a rollback. `--plan` is read-only and emits the same `key=value` block.

2. **`!` pre-execution is the part that actually removes latitude.** A script alone is not deterministic — the model still reads the skill, works out it should run `install.sh`, and types the command. `` !`<command>` `` in a skill body runs *before* the content reaches the model, so the command string is fixed in a file no model rewrites. This is what makes the 2026-09-14 hand-composition unreachable.

3. **Generalization: a gated command takes no model-composed arguments.** The two judgment calls travel through a decision file, not `argv`, so the one traffic-redirecting command is a constant:
   ```
   1. !`install.sh --route --plan`   runs at render, read-only, no arguments
   2. model asks its one question in prose
   3. install.sh --decide --scope project --on-conflict chain   (no settings touched → not gated)
   4. install.sh --route --confirm --scope project --on-conflict chain
                                    an ordinary tool call that SAYS what it will do
   ```
   **The decisions go in `argv`, and an earlier revision had this backwards.** A constant gated command (`install.sh --route --confirm`) names neither the scope, the file, nor the upstream — it optimises away model-composed arguments at the cost of the one screen where a user could verify anything, leaving *two* approvals, neither load-bearing. One yes, on a command that describes itself, is better than two on commands that do not. What remains in argv is a closed vocabulary the script validates (exit 2 on anything unrecognised); in `local` mode the URL is derived from the resolved port and never passed. `attach` keeps a decision file, since there the URL is a real input.

   Step 4 is still **not** a `!` block: every `!` block fires at render, so it would install before the user was asked.

4. **`--mode attach` for DAM.** The standing recommendation is to land the proxy in the gateway, so there is no local port to derive a URL from. `attach` skips install and start entirely and takes a validated `--base-url`. That makes `valid_base_url()` a prerequisite rather than a nicety, since `attach` exists to let a human type a URL.

5. **Named cache strategies, one vocabulary in both deployments.** `tenant.Strategy` already carries a `Name` plus exactly the tuning fields, with durable manager-gated CRUD — so `5-min-ping` should be the name of a seeded strategy *and* the name in the local YAML. Arming in `attach` needs a manager principal; disarming never does, because withdrawal must not be harder than consent.

6. **`1-hour-head` ships labelled, not recommended.** `config/config.go:132` measured it GRANTED on Haiku 4.5 and *silently downgraded* on Sonnet 5, with zero 1h writes across 19,805 production requests.

## Two corrections this document makes to its own earlier revisions

Recorded in place, because both were load-bearing while they were wrong:

- **Uninstall is not affected by the unvalidated URL.** `is_ours()` matches on the recorded `$context-guru.installed_base_url`, not on URL shape — deliberately, since litellm's default is `http://127.0.0.1:4000/anthropic` and two local proxies are indistinguishable by URL. Verified: `remove` cleans up a malformed URL correctly. The defect is confined to `add` writing an unroutable value and reporting success.
- **`--plan` is not "approvable, therefore informed consent."** A `key=value` block a user cannot evaluate and a bash command they cannot evaluate are the same opaque approval. What improves is the *count* of approvals (two-or-more → one) and that the model's question is concrete; legibility comes from the plain-English sentence, not the output block.

## Measured — probed in a sandbox, not assumed

Run as a throwaway `claude -p` session (`--plugin-dir` for a disposable plugin, isolated `CLAUDE_CONFIG_DIR`, `--permission-mode auto`), routed **off** Context Guru so the synthetic traffic neither contaminated Guru's data nor touched the live session.

**1. `!` pre-execution works.** The block ran during rendering and the model received the *output* (`RENDERED: PROBE_MARKER_9F3A ok`). A sentinel file written by the script confirms execution independently of what the model reported.

**2. `${CLAUDE_PLUGIN_ROOT}` is substituted into the command STRING but is NOT exported to the child process.** Both halves are true and they look contradictory until separated: the script was found and ran, yet its own read of `$CLAUDE_PLUGIN_ROOT` reported `EMPTY_OR_UNSET`.

> **Live trap.** Any script invoked from a `!` block must locate itself from `$0`. `check-proxy.sh:54` reads that variable internally today — correct for a hook, where plugin variables *are* present, but it would silently no-op on the `!` path, with no error.

**3. RETRACTED — the classifier's treatment of the two paths is not established.** A second probe added the control the first lacked. Four trials, one per cell:

| # | path | outcome |
|---|---|---|
| 2a | `` !`start-proxy.sh --unrouted --upstream https://…` `` in a skill body | ran |
| 2b | identical command, model tool call, from a **bare prompt** | **denied** — *"`--unrouted` read as a routing/egress bypass"* |
| 3a | model tool call from a skill **with** `allowed-tools` | ran |
| 3b | model tool call from a skill **without** `allowed-tools` — the control | ran |

3b is what breaks the earlier conclusion. The control ran, so `allowed-tools` did not separate 3a from 2b — and 2b differed from 3b in **two** ways at once: bare prompt vs skill invocation, and a different script path that the denial text explicitly cited (*"an unverified script from `/tmp/cgprobe`"*). With the variable not isolated, the denial cannot be attributed to the code path.

**The classifier is also itself a model, so it is probabilistic.** One trial per cell can show a denial is *possible* on a path; it cannot show another path is *exempt*. Honest summary: `!` blocks not gated in 2 trials, tool calls gated in 1 of 3. Establishing exemption needs ~10 trials per cell with one variable moving.

**So nothing in the design depends on predicting it.** The install must be correct whether or not the prompt appears — which is why the design stops trying to make the gate vanish and makes it *legible* instead.

Findings 1 and 2 stand: they are deterministic mechanism facts, not statistics.

**4. Still unanswered:** whether `~/.local/bin` is on the PATH of the shell `!` uses. It is on the author's machine — one data point, not a fact about users. This only affects Appendix A.

## Appendix A: the PATH-shim option, set aside

The doc records the fully model-free form — `!~/.claude/plugins/marketplaces/context-guru/context-guru-plugin/scripts/install.sh --route`, whose path is **stable across plugin updates** (verified: the install cache is versioned or content-hashed, the marketplace checkout is not) — and why a `~/.local/bin` shim on top of it is not the recommendation: it writes an artifact outside every mechanism that exists to remove things. `uninstall` does not know about it, the reset hatch does not record it, and unwinding it is harder than unwinding one settings key.


amiddavid:
## Superseded in #250: the consent mechanism is no longer the approval prompt

Flagging this here rather than silently, because a reviewer reading this document alone will take away
a conclusion the implementation has since reversed.

**What the document says.** Under *"What finding 3 must not be used for"*, it argues that the
traffic-redirecting `--confirm` step should stay an ordinary tool call **"deliberately, so that it is
classified"**, and that **"the gate on that one step is a feature."** The design therefore rests its
consent story on the auto-mode approval prompt.

**Why that no longer holds.** This document's own Measured findings undercut it, and it says so two
paragraphs earlier without following the thread to its conclusion — the prompt is:

- **probabilistic** (finding 3, retracted: the same command was denied in one trial, allowed in two);
- **declarable-away** (finding 5: with `allowed-tools: Bash(.../install.sh)` the command ran with no
  prompt at all);
- **absent by construction** in unattended runs — where the measured outcome was a *complete, silent
  install of a traffic interceptor*.

A gate with those three properties cannot be the thing standing between a user and having their model
traffic intercepted.

**What #250 does instead.** `--route` refuses to act at all without
`--i-consent-to-traffic-interception`. The gate sits before step 1, so a refusal installs nothing,
starts nothing, writes nothing, and an unattended run now **fails closed**. Verified in the same
`claude -p` scenario that previously installed unattended.

**Stated plainly, because it is the weak joint:** the flag is a *precondition, not a proof*. A caller
can pass it without asking anyone, and no script can detect that. What changed is that consent went
from "we hope a prompt fires" to "this refuses without a token that exists only because someone was
asked" — deterministic and auditable, but not evidence of a human. The defence against a model
appending the flag anyway is the install skill's instruction (*a silent or absent answer is a NO*;
*passing this asserts on the user's behalf that a person agreed*) plus a test that the skill still
carries it. That is prose with a test behind it, not a mechanism, and a reviewer should weigh it as
such.

The document's `## Status: implemented in #250` line remains accurate; the prose in that one section
does not, and should be amended to describe the classifier gate as *defence in depth on top of* the
flag rather than as the mechanism. I am isolated to the orchestrator worktree and cannot amend this
branch from here, so this comment is the record until it is.

PR #234, opened 2026-09-10
feat(summarize): the pre-expiry summary gate — compact only when the write was already due, off the hot path

amiddavid:
`summarize` now compacts only at a moment when the cache write it costs was going to be paid anyway, and produces the summary **off the hot path** so that deciding to compact cannot itself become the expensive outcome. A dashboard panel measures what each summary actually earned.

Closes #233. Closes #243.

Three defects were found and fixed *by* this work, two of them in the first version of it. They are the most interesting part of the diff.

## The rule

> **Only ever spend a cache write that was going to be spent anyway.**

Compacting rewrites history, and rewriting history the provider is holding in its prompt cache invalidates that entry — the next turn re-writes the whole suffix at 1.25x fresh. There are two moments when that costs nothing:

| Cache state | Without compaction | With compaction | |
|---|---|---|---|
| **warm** | a cheap read | a write, **and a live entry destroyed** | the only harmful case |
| **near expiry** | a read now; a full rewrite later if the next turn is late | a small write now, cheap reads after | better, *given* an assumption about the next turn's timing |
| **already expired** | a **full** rewrite — the entry is gone, this turn pays a write regardless | a **small** write | **strictly better, unconditionally** |

So the default is `cache_state: pre_expiry_or_cold` — compact whenever it cannot hurt. Note the ordering: *expired* is the reliable half and *pre-expiry* is the half carrying an assumption. The first version of this PR had that backwards, because it reasoned about not destroying something valuable rather than about riding a payment already being made.

Expired is also the **reachable** half. Pre-expiry needs a request to land inside a 60-second slice of a 5-minute lifetime; any gap past the TTL produces an expired turn on the next request.

**Cold is decided by the clock, never by the flag.** `Ctx.ColdCache` is a known false positive on keep-alive'd sessions — `proxy/keepalive.go` never updates the turn tracker, the −$708 mechanism `promexport.go:807` documents — so trusting it would compact **live** prefixes on exactly the sessions someone is paying pings to protect. `coldByArithmetic` requires a known TTL and idle time whose difference is ≤ 0.

## Defect 1 — the fill fraction could never fire (`85130c6`)

`min_request_frac` compared `frac × window` against `schema.MessagesTokens`. Two different rulers: `MessagesTokens` counts message **text only**, while a context window is stated in the tokens the provider bills. This repo already measures the gap — `dash/overview.go`'s `EstimatorDivergence`, over requests where nothing was compacted, reports a median **3.38x**.

[...]

PR #252, opened 2026-09-15
feat(summarize): a configurable head, and an operator prompt

itay-nakash:
## The defect

`summarizeSpan` pinned a hardcoded `headCount = 1`, and **"preserve `msgs[0]`" means two different things across the dialects — only one of them keeps the task:**

| | what `msgs[0]` is | task statement |
|---|---|---|
| Anthropic traffic | the opening user turn (the `system` prompt is a top-level field the pipeline never sees) | **pinned** ✅ |
| OpenAI-shaped | the system prompt | `msgs[1]` — the **first message of the summarized span** ⛔ |

So on OpenAI-shaped traffic the instruction the agent is working from was folded into the summary on every eligible turn, and no config could stop it.

It was survivable rather than obvious because `conversationGoal` scans by **role**, not by index: the task text still reaches the summarizer as the goal, so the summary is written *toward* a task whose verbatim statement is gone. It was reported to me as "the summarizer compacts the starting messages".

## The change

**`keep_first`** (default **1**, unchanged behaviour) makes the head reachable, so an OpenAI-shaped deployment sets `2` and retains `[system, task]`. A negative value is refused rather than clamped — reading a typo as 0 would summarize the task away, the one thing the head exists to pin.

**The head rule generalises rather than moves.** The old code dropped `headCount` to 0 when `msgs[0]` was an assistant message carrying tool calls, since its results lie inside the span and keeping it emits an unanswered call. That is now a retreat loop over the head's *last* message: it reduces to exactly the old rule at `keep_first: 1`, and keeps a call/result pair whole at `keep_first: 2`. A loop rather than one decrement, because consecutive tool-calling assistant messages are malformed input and malformed input is where a provider 400 comes from.

**`custom_prompt`**, and the frame around it is the point: a bare instruction pasted after a long trajectory reads to the model as more trajectory and gets summarized instead of followed, so it is wrapped in text stating whose instruction it is and what precedence it has. Two limits survive it — the ban on inventing anything absent from the conversation (the base prompt's entire purpose) and the `<summary>` output format.

**`summarizer1509`** is a **preset, not a component**: `summarize` configured as the cache-unaware, no-reuse reference summarizer (`cache_state: any`, `resummarize_tokens: 0`, `keep_first: 2`) — the control arm the cache-aware trigger's payoff is measured against, and the right starting point on a backend whose prompt-cache lifetime this repo cannot derive. A preset because turn-aligned cuts, orphaned `tool_result` repair and now a configurable head all live in `summarize` already; a separate component would duplicate ~500 lines of summarizer to deliver three YAML values.

## Tests

`summarize_keepfirst_test.go` drives the **real two-turn path** — the summary is commissioned off the hot path, so turn 1 forwards untouched and turn 2 splices — draining on `WaitForSummaryForTest` rather than sleeping. It carries a vacuity guard, because a fixture that stops reaching the span makes every head assertion pass for the wrong reason.

Four cases: the default reproduces the historical shape (task dropped); `keep_first: 2` retains it; the dangling-tool-call retreat; and `keep_first: 0` honoured vs negative refused. The `keep_first: 2` case also re-asserts the no-system-role-away-from-index-0 contract, since a larger head is a new way to violate it.

`go vet` and `gofmt` clean; `./components/...` and `./config/...` pass. Two pre-existing failures in `cmd/context-guru-proxy` are environmental on my machine (`install.sh` needs bash > 3.2, and one test finds a binary already installed at `~/.local/bin`) — neither package is touched by this diff.

## One note, not a change

`TestSummarizeEmitsNoSystemRoleAwayFromTheHead` currently **skips** on its own fixture (`summarize declined on this fixture`) — the `min_request_frac` 0.9 default plus `FracResolvable` means it never reaches the acting path, so the guard against the `400 messages.1: role 'system' must precede...` regression is not actually exercised. My new test pins `min_request_frac: 0` / `cache_state: any` to get there. Worth a follow-up on that fixture; I left it alone to keep this diff to one subject.


OsherElhadad (commented):
## Review

Reviewed the diff (`components/offload/summarize.go`, `summarize_pairing.go`, `config/config.go`, docs) and ran it against the real `Offload` path (not mocks) with the two-turn commission/splice flow the component actually uses.

**Code quality — solid.** `summarizeSpan(msgs, keepFirst, keepLast)` clamps `keepFirst` to `[0, len(msgs)]` and the tool-call retreat is now a loop rather than a single decrement, which correctly generalizes to `keep_first: 2+` while reducing to the old behavior at 1. `keep_first: -1` is refused at config-load time (`newSummarize`) rather than silently clamped to 0 — good call, since 0 is a real, distinct configuration ("summarize everything").

**Tests** (`summarize_keepfirst_test.go`) drive the real two-turn commission/splice path via `WaitForSummaryForTest`, not a stub of `summarizeSpan` in isolation, and include a vacuity guard against the fixture not reaching the span. Four cases: default (1) reproduces the historical task-drop on OpenAI-shaped traffic, `keep_first: 2` retains it, the dangling-tool-call retreat, and 0-honoured/negative-refused. I ran these plus the full `components/offload` and `config` suites locally — all pass.

**Security — no injection risk.** `custom_prompt` is deployment-time config (YAML), never per-request/user-controlled, so this isn't an untrusted-input path. The wrapping frame (`summarizerCustomPromptFrame`) is a reasonable belt-and-suspenders anyway: it explicitly tells the model not to treat the operator text as part of the conversation, and keeps the anti-fabrication rule and `<summary>` format non-negotiable.

**Docs gap (fixed).** `docs/components/summarize.md`'s configuration table documented every other key (including `keep_last`) but not the two new ones. Added rows for `keep_first` and `custom_prompt` there.

**Merge conflicts (fixed).** `main` moved substantially since this branch's base — notably `docs/how-to/choose-a-preset.md` was folded into `docs/reference/presets.md` in a nav-trim pass (commit `e36cc78`, itself authored to keep `docdrift_test.go` from checking a deleted file). This PR's edits to the now-deleted file, and its `presets.md` addition against the old table shape, needed to be redirected into the merged tables (`Which preset?`, `What each one runs`, `Exact pipelines, verbatim`) instead. Also: `components/offload/cache_aware_summarizer.go` (landed on `main` after this branch's base, PR #234) has its own call to `summarizeSpan` still on the old 2-arg signature — fixed to pass `keep_first: 1` (unchanged behavior for that component). Three test helpers (`holdsText`, `roles`, `itoa`) were independently added by both branches with identical bodies — kept `main`'s copies, dropped the duplicates from this PR's new test file.

Pushed the merge + fixes directly to this branch (`maintainerCanModify` is enabled, so pushed straight to `itay-nakash/context-guru:feat/summarize-keep-first`) rather than opening a supplementary PR.

**Verified:** `go build ./...`, `go vet ./...`, `gofmt -l .` (clean), `go test ./components/offload/... ./config/...` (includes `TestPresetDocsDoNotDrift`) all pass on the merged branch. `go test ./...` has two pre-existing failures (`dash.TestCalculatorPricesFromTheRowsModelNotABlend`, `dash.TestNavHashCompatibility`) that reproduce identically on `main` alone with no changes from this PR — unrelated. The `context-guru-plugin` package's environmental failure the PR description already called out also reproduces on a fresh clone. Did not run a live `claude -p` proxy session against a real backend (would need real credentials/cost and a long multi-turn transcript to cross `summarizer1509`'s 20k-token trigger) — the two-turn `Offload` path is exercised directly and identically by the unit tests instead.

OsherElhadad (approved):

PR #254, opened 2026-09-15
test(summarize): pin the `cache_state: any` configuration on a LIVE cache

amiddavid:
## Why

`summarize` ships **two** configurations, and they differ in exactly one key:

```yaml
# A — cache-aware (the default; writing it is optional)
trigger: { cache_state: pre_expiry_or_cold }

# B — size-only: fire at the fill threshold regardless of cache state
trigger: { cache_state: any }
```

`min_request_frac` defaults to `0.9` in both, so the cache condition is the only thing an operator writes.

B is documented (`docs/components/summarize.md:226`), both example configs under `examples/llm-d-service/` run it, and #234's own docstring calls it out as the opt-out. What was **not** asserted anywhere is the sentence B actually rests on: *it fires while the prompt cache is still live.* Every existing subtest exercises `any` against phase **Unknown** — the one phase where `CacheAllows` has an early return.

It is true today. `p` is not read after that early return, so warm and unknown take the identical route for `any`. But that is a claim about control flow, and the scoping of that early return has already drifted once in this file — the review on #234 found the gate opening over a live 8-message prefix on 13 of 26 turns.

## What landed

Two tests, each a **pair** run against the same transcript and the same `Ctx`, because a single-sided assertion here is worthless:

| phase | `cache_state: any` | the shipped default |
|---|---|---|
| **Warm** (270s of TTL left) | fires | declines, `cache_state_declined_warm` |
| **Unknown** over a live prefix (`CacheAware`, `MaxCachedIdx >= 0`) | fires | declines, `cache_state_declined_unknown` |

A test showing only that `any` fires would still pass with the cache gate deleted outright; one showing only that the default declines would still pass with `any` broken into a no-op. The pair is what pins the difference to the one key.

The second row is the regression the first cannot catch. That guard excludes `any` **by name**, because a raw-API harness with nothing capping its context is precisely the deployment that can present a cached prefix and no readable TTL — drop the clause and configuration B silently stops firing on the deployments the opt-out exists for.

`min_request_frac` is written in neither config, so the `0.9` default is under test as the half the two share, with a precondition asserting it resolved. Preconditions also assert each fixture's own phase: a warm fixture drifting into pre-expiry would make the default fire too, and the test would report an agreement it never demonstrated.

## Verification

Both tests were **mutated to failure**, each hitting only its own target:

- removing `t.CacheState != CacheStateAny` from the unknown-phase guard → `TestCacheStateAnyStillFires…/any_fires_despite_the_live-prefix_guard` fails (`fired=false want true (gates: map[cache_state_declined_unknown:1])`); the warm test still passes.
- making the `any` branch phase-dependent (`return p != CachePhaseWarm`) → `TestCacheStateAnyFiresOnALiveCache…/cache_state:_any_fires_on_a_live_cache` fails (`fired=false want true (gates: map[cache_state_declined_warm:1])`); the unknown test still passes.

`gofmt` and `go vet` clean; `CGO_ENABLED=1 go test -race ./components/...` all green.

Tests only — no behaviour change.

## Context

Raised while confirming which configuration to run now that `summarize` sends a single flattened prompt through `Model.Complete(ctx, string)` and therefore shares no prefix with the agent's cached `tools → system → messages`. Until a cache-reusing summarizer lands, B is the configuration worth measuring against A, and it needed to be pinned rather than merely documented.


amiddavid:
## Review (single round) — the pair shape is right; one subtest can pass for the wrong reason

Tests only, so I verified by inspection rather than by running: no Go toolchain on this machine, and the claimed `gofmt`/`go vet`/`-race` sweep is yours.

**What checks out.** The fixture really does exercise what the PR says it does, which is the part most likely to be quietly untrue:

- `sumCtx(..., "warm", 1_000_000, true, 950_000)` gives 270s of remaining life against a 60s `DefaultPreExpiry` → genuinely Warm, and the precondition catches it if that arithmetic ever drifts.
- The shared conjunct is genuinely open rather than accidentally satisfied: `CtxWindowExact` true and `PrevBilledInput` 950,000 against a 900,000 floor, so `FracResolvable` holds and the fill gate is not what decides either subtest. The `MessagesTokens` guard in the first test pins that it is the *billed* figure doing it — the right precondition, and the one a reader would most want to distrust.
- Both mutations you describe follow from the code, and each really is confined to its own test: dropping `t.CacheState != CacheStateAny` makes the guard decline `any` at Unknown while leaving Warm untouched; `return p != CachePhaseWarm` in the `any` branch breaks Warm while Unknown still passes through. So the two tests are independent, not two views of one assertion.
- Configuration B is genuinely exercised through the registered constructor with `cache_state` absent for the default, so the `0.9` shared half is the shipped one rather than a string a test picked.

### The finding: the second test's negative row asserts no gate

`TestCacheStateAnyFiresOnALiveCacheAndTheDefaultDoesNot` asserts `cache_state_declined_warm`, which pins *why* the default declined. `TestCacheStateAnyStillFiresWhenAnUnknownPhaseHidesALivePrefix` asserts only `wantFires=false` on its default row — so that subtest passes if the default declines for **any** reason at all: `window_not_exact`, `below_request_trigger`, `no_model`, a future gate nobody has written yet. Its entire stated purpose is that *the live-prefix guard* is what declines there, and that is the one thing it does not check.

Your own table in the PR body names the gate — `cache_state_declined_unknown` — so the assertion was intended and just is not in the code. `Offload` files exactly that string (`"cache_state_declined_" + phase.String()`), so it is a two-line addition mirroring the first test.

Worth fixing before merge for a reason specific to this PR: it is the same shape as the defect closed in #234's last round, where a counter test re-performed the swap it existed to detect. A regression test whose negative half cannot fail for the thing it names reads as coverage while proving less than it claims — and here the half that is fully pinned is the one that was already true, while the half left loose is the novel guard.

Two smaller notes, neither worth a change on its own:

- The second test omits the first's `MessagesTokens` precondition. Both use `sumTranscript(6)`, so the fixture is effectively covered once, but if these ever stop sharing a transcript that guard needs copying.
- Both tests reuse session `"s"` against the package-global `inFlight`, which is fine only because the subtests are sequential and each builds its own store. A `t.Parallel()` added later would make them collide through the single-flight registry, which is the kind of thing worth a word in the comment since the file's drain helper hides it.

### Verdict

The regression is real, the pair discipline is the right shape for it, and the reasoning about why `any` must be exempt by name is correct — a raw-API harness presenting a live prefix with no readable TTL is exactly the deployment the opt-out exists for. Fix the missing gate assertion and this is good to merge as far as I am concerned; I cannot approve on GitHub, since it refuses self-approval while the PR and this session share an account.

amiddavid:
Fixed in 65eeadc. The finding was right, and it is the shape my own PR body argued against — a negative half that cannot fail for the thing it names. That the loose half was the *novel* guard while the fully-pinned half was the one already true by control flow is the part that makes it worth a round rather than a follow-up.

**The gate is named.** The unknown-phase default row now asserts `cache_state_declined_unknown`, the string `Offload` actually files, mirroring the warm test.

**The precondition is shared rather than inherited.** Extracted to `assertBilledFigureOpenedTheFill` and called by both tests. Your note was the reason to do it now instead of noting it: the new gate assertion is only meaningful while the shared fill conjunct is open, so the second test relying on `sumTranscript(6)` staying identical to the first's would have been the same class of latent vacuity one layer down.

**The parallel hazard is removed, not commented.** Each subtest gets its own session id (`"warm-"+tc.name`, `"unknown-"+tc.name`), with the reason recorded on `anyGateTrigger`. Cheaper to close than to leave as a caveat a future `t.Parallel()` would have to find.

### The third mutation

Worth stating separately, because it is the one the first two could not catch — I renamed the gate string in `Offload` so the default declines for a different reason while `wantFires=false` stays true:

```
sed -i 's|rep.Gate("cache_state_declined_" + phase.String())|rep.Gate("some_other_decline")|'
```

```
--- FAIL: …AndTheDefaultDoesNot/the_shipped_default_declines_the_same_turn
    summarize_anygate_test.go:82: gate "cache_state_declined_warm" not filed; gates: map[some_other_decline:1]
--- FAIL: …WhenAnUnknownPhaseHidesALivePrefix/the_shipped_default_is_declined_by_it
    summarize_anygate_test.go:147: gate "cache_state_declined_unknown" not filed; gates: map[some_other_decline:1]
```

Line 147 is the assertion that did not exist before this commit — that subtest passed this mutation as filed. Mutations 1 and 2 re-run against the final file and still each fail their own test alone (lines 137 and 77). `gofmt`/`vet` clean, `-race` green across `./components/...`.

Reviewed by inspection on your side and run on mine — noted, and the split is recorded in the commit.

PR #271, opened 2026-09-16
fix(offload): make cache-reuse compaction actually reuse the cache on Anthropic

OsherElhadad:
Follow-up to #256. Three things that shipped with it don't work; this makes them work. Both commits are independent and can be reviewed or reverted separately.

## Why this was needed

#256 landed the right idea — summarize by appending an instruction to the verbatim conversation, so the call reuses the prefix the agent's own turn already cached. Three gaps meant it couldn't deliver that in practice, and none of them could fail a build:

**1. The config we ship as the component's usage example doesn't parse.** `keep_first_turns` was dropped from `cacheAwareSummarizerConfig` when the head stopped being configurable, but the example still set it. Per-component blocks decode strictly, so the proxy refuses to build the pipeline and exits:

```
$ CONFIG=examples/llm-d-service/configs/cache-aware-summarizer.yaml ./context-guru-proxy
level=INFO msg="build pipeline: yaml: unmarshal errors:
  line 3: field keep_first_turns not found in type offload.cacheAwareSummarizerConfig"
```

That's the first file anyone copies. CI is green anyway because **nothing loads any config under `examples/`** — `grep -rl "examples/llm-d-service" --include=*_test.go .` is empty. A table test over `examples/**/configs/*.yaml` asserting each one builds would catch the whole class; left out here so these fixes stay reviewable, and worth its own change.

**2. `profile_fallbacks` reached nothing.** `CacheAwareSummarizerProfileFallbacks` had an exported accessor and no `Snapshot` field, so the counter was incremented into a variable `/stats` could not read. It is the only signal that a deployment's `profiles_path` override was unreadable and the **embedded** registry answered instead — which silently changes every instruction role the arm resolved, the exact misconfiguration the registry exists to prevent. `TestEverySnapshotFieldIsExportedOrExempt` reflects over `Snapshot`, so a counter that never becomes a field is invisible to it too.

**3. On an Anthropic-family model the component saved nothing.** This is the substantive one. Anthropic has **no automatic prefix caching** — it needs an explicit `cache_control` breakpoint — and the mapping sent none. So on any `provider: openai` → gateway → Bedrock/Anthropic route the component ran, reported success, and re-prefilled the whole conversation every call. `cache_read_input_tokens` reads 0 and the arm looks like it worked.

## What will work now, with numbers

Measured on `aws/claude-opus-4-7` through the ETE LiteLLM gateway. Same ~9.4k-token conversation each way, input cost at $5/MTok, 1.25× write / 0.10× read:

| summarization call | fresh | write | read | input $ | |
|---|---|---|---|---|---|
| appended instruction, **no breakpoint** (what main does today) | 9,616 | 0 | 0 | $0.04808 | 8.2× |
| appended instruction, **breakpoint** (this PR) | 237 | 0 | 9,378 | **$0.00587** | 1.0× |
| instruction **prepended**, breakpoint | 5 | 9,610 | 0 | $0.06009 | 10.2× |

Two results, and the second is the one #256 was waiting for:

- The breakpoint is worth **8.2×** on the summarization call.
- It is also what makes #256's central claim measurable at all: **10.2×** between rows 2 and 3 is the end-vs-start argument, confirmed on a real endpoint. The design thesis is right; it just needed the mark to be observable.

At the scale #256 quoted (1,372 calls on a 50-task arm): **$8.06** with the breakpoint, **$65.97** without.

## How the fix works

`CompleteMessages` marks the last content block of the **prefix** — every message except the final one — because `components.MessagesModel`'s contract is that the caller appends exactly one fresh message. Marking that last message instead would write a new entry every turn and read none: the failure this exists to remove, with the 1.25× write premium added on top.

The mark is conditional on the served model id containing `claude`, for two reasons that point the same way: a real OpenAI endpoint caches automatically so the mark buys nothing, and it rejects an unrecognised field inside a content part so sending one would break the call. That is the same substring test `minCacheablePrefix` next door already makes, and it fails in the same safe direction — a gateway alias that hides the real model gets no mark and behaves exactly as it does today.

Two things reused rather than reinvented: the existing `CacheablePrefix` gate, so we never buy a write premium for an entry the provider will ignore and that table stays in one place; and a JSON round-trip of the original content, so every field survives including block kinds this file does not model — a dropped field changes the prefix and loses the read the mark was added to get.

## Verification

`go build`, `go vet`, `gofmt`, and `go test -race ./...` clean on top of `0c21ead`.

- The shipped example config now starts the proxy (previously exited on a parse error).
- `/stats` reports `cache_aware_summarizer_profile_fallbacks`; golden key set and `notExportedWhy` updated following the `Summarize*` precedent.
- Three unit tests pin the request bytes: mark on the prefix, none on the appended instruction, none for a non-Anthropic id, none below the model floor.
- **Verified live, not only against a fake server.** On the shape production uses — the agent's turn writes the entry, the summarization call appends to it — the fixed client earned `write=0 read=6485`, where the unmarked client returned `write=0 read=0 fresh=9616` on every attempt. `zz_live_cache_test.go` keeps that runnable under `CG_LIVE`, because whether the provider *honours* the mark is the entire justification and no local assertion can show it.

## Two caveats, stated rather than discovered later

**Read-landing is not deterministic on a load-balanced gateway.** Requests that differ only in their trailing message can route to replicas with cold entries — I watched 6 consecutive marked calls all write and never read, then a 6,485-token read land on the next shape. The mark is honoured either way: writes appear where none did before. Judge a run on the provider's own `cache_read_input_tokens`, and expect variance that is the router's, not the method's.

**`Ctx.PrefixAsk` is the better answer where it is available, and it already exists.** `components.PrefixAsker` / `Anthropic.CompletePrefixed` already appends one user message to the parent's body and returns `PrefixUsage{CacheRead, CacheWrite, Fresh, Output}` — measured in-repo at 19,595 read against 0 created. It beats this on two counts: it preserves `tools` and top-level `system` byte-for-byte, which `CompleteMessages` structurally cannot since the pipeline sees neither; and it appends to **the body we actually sent last turn** rather than the incoming one. Its own doc says why that matters: *"The provider's prompt cache was populated by what context-guru emitted, which is the COMPACTED form. The incoming body is uncompacted, so it diverges from the cached bytes at the first thing any component removed."* That is worth a follow-up — it is a larger change, and `PrefixAsk` is nil on non-Anthropic routes and on the first turn, so the two are complementary rather than a swap.

Worth copying from there too: `PrefixUsage` is **returned** so a caller can *gate* on whether the read happened, and `extract_llm_sweep` declines when it didn't. This component only counts. A counter cannot support "don't bother if the read isn't landing," and on the evidence above that condition is real.

## Not included

`internal/cheapmodel/cachecontrol.go` groups `opus` with `sonnet` at a 1024-token floor, but Opus 4.7's real minimum is 2048 — so a 1024–2047-token prefix on `aws/claude-opus-4-7` is marked and silently ignored (write 0, read 0, no error). Same class as the haiku unit error that file was written to fix. Pre-existing, shared with other callers, and its own change.


OsherElhadad (commented):
Self-review of this PR, plus a re-test. I found three defects in my own first commit — two of which would have shipped **silently**, which is precisely the failure mode this change exists to remove. Fixed in `f4821b3`, and the live behaviour is materially better as a result.

## What was wrong

**1. It added a breakpoint the caller had already placed.** `bschemas.ChatContentBlock` carries a `CacheControl` field, and the verbatim content passthrough preserves it — so an agent that caches its own prompt (Claude Code does) arrives here **already marked**. Two problems with adding ours on top:

- The caller's breakpoint is the one its cache entry was actually created under, so it's the one a read has to match. Ours buys nothing.
- The provider allows only **four** breakpoints per request, and a fifth is a 400 on the whole call — which on this path costs a summarization and reverts the component.

An already-marked prefix now gets no mark from us.

**2. It could silently send no breakpoint at all.** The index was `len-2` unconditionally, but a breakpoint needs a content block to attach to and an assistant turn that is purely tool calls has **no content**. When that message landed at `len-2` the mark was never applied and the request went out unmarked — no error, no counter, and byte-identical on the wire to the bug being fixed. The index now walks back to the last message that *can* carry a mark, which keeps a shorter prefix: less is cached, nothing breaks.

**3. It duplicated `schema.MessageText`,** and the comment justifying that ("this package sits below the public ones") was simply wrong — this file already imports `components`, and `schema` doesn't import `cheapmodel`, so there's no cycle. `schema.MessageText` is also the better implementation: it avoids the O(n²) accumulation this path would hit on a multi-block tool message.

## Re-tested live, and it's the clean result now

`aws/claude-opus-4-7`, the shape production actually uses — the agent's turn writes the entry, each summarization call appends its instruction:

```
agent turn            fresh=10  write=6485  read=0
summarization call 1  fresh=37  write=0     read=6485
summarization call 2  fresh=37  write=0     read=6485
summarization call 3  fresh=37  write=0     read=6485
summarization call 4  fresh=37  write=0     read=6485
```

**4 of 4 reads.** Before `f4821b3` the summarization calls wrote every time instead. The summarization call now costs **$0.00343** of input where the unmarked shape costs **$0.03261** — **9.5×**.

That also supersedes the caveat I put in the original PR body about read-landing being unreliable on this gateway. It wasn't the router: it was my own bug producing a request shape that couldn't read. Routing variance is real but much smaller than I claimed, and the corrected result is consistent across four consecutive calls.

## Verification

`go build`, `go vet`, `gofmt`, `go test -race ./...` all clean on `0c21ead`. Two new tests, both for cases that would have shipped broken:

- `TestCompleteMessagesLeavesACallerMarkedPrefixAlone` — a caller-marked prefix must end with **exactly one** `cache_control` in the request.
- `TestCompleteMessagesWalksBackToAMarkableMessage` — a tool-call-only message must not be the marked one, and the mark must land on the last message that can hold it.

Five wire tests now pin the request bytes in total, plus `zz_live_cache_test.go` under `CG_LIVE` for the part no fake server can show.

## Still deliberately out of scope

Unchanged from the original description, restated so it isn't lost in this thread:

- **`Ctx.PrefixAsk` remains the better answer where available** — it preserves `tools` and top-level `system` byte-for-byte and appends to the body we actually *sent* rather than the incoming one. Bigger change, nil on non-Anthropic routes, so complementary rather than a swap.
- **`PrefixUsage` is returned so a caller can gate on whether the read happened;** `extract_llm_sweep` declines when it didn't. This component only counts.
- **`cachecontrol.go` groups `opus` with `sonnet` at a 1024-token floor** while Opus 4.7's real minimum is 2048 — so a 1024–2047-token prefix on this gateway's model is marked and silently ignored. Pre-existing, shared with other callers, its own change.

PR #273, opened 2026-09-16
fix(cheapmodel): don't double-mark a cached prefix, and never mark a message that can't carry a breakpoint

OsherElhadad:
Follow-up to #271, which merged while I was reviewing my own change in it. These are the fixes from that self-review — the comment describing them is [here](https://github.com/rossoctl/context-guru/pull/271#pullrequestreview-5224576476), posted a few minutes before the merge, so they never landed.

**Both defects are live in `main` now, and both fail silently** — which is the exact failure mode #271 existed to remove.

## 1. It adds a breakpoint the caller already placed

`bschemas.ChatContentBlock` carries a `CacheControl` field and the verbatim content passthrough preserves it, so an agent that caches its own prompt — Claude Code does — arrives here **already marked**. Adding ours on top is wrong twice:

- The caller's breakpoint is the one its cache entry was actually created under, so it's the one a read has to match. Ours buys nothing.
- The provider allows only **four** breakpoints per request, and a fifth is a **400 on the whole call** — on this path that costs a summarization and reverts the component.

An already-marked prefix now gets no mark from us.

## 2. It can silently send no breakpoint at all

The index was `len-2` unconditionally, but a breakpoint needs a content block to attach to and an assistant turn that is purely tool calls has **no content**. When such a message landed at `len-2` the mark was never applied and the request went out unmarked — no error, no counter, byte-identical on the wire to the bug #271 fixed.

The index now walks back to the last message that *can* carry a mark. That keeps a shorter prefix, which is the safe direction: less is cached, nothing breaks.

## 3. Duplicated `schema.MessageText`

The local `messageText` was a second copy, and the comment justifying it ("this package sits below the public ones") was wrong — this file already imports `components`, and `schema` doesn't import `cheapmodel`, so there's no cycle. `schema.MessageText` is also better here: it avoids the O(n²) accumulation this path would hit on a multi-block tool message.

## What changes measurably

`aws/claude-opus-4-7`, the shape production uses — the agent's turn writes the entry, each summarization call appends its instruction:

```
agent turn            fresh=10  write=6485  read=0
summarization call 1  fresh=37  write=0     read=6485
summarization call 2  fresh=37  write=0     read=6485
summarization call 3  fresh=37  write=0     read=6485
summarization call 4  fresh=37  write=0     read=6485
```

**4 of 4 reads.** Before this change the summarization calls wrote every time instead. The call now costs **$0.00343** of input against **$0.03261** unmarked — **9.5×**.

This also **retracts a caveat I put in #271's description**: I attributed unreliable read-landing to the gateway's load balancing. It wasn't the router — it was defect 2 above producing a request shape that could not read. Routing variance is real but far smaller than I claimed, and the corrected result is consistent across four consecutive calls.

## Verification

`go build`, `go vet`, `gofmt`, `go test -race ./...` clean on `22319b7`. Two new tests, both for cases now live in main:

- `TestCompleteMessagesLeavesACallerMarkedPrefixAlone` — a caller-marked prefix must end with **exactly one** `cache_control` in the request.
- `TestCompleteMessagesWalksBackToAMarkableMessage` — a tool-call-only message must not be the marked one, and the mark must land on the last message that can hold it.

## Still out of scope, restated so it isn't lost

- **`Ctx.PrefixAsk` is the better answer where available** — preserves `tools` and top-level `system` byte-for-byte, and appends to the body we actually *sent* rather than the incoming one. Bigger change; nil on non-Anthropic routes, so complementary rather than a swap.
- **`PrefixUsage` is returned so a caller can gate on whether the read happened** (`extract_llm_sweep` declines when it didn't). This component only counts.
- **`cachecontrol.go` groups `opus` with `sonnet` at a 1024-token floor** while Opus 4.7's real minimum is 2048, so a 1024–2047-token prefix on this gateway's model is marked and silently ignored. Pre-existing, shared with other callers.


OsherElhadad:
## Review

**Confirmed both defects, on `main`, with reproductions.**

1. **Double-mark.** The PR's own `TestCompleteMessagesLeavesACallerMarkedPrefixAlone` happens to pass even against pre-fix `main`, because in that specific test the caller's existing mark sits on the exact same message `len(msgs)-2` would also mark — `markedContent` overwrites the same `cache_control` key rather than adding a second one, so the byte-count assertion can't see the bug. I added a 3-message variant (marked msg, an unmarked middle message, then the instruction) and confirmed pre-fix `main` genuinely emits two `cache_control` blocks in that case (`grep -c cache_control` → 2), and that `d6a151c` (this PR) reduces it back to 1. Logic in `cacheControlPrefixEnd` (internal/cheapmodel/openai.go:95-107) — bailing to `-1` as soon as *any* prefix message already carries a mark — is correct and conservative: it means don't add a second mark anywhere once one exists anywhere in the prefix, since matching the byte-identical marked prefix the caller's own cache entry was written under is what earns the read.

2. **Tool-call-only turn walk-back.** Reproduced/verified via the PR's own `TestCompleteMessagesWalksBackToAMarkableMessage`: on pre-fix `main` (plain `len(msgs)-2`), the mark lands on message 0 as a bare *string* (unmarked) instead of walking back — confirmed by running that exact test against `git show main:internal/cheapmodel/openai.go`. Post-fix it walks back correctly via `markedContent(prefix[i].Content)` (openai.go:118-121).

**Coverage check on `hasCacheControl`.** Checked bifrost's `bschemas.ChatContentBlock`/`ChatMessageContent` (core@v1.7.0) — `CacheControl` only exists on `ChatContentBlock`, never on a bare `ContentStr`, and never on `ChatAssistantMessageToolCall` in the chat-completions schema. So `hasCacheControl` (openai.go:127-137), which only walks `m.Content.ContentBlocks`, has no gap for this schema version.

**Live verification.** Ran the fixed code against a real LiteLLM→Anthropic gateway (not the one referenced in the PR description) with `CG_LIVE=1`, 6 back-to-back identical `CompleteMessages` calls: request bytes were byte-identical across calls (confirmed via a wrapped `http.RoundTripper`) and correctly carried exactly one `cache_control` block each time. However, `cache_read_input_tokens` was 0 and `cache_creation_input_tokens` was paid on *every* call, 0/6 — the opposite of the PR's "4 of 4 reads" result. Byte-identical, correctly-marked requests confirm this PR's logic is doing the right thing at the request-construction layer; whether a read actually lands is downstream, gateway/routing-dependent behavior (matches the "routing variance is real" caveat already in the PR description) — flagging as an observation, not a defect in this diff.

**Minor:** `cacheControlPrefixEnd`'s walk-back loop calls `markedContent` purely to test feasibility (discarding the result), and `CompleteMessages` then calls it again at `markAt` — a second JSON marshal of the same content. Not a bug, just a duplicated marshal; not worth a diff for.

No blocking issues found. `go build`, `go vet`, `gofmt -l`, and `go test ./...` are clean for this PR's own package (`internal/cheapmodel`) both before and after my additional live checks. The two `dash` package test failures visible in a full `go test ./...` run (`TestCalculatorPricesFromTheRowsModelNotABlend`, `TestNavHashCompatibility`) reproduce identically on `main` HEAD (unrelated to this change — a pricing-model drift and a `node`-test-runner version mismatch).

**Mergeability:** `gh pr checks` all green (build-test, purego, trivy, DCO). `mergeStateStatus: BLOCKED` is because `main`'s ruleset requires 1 approving review and this PR has zero reviews yet — not a CI or conflict issue. No merge conflicts with `main` (`mergeable: MERGEABLE`), though the branch is ~162 commits behind `main`; out of scope to rebase for a correctness-only fix, but worth a rebase before merge if that drift is a concern.

PR #279, opened 2026-09-18
feat(dash,proxy): share current/live/all savings totals between /stats and dashboard

amiddavid:
## Summary
- Extracts `dash.DB.Overview()`'s cost/savings arithmetic into a new, smaller `SavingsTotals(f Filter)` query; `Overview()` now calls it instead of duplicating the formula, so the dashboard and `/stats` can never disagree about "total saved".
- `proxy.Handler.stats()` calls `SavingsTotals()` directly on `h.rec.DB()` (single-tenant only, no `dash.API`/dashboard dependency) and reports three scopes on `/stats`: `current` (the session that most recently sent a real request), `live` (every session the keeper still tracks, summed), and `all` (process lifetime). `keeper.LiveSessionKeys()` backs the `live` scope.
- Re-touches the plugin's docs/skills (`README.md`, `docs/how-to/install-plugin.md`, `context-guru-plugin/skills/{status,keepalive,cache-strategy-picker}/SKILL.md`, `plugin.json`) to read `/stats`' `savings` block instead of requiring the dashboard for `keepalive_net_usd`/`total_saved_usd`.

## Test plan
- [x] `dash`: `SavingsTotals` agrees with `Overview()`'s own savings figures on the same rows (cachesplit-only, priced-LLM, and keep-alive traffic).
- [x] `proxy`: real `dash.Recorder(":memory:")`-backed test — `savings.current` matches only the most recent real session, `savings.live` sums multiple live sessions, `savings.all` matches the full-DB `Overview()` figure.
- [x] Negative cases: fresh proxy → no `current`; no dashboard (`h.rec == nil`) → no `savings` block; hosted (`h.opts.Tenants != nil`) → no `savings` block.
- [x] `gofmt -l`, `go vet`, `go test ./dash/... ./proxy/... ./cmd/...` clean on the eval box.
- [x] Live end-to-end proof on the eval box, through the plugin's own `start-proxy.sh` (armed with the real `5-min-ping` cache strategy via `settings.py strategy set`, not a manual binary start): two real turns one session apart against a real upstream, a real autonomous keep-alive ping landing past the 5-minute cache TTL, and `/stats` reporting a real nonzero `keepalive_net_usd`/`total_saved_usd` in `current`, `live`, and `all`. Torn down by exact pid; the box's real `~/.claude/settings.json` and the unrelated model-info server on `:6980` were untouched.


proxy/savings.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil
+	}
+	db := h.rec.DB()
+	out := &SavingsStats{}
+	if all, err := db.SavingsTotals(dash.Filter{TenantAll: true}); err == nil {
+		out.All = savingsScope(all)
+	} else {
+		slog.Default().Warn("cg.stats_savings_all_failed", "err", err)
+	}
+	if live := h.keeper.LiveSessionKeys(); len(live) > 0 {
```
amiddavid, line 84: CONFIRMED: the `live` scope issues one `dash.DB.SavingsTotals` call (itself 2 sequential SQL queries) per live session in this loop, instead of one grouped query. With `maxKeepAliveSessions = 512` (proxy/keepalive.go:188), a single `GET /stats` can trigger ~1024 sequential SQL round trips. The codebase already has a grouped-query pattern for this exact shape (`dash.DB.sumBySession` in dash/keepalive.go) that could compute the same sums with a `session_id IN (...)` query instead of N+1.

proxy/proxy.go
```diff
@@ -260,6 +261,15 @@ type Handler struct {
...
 	metricsInflight singleflight.Group
+
+	// lastSessionMu guards lastSession/lastSessionAt.
+	lastSessionMu sync.Mutex
+	// lastSession is the session id of the most recent REAL (non-ping) request this
+	// handler served — /stats' "current" scope. Set only on the request path (chat),
+	// never on the keep-alive ping path, so a ping can never make itself "current".
+	// Empty until the first real request arrives.
+	lastSession   string
+	lastSessionAt time.Time
```
amiddavid: CONFIRMED: `lastSessionAt` is written in `setLastSession` (proxy/savings.go) but never read anywhere in the tree (verified via full-tree grep). Looks like a forgotten staleness/expiry check for the `current` scope — without it, a session from days ago keeps being reported as "current" forever. If that's not the intent, this field is dead state.

amiddavid (commented):
Automated review (round 1) — 2 confirmed findings.

amiddavid:
Addressed both CONFIRMED findings from review #5247868670 in 4aeec90:

- **N+1 in the `live` scope**: `dash.Filter` gained a `SessionIn []string` dimension (`session_id IN (...)`), and `proxy/savings.go`'s `savingsStats()` now calls `db.SavingsTotals` once with all live session ids instead of looping per session. `live` now costs the same 2 SQL queries as `current`/`all`, regardless of session count.
- **Dead `lastSessionAt`**: removed from `Handler` — it was written on every real request and never read; no staleness check for `current` was ever wired up, so there was nothing to finish.

Added `TestSavingsTotalsSessionInSumsAcrossSessionsInOneQuery` (dash) and updated the existing `/stats` three-scope test to compare against the same grouped query the code now runs, rather than summing two independent per-session queries in Go (which can legitimately differ in the last float64 bit from one SQL SUM). `go build`/`go vet`/`go test ./dash/... ./proxy/...` all green.

Please re-review.

amiddavid (commented):
Round 2: verified commit 4aeec90. Both round-1 findings are fixed correctly — grouped `SessionIn` query replaces the N+1 loop (parameterized IN clause, no injection risk), dead `lastSessionAt` removed. New test compares against the same grouped query rather than summing per-session results, avoiding a float-rounding false negative. No new issues found. LGTM from me (can't self-approve, same account as author).

